PR #36, opened 2025-02-04
feat: Generalize output monad of simulation

dtumad:
PR to help consider potential generalizations to `simulate` and potential trade-offs.
The main change is two new experimental versions of `SimOracle`:
```
structure SimOracle' {ι : Type u} {ιₜ : Type u'}
    (spec : OracleSpec.{u,v} ι) (specₜ : OracleSpec.{u',v'} ιₜ)
    (T : (Type w' → Type (max u' (v' + 1) w')) → Type v → Type w) where
  impl {α : Type v} (q : OracleQuery spec α) :
    T (OracleComp.{u',v',w'} specₜ) α

structure QueryImpl {ι : Type u} (spec : OracleSpec.{u,v} ι)
    (m : Type v → Type w) where
  impl {α : Type v} (q : OracleQuery spec α) : m α
```
The first allows the output to have a varying monad transformer `T`, the second is more general and allows an arbitrary output monad `m`. Lemmas for the first have one tick-mark, those for the second have two. First version can be less verbose but has more complicated universe levels and is less general. (Also changed `OracleSpec` to have a single universe level for the domain and range of oracles to try and alleviate the universe issue a bit. Writing `simulateT'` for use of the first and `simulateQ` for use of the second

This allows for example the use of `WriterT` and `tell` instead of `StateT` and `modify` in the definition of `loggingOracle`:
```
def QueryLog' (spec : OracleSpec ι) : Type :=
  List (Σ i : ι, spec.domain i × spec.range i)

def loggingOracle' : SimOracle' spec spec (WriterT (QueryLog' spec)) where
  impl | query i t => do
    let u ← query i t
    tell [⟨i, (t, u)⟩]
    return u

def logingOracle'' : QueryImpl spec (WriterT (QueryLog' spec) (OracleComp spec)) where
  impl | query i t => do
    let u ← query i t
    tell [⟨i, (t, u)⟩]
    return u

-- `(22, [⟨20, ((), 8)⟩, ⟨20, ((), 14)⟩])`
-- `(9, [⟨20, ((), 6)⟩, ⟨20, ((), 3)⟩])`
#eval OracleComp.runIO (simulateT' loggingOracle' (do
    (Nat.add) <$> $[0..20] <*> $[0..20])).run
```

Similarly for `countingOracle`, using a list of indices rather than a count (Lean tries to use multiplication instead of addition when I use an actual `Nat` to count:
```

-- Track a list of oracle indexes for each query
def countingOracle' {ι : Type u} [DecidableEq ι] {spec : OracleSpec ι} :
    SimOracle' spec spec (WriterT (List ι)) where
  impl | query i t => do tell [i]; query i t

def countingOracle'' {ι : Type u} [DecidableEq ι] {spec : OracleSpec ι} :
    QueryImpl spec (WriterT (List ι) (OracleComp spec)) where
  impl | query i t => do tell [i]; query i t

-- Outputs [3,1,4,1,5] as the state
#eval! OracleComp.runIO (simulateT' countingOracle' (do
    $[0..3]; $[0..1]; $[0..4]; $[0..1]; $[0..5]; return ())).run
```

Also generalizes things like `evalDist`:
```
noncomputable def evalDist' [spec.FiniteRange] :
    (oa : OracleComp spec α) → OptionT PMF α :=
  simulateQ ⟨fun (query i _) => PMF.uniformOfFintype (spec.range i)⟩
```

Can define a version of append that automatically includes lifting as needed:
```
def append'' (m₁ m₂ n : Type u → Type v)
    [MonadLift m₁ n] [MonadLift m₂ n]
    (so : QueryImpl spec₁ m₁) (so' : QueryImpl spec₂ m₂) :
    QueryImpl (spec₁ ++ₒ spec₂) n where impl
  | query (inl i) t => so.impl (query i t)
  | query (inr i) t => so'.impl (query i t)
```


VCVio/OracleComp/SimSemantics/Simulate.lean
```diff
@@ -20,17 +20,28 @@ For example `idOracle` has no effect upon simulation, and we should apply that f
...
-universe u v w
+universe u u' v v' w w' z
 
-variable {ι : Type u} {ιₜ : Type v} {spec : OracleSpec ι} {specₜ : OracleSpec ιₜ}
-  {α β σ : Type w}
+variable {ι ιₜ : Type*} {spec : OracleSpec ι} {specₜ : OracleSpec ιₜ}
+  {α β σ : Type u}
+
+-- T is e.g. `StateT`, `OptionT`, `WriterT`
+structure SimOracle' {ι : Type u} {ιₜ : Type u'}
```
dtumad, line 29: Version of `SimOracle` where the monad transformer being applied to `OracleComp` is allowed to be some arbitrary `T`.

VCVio/OracleComp/SimSemantics/Simulate.lean
```diff
@@ -20,17 +20,28 @@ For example `idOracle` has no effect upon simulation, and we should apply that f
...
 open OracleSpec Prod
 
-universe u v w
+universe u u' v v' w w' z
 
-variable {ι : Type u} {ιₜ : Type v} {spec : OracleSpec ι} {specₜ : OracleSpec ιₜ}
-  {α β σ : Type w}
+variable {ι ιₜ : Type*} {spec : OracleSpec ι} {specₜ : OracleSpec ιₜ}
+  {α β σ : Type u}
+
```
dtumad, line 27: Version of `SimOracle` that allows an arbitrary monad transformer `T` to be applied to the output state

VCVio/OracleComp/SimSemantics/Simulate.lean
```diff
@@ -20,17 +20,28 @@ For example `idOracle` has no effect upon simulation, and we should apply that f
...
+-- T is e.g. `StateT`, `OptionT`, `WriterT`
+structure SimOracle' {ι : Type u} {ιₜ : Type u'}
+    (spec : OracleSpec.{u,v} ι) (specₜ : OracleSpec.{u',v'} ιₜ)
+    (T : (Type w' → Type (max u' (v' + 1) w')) → Type v → Type w) where
+  impl {α : Type v} (q : OracleQuery spec α) :
+    T (OracleComp.{u',v',w'} specₜ) α
+
```
dtumad, line 34: `QueryImpl` as an even more general version, allowing an arbitrary monad `m` in the output type

dtumad:
@quangvdao If you could check this out when you have a chance would appreciate thoughts on this approach. Makes your comment about `WriterT` for logging possible, and I think it has other benefits as well. Still in a rough state but want to settle on an approach before getting too deep in the weeds.

quangvdao:
I'll need to think about these new designs. As a first step, I pushed two changes about accommodating `WriterT` instances for `QueryLog` and `Nat`.

ToMathlib/FreeMonad.lean
```diff
@@ -147,24 +181,92 @@ end mapM
...
+
+-- @[simp] lemma failure_seq [LawfulMonad m] (ob : m α) : (failure : m (α → β)) <*> ob = failure := by
+--   rw [seq_eq_bind_map, failure_bind]
+
+-- end AlternativeMonad
+
+section LawfulFailure
+
+-- Stripped down version of `Alternative` to avoid diamonds.
+class Failure (m : Type u → Type v) where
```
dtumad, line 206: I think we need this type-class rather than `Alternative` in some definitions to avoid the issue of both `Monad` and `Alternative` having `toApplicative` instances, which causes mismatches in proofs. `OptionT.mapM_pure` is a good example where it tries to use the `Pure` instance coming from `Alternative` instead of `Monad`.

`simp` can reduce `Failure.fail` to `failure` if an `Alternative` instance exists so doesn't really break anything.

dtumad:
Merging the changes after adding changes from #38 and fixing merge issue, adding the remaining conversion to a new branch.

PR #51, opened 2025-03-31
feat: improved foundations and cleaned up documentation

dtumad:
The main goal of this PR is to get all examples functional again and to update the documentation of the foundations of the system. There are also a lot of changes to standardize things along with this. (WIP) Summary of changes:

## Random Breaking Changes

* `simulateQ so (oa >>= ob)` now reduces to `simulateQ so oa >>= simulateQ so ∘ ob`, some proofs may need to use `Function.comp_def` now
* Significant restructuring of `OracleComp/SimSemantics` directory that probably breaks imports
* `simulateQ` can be called if the output monad isn't lawful (similar for `mapM`). Some instances are now flagged as unneeded
* Github workflow should now start checking for linting errors (after this merges)

## Generalized Support of a Computation

`OracleComp.support` is now a special case of `OracleComp.supportWhen possible_outputs`, where `possible_outputs` specifies the potential outputs of an oracle query. This makes it possible to express things like "possible outputs if the oracle only returns bitstrings of length `n`". They also both can now be defined via `simulateQ` using monad instances on set (`evalDist` is now defined this way as well, so `simulateQ` is now the only "foundational" semantics):

```
def supportWhen' (oa : OracleComp spec α)
    (possible_outputs : {α : Type v} → OracleQuery spec α → Set α) : Set α :=
  oa.simulateQ ⟨fun | q => possible_outputs q⟩

def support' (oa : OracleComp spec α) : Set α :=
  oa.simulateQ ⟨fun | query i _ => Set.univ⟩
```

## Propositional Traversal of a Computation

Define functions `OracleComp.allWhen` and `OracleComp.someWhen` for traversing a computation and checking some predicate on all nodes, that also allow a set of possible outputs to use during traversal.

For example `allWhen Q F possible_outputs oa` will traverse the computation `oa`, checking `F` on any failures and `Q` on any queries; for each query `q` it only continues traversing the computation on outputs `x` in `possible_outputs q`. `someWhen` is similar but uses `∨` instead of `∧` to accumulate the result. These operations naturally relate to `supportWhen` when proving statements:

```
lemma allWhen_bind_iff (oa : OracleComp spec α) (ob : α → OracleComp spec β) :
    (oa >>= ob).allWhen Q F possible_outputs ↔ oa.allWhen Q F possible_outputs ∧
      ∀ x ∈ oa.supportWhen possible_outputs, (ob x).allWhen Q F possible_outputs := _
```

`OracleComp.all Q F oa` and `OracleComp.some Q F oa` assume any possible query output. `neverFailWhen`, `mayFailWhen`, `neverFail`, and `mayFail` are defined as special cases of this. The old `noFailure` should be deprecated into cases of these.

## Tranformers for Query Implementations

Most of the "Query Tracking" functionality now exists as transformers rather than base oracles. For example `QueryImpl.withLogging` takes an existing query implementation and lifts it to a writer monad where queries can be tracked:

```
def withLogging (so : QueryImpl spec m) :
    QueryImpl spec (WriterT (QueryLog spec) m) where
  impl | q => do let x ← liftM (so.impl q); tell (QueryLog.singleton q x); return x
```

The old definition of `loggingOracle` is definitionally equal to `idOracle.withLogging` and has been redefined to that. similar for `cachingOracle` to `withCaching`, `countingOracle` to `withCounting`, and `seededOracle` to `withPregen`.

Random oracles are now just `unifOracle.withCaching`

## Simplified `ExecutionMethod`

Instead of taking using intermediate monad to execute oracle queries in a computation, `ExecutionMethod` has been stripped to just having a reasonable mapping into `ProbComp`. This allows us to define things like `SignatureAlg` over an arbitrary monad, using the execution method only when we need to reduce down for probability calculations. If the algorithms are all in `ProbComp` then `ExecutionMethod.default` can execute them with `id` giving good definitional equality. This also makes it possible to define algorithms over monads that have a state transformer applied for example. Elgamal for example now looks like:

```
structure ExecutionMethod (m : Type u → Type v) where
    exec {α : Type u} : m α → ProbComp α
    lift_probComp {α : Type u} : ProbComp α → m α
    exec_lift_probComp {α : Type u} (px : ProbComp α) :
      exec (lift_probComp px) = px

protected def default : ExecutionMethod ProbComp where
  exec := id
  lift_probComp := id
  exec_lift_probComp _ := rfl

structure AsymmEncAlg (m : Type → Type u) (M PK SK C : Type)
    extends ExecutionMethod m where
  keygen : m (PK × SK)
  encrypt : (pk : PK) → (msg : M) →  m C
  decrypt : (sk : SK) → (c : C) → Option M

def elgamalAsymmEnc (G P : Type) [SelectableType G] [SelectableType P]
    [AddGroup G] [Group P] [AddTorsor G P] : AsymmEncAlg ProbComp
    (M := P) (PK := P × P) (SK := G) (C := P × P) where
  keygen := do
    let x₀ ←$ᵗ P; let sk ←$ᵗ G
    return ((x₀, sk +ᵥ x₀), sk)
  encrypt := fun (x₀, pk) msg => do
    let g ←$ᵗ G
    return (g +ᵥ x₀, msg * (g +ᵥ pk))
  decrypt := fun sk (c₁, c₂) => do
    return c₂ / (sk +ᵥ c₁)
  __ := ExecutionMethod.default
```

Some security definitions may need to specialize `m` to be `OracleComp spec` still, if they need to add additional oracles (e.g. IND-CCA security)

 ## Examples

* Fixing one-time pad, elgamal, and DH examples
* Updated forking to use `QueryImpl.withPregen` and make the `QueryImpl.withLogging` portion implicit
* Updated F-S to commitment scheme from sigma protocol rather than signature
* KEM+DEM -> PKE 

## Misc Changes

* Some fixes to `LawfulMonadState`. Pull in changes to `LawfulAlternative`
* Better probability lemmas for some list operations that come up in looping
* `sorry` lemmas without downstream use are commented out for now


Examples/HACS/main.lean
quangvdao, line 1: delete this?

quangvdao:
Looks good to me, thanks for the big changes.

For the submission, is the plan to finish the examples? For instance, re-complete the forking lemma, define Fiat-Shamir signature, and finish reduction from ElGamal to DDH?

dtumad:
> Looks good to me, thanks for the big changes.
> 
> For the submission, is the plan to finish the examples? For instance, re-complete the forking lemma, define Fiat-Shamir signature, and finish reduction from ElGamal to DDH?

Yes, I'd like to get all those in as part of this PR as well before it's merged

PR #61, opened 2025-05-07
feat: syntax for binding variables in probability calculations

dtumad:
This PR is a WIP to add support for syntax like `[∀ i : Fin 3, y[i] ≠ 0| y ←$ᵗ Vector F 3]`, where the variable in the predicate is bound at the site of the computation. So far can support things like:

* Multiple variables: `Pr[(x ∧ y) ∨ (x ∧ !z) | x, y, z ←$ᵗ Bool]`
* Multiple computations: `Pr[b ∧ b' ∧ x * y = 1 | b, b' ←$ᵗ Bool; x, y ←$ᵗ F]`
* Arbitrary computations `Pr[!x ∨ !y | x, y ← my_computation]`

The first allows for arbitrary numbers of variables. The last only works for up to three clauses now. Ideally this could be generalized but the syntax manipulation is trickier.

Long-term it would be nice if computations in later clauses can depend on variables in the earlier ones. This seems like it requires manually building up do terms.

### DIFF
--- a/Examples.lean
+++ b/Examples.lean
@@ -5,3 +5,4 @@ import Examples.HHS_Signature
 import Examples.OneTimePad
 import Examples.RF_RP_Switching_alt
 import Examples.Regev
+import Examples.SyntaxTesting
--- a/Examples/SyntaxTesting.lean
+++ b/Examples/SyntaxTesting.lean
@@ -1,9 +1,9 @@
 import VCVio
 
-open Lean OracleComp
+open Lean OracleComp ENNReal
 
 syntax (name := probThing)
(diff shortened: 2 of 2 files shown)


dtumad:
> Note: thanks to @Ferinko we now have a pretty general notation for probability: https://github.com/Verified-zkEVM/ArkLib/blob/main/ArkLib/Data/Probability/Notation.lean

I will probably just try to pull that into #70 for any `HasEvalDist` monad or the followup to it unless there is a reason not to @quangvdao @Ferinko

dtumad:
Largely subsumed by #70

PR #70, opened 2025-05-28
feat: Generalize `evalDist` to arbitrary monads

dtumad:
This PR generalizes `evalDist` from being specific to `OracleComp` to any monad that can lawfully embed into `SPMF`. The resulting definitions and notations for e.g. `probEvent` are similarly changed. 

The definition of `SPMF` should eventually be ported to mathlib. This currently doesn't lead to any breaking changes but we should at some point deprecate the old notation if we want to commit to this.


quangvdao:
Looks good to me. You should merge once all errors are resolved

VCVio/EvalDist/Basic.lean
```diff
@@ -0,0 +1,240 @@
...
+-/
+-- import ToMathlib.ProbabilityTheory.Coupling
+import Mathlib.Probability.ProbabilityMassFunction.Monad
+import ToMathlib.General
+
+
+/-!
+# Denotational Semantics for Output Distributions
+
+This file defines a typeclass for monads that can be canonically embedded in the `PMF` monad.
```
quangvdao, line 14: You mean SPMF

VCVio/EvalDist/Basic.lean
```diff
@@ -0,0 +1,240 @@
...
+  coe sp x := sp.run (some x)
+  coe_injective' p q h := by
+    simp [eq_iff_forall_some]
+    intro x
+    simpa using congr_fun h x
+
+end SPMF
+
+/-- The monad `m` has a well-behaved embedding into the `SPMF` monad. This means the -/
+class HasEvalDist (m : Type u → Type v) [Monad m] where
```
quangvdao, line 63: What's wrong with defining this directly in terms of MonadHom m SPMF?

i.e. ... extends MonadHom m SPMF where

and then the evalDist can be an alias

VCVio/EvalDist/Basic.lean
```diff
@@ -0,0 +1,240 @@
...
+  evalDist {α : Type u} (mx : m α) : SPMF α
+  evalDist_pure {α : Type u} (x : α) : evalDist (pure x : m α) = pure x
+  evalDist_bind {α β : Type u} (mx : m α) (my : α → m β) :
+    evalDist (mx >>= my) = evalDist mx >>= evalDist ∘ my
```
quangvdao, line 67: I think not using Functor.comp may make some rewriting easier

VCVio/EvalDist/Basic.lean
```diff
@@ -0,0 +1,240 @@
...
+@[simp] lemma one_le_probEvent_iff : 1 ≤ Pr[p | mx] ↔ Pr[p | mx] = 1 := by
+  simp only [le_iff_eq_or_lt, not_one_lt_probEvent, or_false, eq_comm]
+@[simp] lemma one_le_probFailure_iff : 1 ≤ Pr[⊥ | mx] ↔ Pr[⊥ | mx] = 1 := by
+  simp only [le_iff_eq_or_lt, not_one_lt_probFailure, or_false, eq_comm]
+
+end bounds
+
+namespace SPMF
+
+/-- Add instance for `SPMF` just to give access to notation. -/
```
quangvdao, line 190: Also add instance for PMF into SPMF

quangvdao: We should think about getting PMF theorems "for free", or equivalently, giving API support for lossless distribution (i.e. PMF within SPMF)

VCVio/EvalDist/Basic.lean
```diff
@@ -0,0 +1,240 @@
...
+
+@[simp] lemma probEvent_eq : probEvent p = p.run.toOuterMeasure ∘ Set.image some := rfl
+
+@[simp] lemma probFailure_eq : probFailure p = p.run none := rfl
+
+end SPMF
+
+namespace PMF
+
+noncomputable instance hasEvalDist : HasEvalDist PMF where
```
quangvdao, line 210: Oops nevermind my comment above, here is the instance

PR #71, opened 2025-05-28
chore: adjust basic poly-time def

dtumad:
This PR adjusts the definition of polynomial time computations to be a reasonable placeholder until a full solution has been implemented. The definition is meant to be sufficient as a placeholder until a more comprehensive definition exists.

This is a partial-blocker for the definition of a decider and in particular the elgamal security proofs. The concrete reductions don't require it but it allows some nicer abstractions.

The extension to `OracleComp` is not complete in that some "obviously" poly-time computations aren't covered, but it should be sufficient in that every adversary has an equivalent poly-time implementation, which is enough when quantifying over all adversaries.


dtumad:
Very stale at this point

PR #79, opened 2025-07-14
feat: Update base monad model 

dtumad:
This PR makes three closely related changes the base monad model used to represent computations. Definitions have been updated but some lemmas may be missing/moved/deprecated:

- `OptionT` is removed from the basic monad stack and manually included as needed
- `Pfunctor.FreeM` is used over the general Free Monad, and `OracleSpec` is an alias for `PFunctor`.
- Probability semantics now depend on generic type-classes

In particular we now have:
```
def OracleComp (spec : OracleSpec.{u,v}) : 
    Type w → Type (max u v w) :=
  PFunctor.FreeM spec

abbrev ProbComp : Type z → Type z := OracleComp unifSpec
```

So that type universe levels no longer rise when applying the monads.

### Generalized Probability Semantics

`HasEvalDist` now extends a `HasSupport` type-class and is extended by a `HasFinEvalDist` type-class that provide all the previous `OracleComp` specific things. All of these are special cases of having natural `MonadHom` into `Set`, `SPMF`, and `Finset` respectively, with additional compatibility conditions.

The classes for these extend each other in that order, and require compatibility. So e.g. the `support` must be exactly the set of things with positive output probability.

This allows for e.g. both computations with or without an `OptionT` or `ErrorT` to be used with these semantics. It also allows automatic `NeverFails` synthesis in many cases which can improve automation and be useful semantically in protocol definitions. Most monads should also implement `HasEvalDist.LawfulFailure` stating that failure is in fact mapped to always failing.

`HasEvalDist.hasPMF` extends with a `MonadHom` to `PMF` that is compatible with `evalDist`. This implicitly gives `NeverFails` instances.

If a monad `m` has these semantics then we also induce the natural semantics on `OptionT m` by sending `failure` to `failure` where appropriate. For the case of `support` specifically we also extend to `StateT m` by considering possible input states.

### Polynomial Functors

`OracleSpec` is now just an alias for `PFunctor`, essentially combining the oracle input and the oracle index together. Things like `DecidableEq` and `InhabitedRange` are moved down into the `PFunctor` context. `OracleComp` is now just an alias for the corresponding free monad.

`OracleQuery` is now defined to be `↑spec α`, e.g. viewing the polynomial functor spec as an actual functor. `query` is not an alias for `PFunctor.FreeM.liftPos`. Also add lemmas for the non-positive lifting in case it shows up.

This is more general in that the output type of an oracle may depend on the oracle's input and not just its index. It also gives some better type equalities that work better with `simp` and allows for more optimized execution.

Since it doesn't increase universe levels, it also allows for protocols where one party returns a continuation in the form of another monadic computation. This is especially useful in variable round protocols, and also avoids needing to carry internal state values around between a party's operations.

### Other Misc Changes

- Generalized `NeverFails` typeclass that should be more useful
- Modeling for basic crypto primitives (partially) adjusted to accommodate the foundational changes.
- Significant expansion of probability lemmas for `BitVec` and uniform selection from shifted intervals.
- Notation `$[m⋯n]` for selection from an arbitrary range. Similar to `xs[i]` notation it will attempt to auto-derive the non-emptiness of the range using a couple different tactics.

### Future Issues

- Evaluate moving "traversal" stuff to typeclasses similar to `HasEvalDist`. Many other monads have similar operations.
- Clean up `ToMathlib` which is becoming unwieldy and has many deprecated things (and also start PRs for more of them)
- Add instances and support for `ErrorT` and `EStateM`, similar to what we have for `OptionT` now. [This PR](https://github.com/leanprover-community/batteries/pull/1313) is relevant, and we need a similar one for lawful error monad behavior

### Breaking Changes

Note that some of these could be temporarily kept as a deprecation instead of fully swapped:

- `Support` and `FinSupport` are no longer namespaced, breaking dot notation
- `[= x | mx]` is deprecated for `Pr[=x | mx]`. Similar for other `EvalDist` things
- Basic induction principles don't have a `failure` case. The old versions still exist when `OptionT` is included
- Explicit universe level annotations will largely break due to polynomial functors
- Many `OracleSpec` things are moved down to the `PFunctor` level (and namespace)


quangvdao:
re: the `PFunctor` file getting long, you should merge the latest change on master which does precisely this breaking up

dtumad:
One thing that might still be needed for ArkLib is full sub-spec coercions. Even if not I plan to fix those today sometime

VCVio/OracleComp/SimSemantics/SimulateQ.lean
```diff
@@ -9,144 +9,221 @@ import ToMathlib.Control.StateT
...
+--     toFun mx := by
+--       have : n (Option α) := simulateQ impl (OptionT.run mx)
+--       sorry
+--     toFun_pure' := _
+--     toFun_bind' := _
+--   }
+--   simulateQ_liftM impl q := sorry
+
+/-- Simulate under a state transformer. -/
+instance [Monad m] [MonadLiftT (OracleQuery spec) m] [Monad n]
```
dtumad, line 160: Note that this type of thing shouldn't work with `StateT`, so we need something different for cases like this (Something like `liftTarget` or `liftM`). See the examples below, and the ones that break without this instance (if we even want to have it). I've had little luck trying to modify `HasSimulateQ` to handle this. cc @quangvdao 

On the other hand the `OracleComp`, `OptionT`, and `ErrorT` chaining cases are fine in all cases. These are probably the most relevant cases anyways in my opinion. In particular mixing and matching simulation with or without `OptionT` in a chain should work fine, and you can auto-deduce nice properties about failure odds after simulation still.

quangvdao: Yes, I think stateful reasoning will be very important for more complex proofs. I don't have any good intuition for what that would look like, perhaps mpl is the best example to copy?

dtumad:
Subsumed by additional changes in https://github.com/Verified-zkEVM/VCV-io/pull/90

PR #84, opened 2025-08-14
feat: tiny fix to v4.22.0 bump

quangvdao:
The workflow doesn't build all `ToMathlib` files, but I did, and noticed one thing breaking in `OrderEnrichedCategory`. So I fixed that.


alexanderlhicks:
huh interesting, it was fine locally :thinking:

PR #248, opened 2026-04-02
Add query-preserving monad morphisms

quangvdao:
## Summary
- add a small `QueryHom` layer for query-preserving monad morphisms and a helper notion for preserving `ProbComp` lifts
- add `QueryImpl.toMonadQuery` plus `SignatureAlg.map`/extensionality helpers for transporting generic constructions
- prove `FiatShamir.map_construction`, showing the generic Fiat-Shamir construction commutes with query-preserving semantics morphisms

## Verification
- `lake build VCVio.OracleComp.MonadQuery VCVio.CryptoFoundations.SignatureAlg VCVio.CryptoFoundations.FiatShamir`
- `lake build`

Disclosure: This PR body was posted by the assistant on behalf of Quang Dao using GPT-5.


VCVio/OracleComp/HasQuery.lean
```diff
@@ -24,13 +26,29 @@ standard transformer lifts such as `StateT`, `ReaderT`, `ExceptT`, and `WriterT`
 
 open OracleSpec
 
-universe u v w
+universe u v w x
 
 /-- Capability to issue queries to the oracle family `spec` inside the ambient monad `m`. -/
 class HasQuery {ι : Type u} (spec : OracleSpec.{u, v} ι) (m : Type v → Type w) where
   /-- Issue a single oracle query. -/
   query : (t : spec.Domain) → m (spec.Range t)
```
dtumad: Do you think maybe this should replace the current definition of `OracleQuery.query`? Thats already a derived definition so could just become a specific instance of this with the current definition as the instance definition. Would require bigger refactoring but feels very natural to me and conceivably could have better type inference especially when lifting is involved.

quangvdao: Yeah that makes sense, that's the direction I'm leaning toward as well, but I'm investigating to see what is the most ergonomic.

PR #265, opened 2026-04-07
Remove redundant `by exact` and `by rfl` tactic wrappers

SamuelSchlesinger:
Mechanical proof golf: replace `by exact term` with direct term-mode and `by rfl` with `rfl` where the goal is a definitional equality. No semantic changes.


dtumad:
LGTM, thanks!

PR #272, opened 2026-04-11
feat(SimSemantics): cross-type bind, OptionT/simulateQ, and Q-state invariant decomposition

XC0R:
## Summary

- `bind_eq_of_map_eq` — cross-type monadic bind composition via projection, generalizing `bind_congr` to different source types (ToMathlib/General.lean)
- `simulateQ_optionT_{bind,lift}` — `simulateQ` distributes through `OptionT` (SimSemantics/SimulateQ.lean)
- `QueryImpl.projectFst` + `simulateQ_run{,'}_eq_of_snd_invariant` — when the second component of product state is invariant under oracle queries, the full `simulateQ` decomposes into the projected single-component implementation (ProgramLogic/Relational/SimulateQ.lean)

### DIFF
--- a/VCVio/OracleComp/SimSemantics/SimulateQ.lean
+++ b/VCVio/OracleComp/SimSemantics/SimulateQ.lean
@@ -144,9 +144,8 @@ omit [LawfulMonad n] in
   rw [simulateQ_bind]
   exact bind_congr fun x => simulateQ_option_elim impl x my my'
 
-omit [LawfulMonad n] in
 /-- `simulateQ` distributes through `OptionT.bind`, stated via `OptionT.run`. -/
-lemma simulateQ_optionT_bind' [LawfulMonad n]
+lemma simulateQ_optionT_bind'
     (mx : OptionT (OracleComp spec) α) (f : α → OptionT (OracleComp spec) β) :
     simulateQ impl (mx >>= f).run =
     (simulateQ impl mx.run >>= fun a => simulateQ impl (f a).run : OptionT n β) := by
@@ -154,26 +153,23 @@ lemma simulateQ_optionT_bind' [LawfulMonad n]
   refine bind_congr fun x => ?_
   induction x <;> simp only [Option.elim_none, Option.elim_some, simulateQ_pure]
 
-omit [LawfulMonad n] in
 /-- `simulateQ` distributes through `OptionT.bind`, stated via `Option.elimM`. -/
-lemma simulateQ_optionT_bind'' [LawfulMonad n]
+lemma simulateQ_optionT_bind''
     (mx : OptionT (OracleComp spec) α) (f : α → OptionT (OracleComp spec) β) :
     simulateQ impl (mx >>= f).run =
     Option.elimM (simulateQ impl mx.run) (pure none) (fun a => simulateQ impl (f a).run) := by
   simp
 
-omit [LawfulMonad n] in
 /-- `simulateQ` distributes through `OptionT.bind`: the simulated OptionT-bind is the
     OptionT-bind of the simulated pieces. -/
-lemma simulateQ_optionT_bind [LawfulMonad n]
+lemma simulateQ_optionT_bind
     (mx : OptionT (OracleComp spec) α) (f : α → OptionT (OracleComp spec) β) :
     simulateQ impl (mx >>= f : OptionT (OracleComp spec) β) =
     (simulateQ impl mx >>= fun a => simulateQ impl (f a) : OptionT n β) := by
   apply simulateQ_optionT_bind'
 
-omit [LawfulMonad n] in
 /-- `simulateQ` commutes with `OptionT.lift`. -/
-@[simp] lemma simulateQ_optionT_lift [LawfulMonad n]
+@[simp] lemma simulateQ_optionT_lift
     (comp : OracleComp spec α) :
     simulateQ impl (OptionT.lift comp : OptionT (OracleComp spec) α) =
       (OptionT.lift (simulateQ impl comp) : OptionT n α) := by
--- a/VCVio/OracleComp/SimSemantics/SimulateQ.lean
+++ b/VCVio/OracleComp/SimSemantics/SimulateQ.lean
@@ -127,3 +127,34 @@ example (mx : OptionT (OracleComp spec₁) α)
   simulateQ impl₂ <| simulateQ impl₁ <| mx
 
 end tests
+
+section OptionT
+
+omit [LawfulMonad n] in
+/-- `simulateQ` distributes through `OptionT.bind`: the simulated OptionT-bind is the
+    OptionT-bind of the simulated pieces. -/
+lemma simulateQ_optionT_bind [LawfulMonad n]
+    (mx : OptionT (OracleComp spec) α) (f : α → OptionT (OracleComp spec) β) :
+    simulateQ impl (@Bind.bind (OptionT (OracleComp spec)) _ _ _ mx f) =
+    @Bind.bind (OptionT n) _ _ _
+      (simulateQ impl mx)
+      (fun a => simulateQ impl (f a)) := by
+  change simulateQ impl (@Bind.bind (OracleComp spec) _ _ _
+    mx (fun opt => match opt with | some a => f a | none => pure none)) = _
+  rw [simulateQ_bind]
+  change _ = @Bind.bind n _ _ _
+    (simulateQ impl mx) (fun opt => match opt with
+      | some a => simulateQ impl (f a) | none => pure none)
+  exact bind_congr fun opt => match opt with
+    | none => simulateQ_pure impl none
+    | some _ => rfl
+
+omit [LawfulMonad n] in
+/-- `simulateQ` commutes with `OptionT.lift`. -/
+@[simp] lemma simulateQ_optionT_lift [LawfulMonad n]
+    (comp : OracleComp spec α) :
+    simulateQ impl (OptionT.lift comp : OptionT (OracleComp spec) α) =
+      (OptionT.lift (simulateQ impl comp) : OptionT n α) := by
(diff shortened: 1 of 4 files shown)


VCVio/OracleComp/SimSemantics/SimulateQ.lean
```diff
@@ -127,3 +127,34 @@ example (mx : OptionT (OracleComp spec₁) α)
...
+
+omit [LawfulMonad n] in
+/-- `simulateQ` distributes through `OptionT.bind`: the simulated OptionT-bind is the
+    OptionT-bind of the simulated pieces. -/
+lemma simulateQ_optionT_bind [LawfulMonad n]
+    (mx : OptionT (OracleComp spec) α) (f : α → OptionT (OracleComp spec) β) :
+    simulateQ impl (@Bind.bind (OptionT (OracleComp spec)) _ _ _ mx f) =
+    @Bind.bind (OptionT n) _ _ _
+      (simulateQ impl mx)
+      (fun a => simulateQ impl (f a)) := by
```
quangvdao: I wonder if you could state this more nicely, why can't `mx >>= f` work here? what if you put `: OptionT (OracleComp spec) \alpha`?

VCVio/ProgramLogic/Relational/SimulateQ.lean
```diff
@@ -676,4 +676,82 @@ theorem tvDist_simulateQ_le_probEvent_bad_dist
...
+def QueryImpl.projectFst {ι : Type} {spec : OracleSpec ι} {σ Q : Type}
+    (impl : QueryImpl spec (StateT (σ × Q) ProbComp)) (q₀ : Q) :
+    QueryImpl spec (StateT σ ProbComp) :=
+  fun t => StateT.mk fun s => Prod.map id Prod.fst <$> (impl t).run (s, q₀)
+
+/-- Support-aware bind congruence for ProbComp: if two continuations agree on all
+    support elements, then the binds are equal. -/
+private theorem ProbComp.bind_congr_of_forall_mem_support
+    {α β : Type} (mx : ProbComp α) {f g : α → ProbComp β}
+    (h : ∀ x ∈ support mx, f x = g x) : mx >>= f = mx >>= g := by
```
quangvdao: This should be hoisted out to be a more general lemma, over all `[HasEvalSupport m]` or something (if it still holds there), and moved to the relevant file

VCVio/ProgramLogic/Relational/SimulateQ.lean
```diff
@@ -676,4 +676,82 @@ theorem tvDist_simulateQ_le_probEvent_bad_dist
...
   exact le_trans h_map h_tv_joint
 
+section SndInvariant
+
+/-- Project a product-state query implementation to the first component,
+    fixing the second component at a constant value. -/
+def QueryImpl.projectFst {ι : Type} {spec : OracleSpec ι} {σ Q : Type}
+    (impl : QueryImpl spec (StateT (σ × Q) ProbComp)) (q₀ : Q) :
+    QueryImpl spec (StateT σ ProbComp) :=
+  fun t => StateT.mk fun s => Prod.map id Prod.fst <$> (impl t).run (s, q₀)
```
quangvdao: this seems particular to the StateT monad transformer and also the choice of state on the left. can u tighten the name and/or provide alternative defs as well?

quangvdao (changes requested):

VCVio/OracleComp/SimSemantics/SimulateQ.lean
XC0R: Commentary heeded — statement now uses `(mx >>= f : OptionT (OracleComp spec) β)` and `(... : OptionT n β)` instead of explicit `@Bind.bind`.

VCVio/ProgramLogic/Relational/SimulateQ.lean
XC0R: Commentary heeded — renamed to `QueryImpl.fixSndStateT` to reflect StateT-specific, left-product-fixed semantics.

VCVio/ProgramLogic/Relational/SimulateQ.lean
XC0R: Commentary heeded — generalized from `ProbComp` to `OracleComp spec` and moved to `OracleComp.EvalDist` as public `OracleComp.bind_congr_of_forall_mem_support`. A further generalization over `[HasEvalSet m]` would require additional axioms beyond what the typeclass provides (support faithfulness for bind equality), so `OracleComp` with structural induction is the natural level.

VCVio/OracleComp/SimSemantics/SimulateQ.lean
dtumad: We aren't globally consistent about this, but using `OptionT.run` makes this a lot simpler (even though it doesn't *actually* do anything) since it makes the cast from `OptionT ProbComp A` to `ProbComp (Option A)` more explicit.

In particular you can avoid unfolding so many things:

```lean4
/-- `simulateQ` distributes through `OptionT.bind`: the simulated OptionT-bind is the
    OptionT-bind of the simulated pieces. -/
lemma simulateQ_optionT_bind' [LawfulMonad n]
    (mx : OptionT (OracleComp spec) α) (f : α → OptionT (OracleComp spec) β) :
    simulateQ impl (mx >>= f).run =
    (simulateQ impl mx.run >>= fun a => simulateQ impl (f a).run : OptionT n β) := by
  rw [OptionT.run_bind, Option.elimM, simulateQ_bind]
  refine bind_congr fun x => ?_
  induction x <;> simp only [Option.elim_none, Option.elim_some, simulateQ_pure]

lemma simulateQ_optionT_bind [LawfulMonad n]
    (mx : OptionT (OracleComp spec) α) (f : α → OptionT (OracleComp spec) β) :
    simulateQ impl (mx >>= f : OptionT (OracleComp spec) β) =
    (simulateQ impl mx >>= fun a => simulateQ impl (f a) : OptionT n β) := by
  apply simulateQ_optionT_bind'
``` 

Maybe it is good to have both versions, since people don't consistently use / not use `OptionT.run` when defining things?

dtumad: I'd argue it might make most sense to simplify things by preserving `Option.elimM` rather than unfolding everything:

```lean4
omit [LawfulMonad n] in
@[simp] lemma simulateQ_option_elim (x : Option α) (my : OracleComp spec β)
    (my' : α → OracleComp spec β) : simulateQ impl (x.elim my my') =
    x.elim (simulateQ impl my) (fun x => simulateQ impl (my' x)) := by
  cases x <;> simp

@[simp] lemma simulateQ_option_elimM (mx : OracleComp spec (Option α)) (my : OracleComp spec β)
    (my' : α → OracleComp spec β) : simulateQ impl (Option.elimM mx my my') =
    Option.elimM (simulateQ impl mx) (simulateQ impl my) (fun x => simulateQ impl (my' x)) := by
  simp [Option.elimM]

lemma simulateQ_optionT_bind''
    (mx : OptionT (OracleComp spec) α) (f : α → OptionT (OracleComp spec) β) :
    simulateQ impl (mx >>= f).run =
    Option.elimM (simulateQ impl mx.run) (pure none) (fun a => simulateQ impl (f a).run) := by
  simp
```

but again I don't think we have much consistency globally, so would be good to just include all the possible versions.

XC0R:
Addressed both suggestions from @dtumad:

1. Added `simulateQ_optionT_bind'` (OptionT.run form) and simplified `simulateQ_optionT_bind` to delegate via `apply simulateQ_optionT_bind'`. Both versions kept.

2. Added `simulateQ_option_elim`, `simulateQ_option_elimM` (both `@[simp]`), and `simulateQ_optionT_bind''` (Option.elimM form). Minor adjustment: `simulateQ_option_elimM` proof needed `unfold Option.elimM; rw [simulateQ_bind]; exact bind_congr ...` rather than `simp [Option.elimM]` since the bind doesn't reduce under simp alone.

dtumad (approved):
My only other other comment is that `omit` and re-adding the lawful monad instances seems a bit strange. Other than that LGTM

XC0R:
Removed the redundant omit/re-add pattern — section variable now provides LawfulMonad n directly.

VCVio/ProgramLogic/Relational/SimulateQ.lean
XC0R: Clarification on the generalization boundary: the proof uses `OracleComp.inductionOn` — in the `query_bind` case, all oracle outputs are in support, so the inductive step reduces via unconditional `bind_congr`. This structural recursion on the free monad is what makes term-level equality (`mx >>= f = mx >>= g`) provable. A generic `[HasEvalSet m]` version would need an injectivity axiom that the typeclass doesn't provide (the `toSPMF` monad hom doesn't reflect equality back to `m`). The distributional analogues (`probOutput_bind_congr` etc. in `EvalDist/Monad/Basic.lean`) already work at that general level.

dtumad: That makes sense, I think this can be merged now as is.

XC0R:
Addressing @quangvdao's earlier comment on generalizing `bind_congr_of_forall_mem_support` beyond `OracleComp` more thoroughly:

The proof uses `OracleComp.inductionOn` — in the `query_bind` case, all oracle outputs are in support, so the inductive step reduces via unconditional `bind_congr`. This structural recursion on the free monad is what makes term-level equality (`mx >>= f = mx >>= g`) provable. A generic `[HasEvalSet m]` version would need an injectivity axiom that the typeclass doesn't provide (the `toSPMF` monad hom doesn't reflect equality back to `m`). The distributional analogues (`probOutput_bind_congr` etc. in `EvalDist/Monad/Basic.lean`) work at that general level.

dtumad:
I think it does maybe make sense to add an appropriate typeclass to get a general set version as well, but I can do that in another PR, I don't think it makes sense to tack that on here.

PR #303, opened 2026-04-19
feat(VCVio): update constructions in `QueryTracking` to use `OracleSpec` namespace

dtumad:
This PR moves definitions like `loggingOracle` into the `OracleSpec` namespace, since they already depend on them implicitly and they are defined at root level. This allows for simpler dot notation in many cases where the specification needs to be made explicit for type checking reasons.


quangvdao (approved):

dtumad:
Marking WIP until build time fixes

PR #357, opened 2026-04-27
chore: move Merkle tree files

BoltonBailey:
I am trying to port over some of my changes from https://github.com/Verified-zkEVM/ArkLib/pull/144

To start with, this change moves Merkle tree content to dedicated folder, with subfolders for different versions with different tree structure. It also moves the completeness theorems on the inductive side into its own file separate from the the defs, in anticipation of more different files for the main theorems about security properties.


dtumad (approved):
Looks good to me, thanks. By the way, the question a couple weeks ago about `IsQueryBound` led to a definition `IsQueryBoundP` that may be more useful for this use case if you still need that functionality.

PR #361, opened 2026-04-28
feat: close some misc `sorry` calls

dtumad:
Closes some of the simpler `sorry` calls left in the library and removes some unneeded print statements.


ToMathlib/OrderEnrichedCategory.lean
```diff
@@ -71,8 +71,7 @@ variable (C : Type u₁) (D : Type u₂) [𝒞 : EnrichedCategory V C] [𝒟 : E
 @[simps]
 instance instProduct : EnrichedCategory V (C × D) where
   Hom X Y := (𝒞.Hom X.1 Y.1) ⊗ (𝒟.Hom X.2 Y.2)
-  id X := (λ_ _).inv ≫ (sorry)
-  -- ((𝒞.id X.1) ⊗ (𝒟.id X.2))
+  id X := (λ_ _).inv ≫ ((𝒞.id X.1) ⊗ₘ (𝒟.id X.2))
```
dtumad, line 74: @quangvdao is this the expected implementation? Not sure why it's commented out still

quangvdao: I think so. maybe there was some elaboration issue

PR #370, opened 2026-05-04
feat: prove unlinkBadExp_le_sessionCollisionBound

### LINKED ISSUE #206: Proof obligation for `unlinkBadExp_le_sessionCollisionBound` in `Examples/PRFTagReader.lean`
A proof in `Examples/PRFTagReader.lean` contains a `sorry`.

[Link to the sorry on GitHub](https://github.com/Verified-zkEVM/VCV-io/blob/master/Examples/PRFTagReader.lean#L524)

**Code Snippet:**
```lean
              (Digest := Digest) (sessionsPerTag := sessionsPerTag) adversary]).toReal := by
  sorry

/-- A pointwise bound on the nonce sampler turns the bad-event probability into an explicit session
collision bound. -/
theorem unlinkBadExp_le_sessionCollisionBound
```

jCabala:
### Summary

Proves `unlinkBadExp_le_sessionCollisionBound` (closes #206) by induction on the adversary's oracle-computation tree. Introduces some supporting lemmas and definitions: state invariants (`unlinkBadRemaining`, `unlinkBadCacheBounded`), preservation lemmas for tag and reader steps, and the core union-bound lemma `simulateQ_unlinkBad_prob_le`, which yields the explicit bound `sessionsPerTag² · |TagId| · maxNonceProb`.

### DIFF
--- a/Examples/PRFTagReader.lean
+++ b/Examples/PRFTagReader.lean
@@ -628,6 +628,394 @@ theorem unlinkabilityAdvantage_le_two_prf_plus_collision
               (Digest := Digest) (sessionsPerTag := sessionsPerTag) adversary]).toReal := by
   sorry
 
+/-- The number of still-available successful tag sessions in a bad-event state. -/
+private def unlinkBadRemaining (st : UnlinkBadState TagId Nonce Digest) : ℕ :=
+  (Finset.univ : Finset TagId).sum fun tag => sessionsPerTag - st.sessionsUsed tag
+
+/-- Reachable bad-event states only cache nonces that came from successful tag sessions. For each
+tag, we retain a finite witness set of cached nonces whose size is bounded by that tag's session
+counter. -/
+private def unlinkBadCacheBounded (st : UnlinkBadState TagId Nonce Digest) : Prop :=
+  ∀ tag : TagId, ∃ nonces : Finset Nonce,
+    nonces.card ≤ st.sessionsUsed tag ∧
+      ∀ nonce : Nonce, (st.responses (tag, nonce)).isSome = true → nonce ∈ nonces
+
+/-- State produced by a successful `RF_bad` tag query after sampling `nonce` and `auth`. -/
+private def unlinkBadTagNext
+    (tag : TagId) (st : UnlinkBadState TagId Nonce Digest)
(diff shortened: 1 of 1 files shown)


Examples/PRFTagReader.lean
```diff
@@ -638,7 +1026,46 @@ theorem unlinkBadExp_le_sessionCollisionBound
     (Pr[= true | unlinkBadExp (TagId := TagId) (Nonce := Nonce)
       (Digest := Digest) (sessionsPerTag := sessionsPerTag) adversary]).toReal ≤
       ((sessionsPerTag ^ 2 * Fintype.card TagId : ℕ) : ℝ) * maxNonceProb := by
```
dtumad, line 1028: Since `Pr[= nonce | ($ᵗ Nonce)]` should always just be `1 / Fintype.card Nonce`, it might make sense to also include also a corollary that doesn't take in a `maxNonceProb` or `hmax` and just proves the tightest possible bound as being that, although not really that different in practice.

Other than that looks good to me.

dtumad (approved):

Examples/PRFTagReader.lean
jCabala: Is it true that `Pr[= nonce | ($ᵗ Nonce)]` is always uniform? Where do we make this assumption?

dtumad: https://github.com/Verified-zkEVM/VCV-io/blob/97292ea9c11d1438bb053ecc99c4983f837a251b/VCVio/OracleComp/Constructions/SampleableType.lean#L48

It follows from the definition of `SampleableType` here.

dtumad: Although maybe the original intention was that a non-uniform nonce generator could be swapped in here and the bound (maybe) still holds in the more general form.

jCabala: Nice, I haven't seen that lemma. Then I guess the colorarly should be on the main theorem (`theorem unlinkabilityAdvantage_le_two_prf_plus_sessionCollisionBo`) level I guess?

dtumad: Yeah just having the "tightest possible" bound on the final statement seems fine to me.

Issue #377, opened 2026-05-06
Regression in Merkle Tree completeness proof


BoltonBailey:
There seems to be some kind of problem with the Inductive Merkle tree completeness proof. Going back to when this lived in ArkLib, the original proof of the final theorem was a dozen lines and depended directly on the functional lemma immediately above it. Then [this ArkLib PR](https://github.com/Verified-zkEVM/ArkLib/commit/263def1fb7a73b9d2c65e0d38b3a744fa5396ae6#diff-fd1cf2198efca76fd3cc8649eedea5e4bcd08a014e6e84933e38d9556bdf22e9) redefined the lemma and changed it to a two line grind call. In fact `rw [HasEvalSPMF.neverFail_iff, HasEvalPMF.probFailure_eq_zero]` seems to work, which indicates to me that the redefinition has somehow turned the theorem into a triviality.

dtumad:
Is this the theorem on non-failure? `OracleComp` has no failure by construction now so `neverFails` should be trivial unless you have `OptionT` or `ErrorT` in the final computation failure is prevented at the type level

BoltonBailey:
Yes, I think this is what is causing the theorem to now be trivial.

BoltonBailey:
The diff above also changed the type of `verifyProof` from ` OracleComp (spec α) Unit` to ` OptionT (OracleComp (spec α)) Unit`, now it seems it is `OptionT m Unit`. I guess one of those changes affects the meaning of the `guard` as no longer carrying through to failure?

dtumad:
Ah I see now, the `simulateQ` in the completeness theorem is unfolding the optionT and making this a lemma about `OracleComp _ (Option Unit)` which has the never fails instance. So it probably needs `OptionT.mapM` or something like that so you simulate under the transformer and have a statement about `OptionT (OracleComp _ Unit).

dtumad:
I think the conversion assumed we would land monad polymorphism for `simulateQ` which would have avoided needing the mapping layer for `OptionT`

BoltonBailey:
I thought that the `completeness` theorem had been sorry-free at some point, but I guess I was mistaken about that - it looks like it always depended on these `runWithOracle` methods that relate the non-monadic execution to the monadic execution, but it seems like these methods never made it to VCVio.

BoltonBailey:
#378 I think should at least make the statement meaningful.

BoltonBailey:
Fixed by #378

PR #378, opened 2026-05-06
fix(MerkleTree): fix completeness theorem statements

BoltonBailey:
See #377. In the process of upstreaming from ArkLib and other upgrades, the statement of the main monadic completeness theorem for Merkle trees was accidentally turned into a trivial statement saying a computation never failed, when in fact it never failed by definition, rather than due to the correctness of the implementation.

This PR is to fix this. It redefines the proof verification procedure to return a boolean and the main Merkle tree completeness theorem to assert that this boolean is true. It then also integrates and unsorries the background monad infrastructure to prove the main theorem from the non-monadic version

AI summary below:

## Summary

- Adds `OracleComp.runWithOracle` (a deterministic interpretation of an `OracleComp` against an answer-function `f : spec.FunctionType`) and `OracleSpec.QueryCache.containedIn` for cache-compatibility.
- Proves the bridge lemma `probEvent_eq_one_simulateQ_randomOracle_run_iff`: `Pr[fun v => p v.1 | (simulateQ randomOracle oa).run cache] = 1 ↔ ∀ f extending cache, p (runWithOracle f oa)`. The proof goes through a structural support-characterization lemma `exists_containedIn_runWithOracle_eq_iff_mem_support` plus a `NeverFail` lemma for the random-oracle simulation.
- Uses the bridge to prove `InductiveMerkleTree.completeness` as the strong probability-1 statement `Pr[verify = true | …] = 1` (rather than just `NeverFail`).
- Switches `verifyProof` from `OptionT m Unit` (failing via `guard`) to `m Bool` so the verifier output is observable. Without this, completeness can only be stated as `NeverFail`, conflating "always returns" with "always returns true".

## Background

This ports infrastructure originally drafted in [ArkLib's `ToVCVio/Oracle.lean` (commit 9a06cc55)](https://github.com/Verified-zkEVM/ArkLib/blob/9a06cc55/ArkLib/ToVCVio/Oracle.lean) — where the bridge lemmas were left as `sorry` and later removed entirely. The version here adapts to current VCVio's API (no built-in `OracleComp` failure; queries indexed by `spec.Domain` with dependent `spec.Range t`) and provides full proofs.

## Files

- New: `VCVio/OracleComp/SimSemantics/RunWithOracle.lean`
- Modified: `VCVio/CryptoFoundations/MerkleTree/Inductive/Defs.lean` (`verifyProof` returns `Bool`)
- Modified: `VCVio/CryptoFoundations/MerkleTree/Inductive/Completeness.lean` (`completeness` proves `Pr[…] = 1`)
- Modified: `VCVio.lean` (registers the new module)

## Test plan

- [x] `lake build VCVio.OracleComp.SimSemantics.RunWithOracle` — clean
- [x] `lake build VCVio.CryptoFoundations.MerkleTree.Inductive.Completeness` — clean
- [x] No `sorry`s introduced; the two ArkLib parity lemmas (`randomOracle_*_neverFails_iff_runWithOracle_neverFails`) are proven as trivial corollaries of the new infrastructure.


quangvdao:
Posted by Cursor assistant (model: GPT-5) on behalf of the user (Quang Dao) with approval.

I pushed a follow-up cleanup commit (`7ff5ca83`) to reorganize the new random-oracle bridge API and avoid the misleading `runWithOracle` naming.

Summary of changes:

- Removed `VCVio/OracleComp/SimSemantics/RunWithOracle.lean`.
- Added `OracleComp.evalWithAnswerFn` in `VCVio/OracleComp/SimSemantics/SimulateQ.lean` as the `Id`-valued specialization of `simulateQ`.
- Added `QueryCache.AgreesWithFn` in `VCVio/OracleComp/QueryTracking/Structures.lean`, with the cache argument last so dot notation works naturally as `cache.AgreesWithFn f`.
- Moved the random-oracle support/probability-one bridge lemmas into `VCVio/OracleComp/QueryTracking/RandomOracle/Simulation.lean`, which is a better conceptual home for them.
- Dropped the vacuous `randomOracle_*_neverFails_iff_*` parity lemmas. Since plain `OracleComp` has no failure, those statements risk preserving the old conceptual trap this PR is fixing.
- Updated the Merkle completeness proofs to use the relocated bridge and the `QueryImpl spec Id` answer-function surface directly.

Verification: `lake build VCVio` passed locally after the cleanup. It replayed existing unrelated warnings from other modules, but the new/changed modules build cleanly.

VCVio/CryptoFoundations/MerkleTree/Vector/Basic.lean
```diff
@@ -209,12 +210,12 @@ def getPutativeRoot {m : Type _ → Type _} [Monad m] [MonadLiftT (OracleQuery (
 /-- Verify a Merkle proof `proof` that a given `leaf` at index `i` is in the Merkle tree with given
   `root`.
   Works by computing the putative root based on the branch, and comparing that to the actual root.
-  Outputs `failure` if the proof is invalid. -/
+  Returns `true` if the proof is valid, and `false` otherwise. -/
 def verifyProof {m : Type _ → Type _} [Monad m] [MonadLiftT (OracleQuery (spec α)) m]
```
dtumad: Not sure if this is the right place for this, but also might be good to adapt this to use `HasQuery` rather than a manual lift? Avoids needing to manage the lifts manually.

dtumad: Could reasonably be part of a broader PR focused on standardizing that instead

BoltonBailey: I'll try it out!

PR #384, opened 2026-05-09
feat(MerkleTree): Added Binding proof

vbuterin:
Adding a proof that Merkle branches are binding (ie. a collision in a Merkle branch implies a collision in the underlying hash)

(EDIT: the statement now does not require sister nodes to be the same)


dtumad:
Besides the sister node thing, I'd probably suggest not currying and instead writing the hash as having type `H : α → α → α` instead of taking a pair as input. You can pass `H` as is to `getPutativeRootWithHash` then, and curried functions just tend to work worse in larger Lean proofs because tactics often won't reduce them without first pattern matching the input into an explicit pair.

quangvdao:
This is sane, but the right next generalization is to let the two openings use different sibling proofs.

Concretely, I’d change the statement from one shared proof to two proofs:

```lean
theorem getPutativeRootWithHash_binding
    (hash : α → α → α)
    {s : Skeleton} (idx : SkeletonLeafIndex s)
    (proof₁ proof₂ : List.Vector α idx.depth)
    (x y : α)
    (hne : x ≠ y)
    (heq : getPutativeRootWithHash idx x proof₁ hash
         = getPutativeRootWithHash idx y proof₂ hash) :
    ∃ l₁ r₁ l₂ r₂,
      (l₁, r₁) ≠ (l₂, r₂) ∧ hash l₁ r₁ = hash l₂ r₂
```

The induction barely changes. At each level, compare the full compression inputs:

```lean
-- left branch
(subL, proof₁.head) =? (subR, proof₂.head)

-- right branch
(proof₁.head, subL) =? (proof₂.head, subR)
```

If those pairs are equal, recurse using the child equality. If not, the current level gives the collision.

I’d also make `Collision` take the curried hash type `hash : α → α → α`, matching `getPutativeRootWithHash`, instead of wrapping through `α × α → α`.

Posted by Codex assistant (model: GPT-5.5) on behalf of the user (Quang Dao) with approval.

vbuterin:
Thank you, this worked!

quangvdao (approved):

Issue #205, opened 2026-04-01
Proof obligation for `unlinkabilityAdvantage_le_two_prf_plus_collision` in `Examples/PRFTagReader.lean`


quangvdao:
A proof in `Examples/PRFTagReader.lean` contains a `sorry`. Note the scope of this is somewhat increased by the weakening of the statement in #203.

**Goal:** Replace the `sorry` with a complete proof.

[Link to the sorry on GitHub](https://github.com/Verified-zkEVM/VCV-io/blob/master/Examples/PRFTagReader.lean#L509)

**Code Snippet:**
```lean
  sorry

/-- Unlinkability reduction statement: the multiple-vs-single gap is bounded by one PRF advantage
for the multiple-session world, one PRF advantage for the single-session world, and the bad-event
probability from the intermediate collision world. -/
theorem unlinkabilityAdvantage_le_two_prf_plus_collision
```

PR #399, opened 2026-05-15
Close(?) #203: prove the authentication PRF reduction (with corrected statement)

### LINKED ISSUE #203: Proof obligation for `authExp_le_prfAdvantage_add_authIdeal` in `Examples/PRFTagReader.lean`
A proof in `Examples/PRFTagReader.lean` contains a `sorry`.

[Link to the sorry on GitHub](https://github.com/Verified-zkEVM/VCV-io/blob/master/Examples/PRFTagReader.lean#L488)

**Code Snippet:**
```lean
  [DecidableEq Digest] [SampleableType Digest]
  {sessionsPerTag : ℕ} [Fact (0 < sessionsPerTag)]

/-- Authentication reduction statement: the real game is bounded by a PRF distinguishing advantage
plus the success probability in the ideal authentication world. -/
theorem authExp_le_prfAdvantage_add_authIdeal
```

alik-eth:
## Summary

Part of #203. Removes the unprovable `sorry` in the authentication PRF reduction and lands the reduction infrastructure. **Does not yet deliver a closed-form security bound** — see *Scope* below.

## The original statement was false

The `sorry`'d `authExp_le_prfAdvantage_add_authIdeal` claimed `authExp ≤ prfAdvantage + authIdealExp`. Since `authIdealExp = 0` (`authIdealExp_eq_zero`), it collapses to `authExp ≤ prfAdvantage` — which is **false**, not merely unproven: for a perfect-PRF instance `prfAdvantage` of the reduction is `0` while `authExp` is still positive (the adversary forges against a random function with probability ~`|TagId|/|Digest|`). No honest proof can keep `authIdealExp` on the RHS, so the original declaration cannot be salvaged.

## What this PR delivers

- **Reduction infrastructure** — `authToPRF{Tag,Reader,}QueryImpl`, `authToPRFReduction`, and the inductive `simulateQ` equivalence lemmas (`simulateQ_prfReal_authToPRF*_run`).
- **The genuine result** — `prfRealExp_authToPRFReduction_eq_authExp`: the reduction's real PRF experiment reproduces `authExp` exactly. This is the load-bearing lemma.
- **A provisional bound** — `authExp_le_prfAdvantage_add_authRF` and its existential form. Honest disclosure: `authRFExp` is *defined* as `prfIdealExp (authToPRFReduction adversary)`, so this theorem is the generic PRF triangle inequality wrapped around the reduction equality. It is true and axiom-clean but **not yet a security bound** — `authRFExp` is left as an opaque term.

## Scope — what is *not* done

The cryptographically meaningful statement #203 wants is a bound on `authRFExp` by a concrete collision probability. That is **out of scope here** and tracked as a follow-up (`feat/authrf-collision-bound`). This PR therefore does **not** close #203; it removes the false `sorry` and lays the groundwork.

`#205`'s `sorry` in `unlinkabilityAdvantage_le_two_prf_plus_collision` is untouched. Auth-only: no `VCVio/` changes.

## Verification

- `lake env lean Examples/PRFTagReader.lean` builds; the one remaining `sorry` is the pre-existing #205 obligation.
- `#print axioms` on both auth theorems → `propext, Classical.choice, Quot.sound`. No `sorryAx`.

## Review ask

Confirm the approach: (1) deleting the false `authIdealExp`-RHS theorem rather than aliasing its name, and (2) landing the reduction infrastructure now with the concrete `authRFExp` bound deferred to a follow-up PR that actually closes #203.


dtumad (approved):
> Confirm the approach: (1) deleting the false authIdealExp-RHS theorem rather than aliasing its name, and (2) landing the reduction infrastructure now with the concrete authRFExp bound deferred to a follow-up PR that actually closes https://github.com/Verified-zkEVM/VCV-io/issues/203.

This all seems reasonable to me. It might be nice to add the `sorry` version with more concrete bounds as well to point #203 at, but I can also just update the issue or treat that as part of #205 instead.

PR #402, opened 2026-05-17
feat: replace `HasEvalSPMF`/`HasEvalSet` with built-in monad lift machinery

dtumad:
This PR replaces some custom type-classes with more primitive definitions to make them more broadly applicable, and allows them to function better with existing monad lift infrastructure:
- `HasEvalSet` -> `MonadLiftT _ SetM` + `LawfulMonadLiftT _ SetM`
- `HasEvalSPMF` -> `MonadLiftT _ SPMF` + `LawfulMonadLiftT _ SPMF`
- `HasEvalPMF` -> `MonadLiftT _ PMF` + `LawfulMonadLiftT _ PMF`

Also adds a bare-bones `EvalDistCompatible` instance connecting support to distribution probability, which allows us to avoid introducing any diamonds with the conversion.

This should make it possible to implement `MonadLiftT (OracleQuery spec) PMF` in any way you like for a given `spec`, and then use probability notation in the expected way for those semantics.


dtumad:
Subsumed into #419

PR #405, opened 2026-05-24
refactor(MLDSA/Concrete/Rounding): remove intToCoeff; simplify proofs

ChihChengLiang:
## Summary

Cleans up `LatticeCrypto/MLDSA/Concrete/Rounding.lean` in three ways:

**Remove aliases and thin wrappers.**
- Drop `intToCoeff` (`def intToCoeff (z : ℤ) : Coeff := z` — a redundant name for the coercion `ℤ → Coeff`) and replace every call site with a direct cast.
- Drop `centeredRepr_eq_of_natAbs_le` (alias for `centeredRepr_intCast_eq_of_natAbs_le`).
- Remove `coeffRing_sub_eq`, `coeffRing_add_eq`, `gamma2_half`, `twoGamma_isUnit_of_isApproved`, and `power2RoundLow_centeredRepr`.
- Remove 8 public thin-wrapper theorems (`*_field_of_isApproved`, `concreteRounding_high_low_decomp_of_isApproved`, `concreteRounding_lowBits_bound_of_isApproved`, `concretePower2Round_bound_field`) that simply forwarded to the already-public general versions. The `Laws` structures are now built directly.

**Restructure private proof infrastructure.**
- Add `coeff_cast_div_add_mod` to avoid repeating the div-add-mod identity in multiple proofs.
- Enrich `BalancedDecomp` with `hγ`, `h2α`, `hγ2`, `hmdef`, `hunit` fields, absorbing `gamma2_half` and `twoGamma_isUnit_of_isApproved`.
- Change private theorem signatures to take `alpha` and `m` directly via `ctx : BalancedDecomp alpha m` instead of threading `p : Params` through every lemma.
- Consolidate `highBitsCoeff_of_nonneg_val` + `highBitsCoeff_of_neg_val` into a single `highBitsCoeff_of_euclid`.

**Simplify proof bodies.**
- Replace multi-step manual rewrites with `simp`, `push_cast`, `ring_nf`, and `omega` throughout.

### DIFF
--- a/LatticeCrypto/MLDSA/Concrete/Rounding.lean
+++ b/LatticeCrypto/MLDSA/Concrete/Rounding.lean
@@ -158,7 +158,6 @@ private structure BalancedDecomp (alpha m : ℕ) : Prop where
   hα   : 0 < alpha
   hγ   : 0 < alpha / 2
   h2α  : 2 * (alpha / 2) = alpha
-  hγ2  : 2 * (alpha / 2) < modulus
   hm   : 0 < m
   hmdef: (modulus - 1) / alpha = m
   hqm1 : alpha * m = modulus - 1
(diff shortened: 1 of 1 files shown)


LatticeCrypto/MLDSA/Concrete/Rounding.lean
```diff
@@ -157,163 +154,145 @@ def hintWeight (h : Hint) : ℕ :=
...
-
-@[simp] theorem coeffRing_add_eq (r s : Rq) : coeffRing.add r s = r + s := by
-  ext i; simp only [NegacyclicRing.coeff_add, coeffRing.add_coeff]
-
 private structure BalancedDecomp (alpha m : ℕ) : Prop where
   hα   : 0 < alpha
+  hγ   : 0 < alpha / 2
+  h2leα: 2 ≤ alpha
+  h2α  : 2 * (alpha / 2) = alpha
+  hγ2  : 2 * (alpha / 2) < modulus
```
dtumad: Do most of these properties not follow from `hα` and `Even α`? I don't know this part of the codebase that well but it seems like most of these fields could be implemented as lemmas on a more basic structure.

ChihChengLiang: Good point. I've figured out a way to keep the parameter facts minimum. Derived facts are added as private lemmas.

PR #406, opened 2026-05-24
feat(ToMathlib): codensity monad with LawfulMonad and LawfulMonadLift

MavenRain:


dtumad:
I think this probably makes more sense as a PR to https://github.com/Verified-zkEVM/PolyFun, it is a dependency now that we can update pretty easily if you want to use the definition for something here.

MavenRain:
Closing this, for now . . . may revisit later

PR #418, opened 2026-05-30
feat: generalize monad in SigmaProtocol

BoltonBailey:
This PR generalizes the monad in ChallengeVerifyProtocol/SigmaProtocol from ProbComp to any monad `m`, and adds a field for a monadic challenge sampling procedure.

This is part of an attempt to generalize ChallengeVerifyProtocol enough to express the Kilian transformation as returning a ChallengeVerifyProtocol.


quangvdao (approved):
Reviewed at `ef7b93ab5ebd4610856d0ae503f0703ca1161196` against current `main`.

No blocking findings. The monad parameter is threaded consistently through `ChallengeVerifyProtocol` / `SigmaProtocol`; `PerfectlyComplete` is deliberately strengthened to the pointwise-challenge statement, and `PerfectlyComplete.probOutput_uniform_challenge_eq_one` recovers the previous sampled form for `ProbComp`. The Fiat–Shamir constructor lifts the sigma protocol's own monad into the ambient RO monad, while the existing proof developments that need pure probability remain explicitly specialized to `ProbComp`. Schnorr and the Fischlin/Fiat–Shamir consumers were updated coherently.

I found no added `sorry`, `admit`, axiom, unsafe/native trust, or linter suppression in the diff. Exact-head hosted CI is fully green, including the project build, downstream-consumer build, imports, style, isolation, and axiom gates. Ready to merge.

Non-blocking follow-up: when the first non-`ProbComp` sigma protocol lands, bank a small consumer canary that instantiates the generalized constructor through its oracle lift; this PR proves the API shape but its in-tree concrete example remains Schnorr over `ProbComp`.

dtumad:
Deep semantic audit across the open PR set (semantics only). The direction — a monad-parametric `SigmaProtocol` so provers can query oracles — is one we want, and the `ChallengeVerifyProtocol`-compatible re-cut is welcome. But as stated the generalization makes the security notions *unfalsifiable* at exactly the instantiations that motivate it:

1. **BLOCKING: no lawfulness on the semantic lifts.** The probability-bearing notions are stated via `[MonadLiftT m SPMF]` and the support-bearing ones via `[MonadLiftT m SetM]`, with no `LawfulMonadLiftT` on either and no `EvalDistCompatible m` tying the two routes together (the diff has zero occurrences of any of the three). Consequences: without `LawfulMonadLiftT m SPMF`, `Pr[= true | commit >>= respond >>= …] = 1` does not decompose — `PerfectlyComplete`/`HVZK` become statements about an uninterpreted function of the composite; and without `EvalDistCompatible`, `SpeciallySound` (via `SetM`) and completeness (via `SPMF`) can be about *different semantics of the same protocol* — a scheme could be "complete" in one lift and "sound" in another with no connection. #450 gets this signature right (`[LawfulMonadLiftT m SPMF] [LawfulMonadLiftT m SetM] [EvalDistCompatible m]`); please match it on every probability- or support-bearing definition.
2. **BLOCKING (semantic hazard for the stated motivation): the docstring's target instantiation is the repaired-away ROM semantics.** "a general `m` lets the prover additionally query oracles (e.g. a hash oracle for the Kilian transform)" — but `OracleComp spec` acquires its probability denotation through fresh-independent-response-per-query semantics, exactly what #558 removed as unsound for a shared random oracle (equal hash inputs may get different answers). An oracle-parametric Σ-protocol at a ROM spec needs the caching/shared-ROM layer, not a bare lift. Either scope the docstring to oracles where independent responses are the intended semantics, or route the ROM case through `withCacheOverlay` explicitly.
3. **The `PerfectlyComplete` change is a silent public-definition strengthening.** The pointwise-`∀ ω` form is arguably the better definition (and the docstring's equivalence argument is right *at `m := ProbComp`* with a uniform challenge over a finite type), but only one direction is proved (`probOutput_uniform_challenge_eq_one`) and only at `ProbComp`. Split this into its own commit with the reverse implication proved at `ProbComp`, so downstream consumers (Fischlin, FiatShamir/Sigma) inherit a documented equivalence, not a drive-by meaning change.
4. **The generalization should pay rent before landing**: there is no `m ≠ ProbComp` instantiation anywhere in the PR. Given item 2, the natural renter is not yet available — which suggests either landing this together with one concrete non-`ProbComp` consumer, or waiting until the caching layer gives you one.

Bookkeeping: the diff still states `PerfectHVZK` over `𝒟[·]`; on current main `𝒟[·]` is a `Measure` and the SPMF surface is `𝒮[·]` — restate on rebase. And coordinate sequencing with #450 (both rewrite `SigmaProtocol.lean`; this PR should land first once the above are fixed, and #450 rebases onto it).

BoltonBailey:
I see there is now some kind of blocker related to clearing the library of SPMF/PMF uses? I have asked Claude to fix it but if there is some library-wide pass that needs to be done before merging this, that would be OK.

dtumad:
I think the script *should* be set up so that even if you add more uses it will flag that but not actually prevent merging. Ideally switching to `Measure` proofs would be good but also I think we have a few major Lean versions still to make the cutover fully.

dtumad:
> I see there is now some kind of blocker related to clearing the library of SPMF/PMF uses? I have asked Claude to fix it but if there is some library-wide pass that needs to be done before merging this, that would be OK.

~Ah I guess the warning budget now increases since `PMF`/`SPMF` are being deprecated? I thought my scripting would allow old uses to stay but I guess it may get weird in places that just add a few additional uses.~ Ideally things would convert along the lines of this sequence: #548 and use `Measure` and the new mathlib core definitions wherever possible and avoid bridging through the old discrete versions.

Seems like maybe this is just the script running wrong? I'll investigate more

dtumad:
Follow-up after checking both CI failures: the earlier PMF/SPMF boundary failure at 5ee728b8 was genuine (18 direct occurrences in SigmaProtocol against a zero baseline). Commit 3d634166 resolved it by making the properties Measure-native and isolating five compatibility references; the branch is now net -13 PMF/SPMF occurrences versus main. The remaining red Build Project check was unrelated: Mathlib reported a whitespace warning in SigmaProtocol.lean (Pr[fun versus Pr[ fun). I pushed d1b7c959 and merged current main. The focused file compile, PMF boundary fixtures/ceiling/report, full local lake build, and the complete hosted CI suite now all pass.

PR #427, opened 2026-06-06
Bump to Lean/Mathlib v4.30.0

codygunton:
Bumps VCV-io from Lean and Mathlib **v4.29.0** to **v4.30.0** to synchronize versions across ArkLib, CompPoly and VCV-io.

### What changed
- Updated `lean-toolchain` and the `mathlib` Lake requirement to `v4.30.0`; pinned PolyFun to v4.30-compatible `5d3a160`.
- Kept `loom2` and `Hax` source pins unchanged; they are only rebuilt under the root toolchain.
- Applied mechanical API updates: `zero_le`'s element argument is now implicit, ENNReal lemmas use current camel-case names, and the deprecated `Mathlib.Data.Array.Extract` import is replaced by `Batteries.Data.Array.Lemmas`.
- Kept proof repairs narrow:
  - EvalDist PMF/Id support proofs replace v4.30-broken `grind` calls with direct support/rfl rewrites.
  - Merkle dependent-vector proofs keep their old `grind` shape and add the needed `SkeletonLeafIndex.depth` equations.
  - MLDSA rounding replaces the failing `ring_nf` tail with local `hdam` rewrites.
- Removed formatting-only churn and style-exception changes. The remaining prologue/header edits are only to satisfy the repository non-sorry warning budget.
- Dropped local `FinEnum` instances now provided by Mathlib v4.30, keeping only local instances Mathlib does not provide.


dtumad:
It should be possible to drop the `FinEnum UInt8` and similar instances in `SampleableType.lean` as part of this as well, they are in mathlib now.

codygunton:
> It should be possible to drop the `FinEnum UInt8` and similar instances in `SampleableType.lean` as part of this as well, they are in mathlib now.

ty @dtumad , did that in https://github.com/Verified-zkEVM/VCV-io/pull/427/commits/91d0437d7434eba44fa3505980587546da5231d5

codygunton:
Idk if I'm missing anything from the concurrent attempt #432, happy to combine and get the best of both.

alexanderlhicks (approved):

PR #419, opened 2026-05-30
refactor(EvalDist): rebuild probability semantics on MonadLiftT

dtumad:
The motivation of this PR is to reduce the requirements needed to use the various definitions and notation in the library, making them more broadly applicable and hopefully leading to proofs that are easier to integrate with other work. Replaces the bespoke `HasEvalSPMF` / `HasEvalPMF` / `HasEvalSet` hierarchy with Lean's standard `MonadLiftT` lifts, plus one propositional bridge class `EvalDistCompatible`. Each probability notion now states exactly the assumption it needs:

| Notion | Notation | New assumption(s) |
|---|---|---|
| `evalDist` | `𝒟[mx]` | `[MonadLiftT m SPMF]` (`evalDist := liftM`) |
| `probOutput` | `Pr[= x \| mx]` | `[MonadLiftT m SPMF]` |
| `probEvent` | `Pr[p \| mx]` | `[MonadLiftT m SPMF]` |
| `probFailure` | `Pr[⊥ \| mx]` | `[MonadLiftT m SPMF]` |
| `support` | `support mx` | `[MonadLiftT m SetM]` alone — no probability structure |
| `finSupport` | `finSupport mx` | custom `[HasEvalFinset m]` (now built on `MonadLiftT m SetM`) |
| **support ↔ prob connection** | `mem_support_iff`, `probOutput_eq_zero_iff`, `probEvent_mono`, `probEvent_eq_one_iff`, `probFailure_eq_one_iff`, … | both lifts **plus** new `[EvalDistCompatible m]` |
| never-fails / total-mass | `probFailure_of_liftM_PMF`, `tsum_probOutput_of_liftM_PMF`, `NeverFail` instance | `[MonadLiftT m PMF]` |
| bind-distributivity | `probOutput_bind_eq_*`, `*_bind_eq_sum_fintype` | additionally `[LawfulMonadLiftT m SPMF]` |

The new `EvalDistCompatible m` (`[MonadLiftT m SetM] [MonadLiftT m SPMF] : Prop`) carries a single field `support_eq_SPMF_support` — `SetM.run (liftM mx) = SPMF.support (liftM mx)` — and is the *only* glue tying the qualitative (`support`) and quantitative (`Pr[…]`) views together.

### `OracleComp spec`

[...]

PR #430, opened 2026-06-07
refactor(ring): notations, CommRing instances, and ergonomic vector ops

ChihChengLiang:
## Summary

- **`TransformOps` / `NTTRingOps`**: converted from `structure` to `class`; removed `zeroHat`/`addHat`/`subHat` fields (now inherited from `[AddCommGroup Hat]`); extended `Laws` with `mul_add`, `mul_sub`, `mul_comm`, `mul_assoc`
- **`NegacyclicRing`**: added `one` field and `one_sound` to `NegacyclicRingSemantics`; added `Mul` and `One` typeclass instances for `ring.Poly`
- **`Poly`**: changed from `abbrev` to `def` to prevent simp loops from the new `CommRing` instance; forwarding instances (priority 1100) bridge the gap
- **`vectorNegacyclicRing_instCommRing`**: new `CommRing (Poly Coeff n)` instance derived by lifting axioms through the injective semantics map
- **`vectorNegacyclicSemantics`**: now requires `0 < n` (needed for `one_sound`); ML-DSA/ML-KEM pass `by norm_num`, Falcon passes `hn` from context
- **Ergonomic notation**: `c • v` for scalar-vector mul, `A * v` for matrix-vector mul via scoped `TransformOps` instances; scheme files updated throughout
- **`ofBackend_injective`** and `toPolynomial_injective` added to `Ring/Core.lean`
- **`AddCommGroup (TransformPoly ring)`** instance added to `Transform.lean`
- **Docstrings fixed**: `TransformOps`, `Laws`, `Poly` module header updated to reflect removed fields and new constraints
- **Test file**: `LatticeCryptoTest/MLDSA/Main.lean` updated to use `attribute [local instance]` instead of broken positional `nttOps` arguments
- **Cleanup**: `vectorRing_mul_sub_right` added to `VectorBackend`; duplicated `negacyclicMul_{add,sub}_right` in MLDSA/MLKEM NTT files now delegate to the generic versions


LatticeCrypto/MLDSA/Security.lean
```diff
@@ -67,23 +67,22 @@ variable
...
 here to separate the probabilistic argument from the algebraic one. -/
 theorem idsWithAbort_complete' :
-    (identificationScheme p prims nttOps).Complete := by
+    (identificationScheme p prims).Complete := by
   classical
   intro pk sk hvalid
   rw [probOutput_eq_one_iff_forall]
-  refine ⟨HasEvalPMF.probFailure_eq_zero _, ?_⟩
-  intro b hb
+  refine ⟨HasEvalPMF.probFailure_eq_zero _, fun b hb => ?_⟩
```
dtumad: This needs to use `probFailure_of_liftM_PMF` or some similar lemma instead after #419.

ChihChengLiang: Thanks for the hint. 
It might be cool to mark theorems with something like `@[deprecated probFailure_of_liftM_PMF "since #419" (since := "2026-06-08")]`, so that they can be gradually phased out in up coming PRs.

dtumad: I was a bit unsure since the actual type-signatures would need to change destructively anyways, but yeah I think keeping the names temporarily under a modified signature probably makes more sense going forward.

LatticeCrypto/Ring/VectorBackend.lean
```diff
@@ -29,8 +29,12 @@ universe u
 
 namespace LatticeCrypto
 
-/-- A degree-`n` polynomial represented by a coefficient vector. -/
-abbrev Poly (Coeff : Type u) (n : Nat) := Vector Coeff n
+/-- A degree-`n` polynomial represented by a coefficient vector.
```
dtumad, line 32: I wonder if the `Array` based polynomials from `CompPoly` would be sufficient to replace this lattice-specific `Vector` definition? Would probably at least give more lemmas for free but I'm not sure if there might be instance mismatch problems in how this gets used for lattices. Doesn't need to be part of this PR but worth considering I think.

cc @dhsorens

dtumad (approved):

PR #450, opened 2026-06-24
feat: The Kilian transformation

BoltonBailey:
Introduces the Kilian '92 transformation to derive a SNARG from a PCP.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


alexanderlhicks:
FYI: https://github.com/z-tech/z-lean/tree/main/Kilian

cc @z-tech

VCVio/CryptoFoundations/VectorCommitment/Basic.lean
```diff
@@ -0,0 +1,233 @@
...
+
+namespace VectorCommitment
+
+variable {m : Type → Type} [Monad m]
+  [MonadLiftT m SetM] [LawfulMonadLiftT m SetM]
+  {ι α Commit State Opening : Type}
+
+/-- A vector commitment is **perfectly correct** if, for every vector and position, any honestly
+produced commitment/state opens at that position to the committed value and passes verification. -/
+def PerfectlyCorrect (vc : VectorCommitment m ι α Commit State Opening) : Prop :=
```
dtumad, line 84: This concludes `verifyOpen c i (vc.decode st i) op = true` and never relates `decode st i` to `data i`, so a commitment that ignores `data` satisfies it — see the compiled countermodel in the main comment, which is also position binding. The docstring just above promises more than this ("opens at that position to the committed value").

VCVio/CryptoFoundations/VectorCommitment/Basic.lean
```diff
@@ -0,0 +1,233 @@
...
+
+section Security
+
+variable [Monad m] [MonadLiftT m SetM] [LawfulMonadLiftT m SetM]
+
+/-- A batch-opening vector commitment is **perfectly correct** if, for every vector and every list
+of positions, any honestly produced commitment/state opens that list to a batch opening that
+verifies the corresponding decoded claims. The claim list `is.map (fun i => (i, decode st i))` is
+exactly the shape consumers (such as the Kilian transformation) hand to `verifyBatch`. -/
+def PerfectlyCorrect
```
dtumad, line 158: Same shape as the single-position version: the claim list is built from `decode st i`, so nothing ties it to `data`.

VCVio/CryptoFoundations/ChallengeVerifyProtocol/Basic.lean
```diff
@@ -0,0 +1,223 @@
...
+Commitments are split into a public part `Commit` (revealed to the verifier) and a
+private part `PrvState` (retained by the prover). Verifier challenges are drawn from `Chal`
+via the `sampleChal` computation. Prover responses are in `Resp`.
+
+The prover's computations live in an arbitrary monad `m` carrying probability semantics. Taking
+`m := ProbComp` recovers the usual notion of a Σ-protocol whose only randomness is uniform
+sampling, but a general `m` lets the prover and challenge sampler additionally query oracles.
+
+We leave properties like special soundness as separate definitions for better modularity. -/
+structure ChallengeVerifyProtocol
```
dtumad: This has the same six type parameters, the same `rel`, and the same `commit`/`respond`/`verify` as `SigmaProtocol` (`VCVio/CryptoFoundations/SigmaProtocol.lean:48`) — the differences are `m`-genericity (worth keeping) and `sampleChal` in place of `sim`/`extract`. Could this be the `m`-generalization of that structure with `sim`/`extract` factored into a mixin, rather than a second protocol record? `SigmaProtocol` already carries Fischlin, the FS-Sigma layer, `IdenSchemeWithAbort` and Schnorr.

dtumad (commented):
I read this for the soundness argument and the vector-commitment contract it rests on. The soundness transfer looks right to me: the commitment is fixed before the challenge is drawn, the response may depend on the challenge arbitrarily so the prover is genuinely adaptive, perfect `BatchPositionBinding` forces any two accepted openings — including conflicting claims at a duplicated position — to agree with one global committed string, a missing or failed opening makes the replay fail rather than silently supplying data, and the accepted query path replays against that same global string so PCP soundness transfers with the same `ε`. That is a good consumer theorem for the interaction and commitment abstractions.

One contract issue, in the generic API rather than in the Kilian proof.

`VectorCommitment.PerfectlyCorrect` says supported openings verify `vc.decode st i`. It never says `vc.decode st i = data i`, so it does not rule out a commitment that ignores its input entirely. The docstring above it does make that stronger promise — "opens at that position to the committed value" — but the predicate does not. A commitment that discards `data`, decodes a fixed vector, and opens that is perfectly correct *and* position binding:

```lean
def ignoresInput (m : Type → Type) [Monad m] :
    VectorCommitment m Unit Bool Unit Unit Unit where
  commit _ := pure ((), ())
  decode _ _ := false
  openAt _ _ := pure ()
  verifyOpen _ _ v _ := !v

theorem ignoresInput_perfectlyCorrect : (ignoresInput m).PerfectlyCorrect := by
  intro data i c st _ op _; rfl

theorem ignoresInput_positionBinding : (ignoresInput m).PositionBinding := by
  intro c i v₁ op₁ v₂ op₂ h₁ h₂
  simp only [ignoresInput, Bool.not_eq_true'] at h₁ h₂
  simp [h₁, h₂]

theorem ignoresInput_decode_ne_data :
    ∃ (data : Unit → Bool) (c : Unit) (st : Unit),
      (c, st) ∈ support ((ignoresInput m).commit data) ∧
      (ignoresInput m).decode st () ≠ data () := by
  refine ⟨fun _ => true, (), (), ?_, by simp [ignoresInput]⟩
  simp [ignoresInput]
```

`ignoresInput_perfectlyCorrect` needs no axioms at all — it is `rfl`. The binding half matters here: this is not a degenerate scheme that fails some other law, it satisfies both stated properties while committing to nothing.

Kilian is not affected, because completeness assumes decode fidelity, commit success and opening success separately. But the batch analogue has the same shape, and once this is the generic vector-commitment API other consumers will reasonably read the name as "opening recovers what was committed". I would either add a named decode-fidelity law (`decode st i = data i` on supported commitments) and bundle it, or rename the current predicate to something like `SupportedOpeningsVerify` and let the docstring say only what it proves. Totality can stay separate.

The other thing I would settle before this grows consumers is `ChallengeVerifyProtocol` itself, because we already have this structure. `VCVio/CryptoFoundations/SigmaProtocol.lean:48` declares

```lean
structure SigmaProtocol (Stmt Wit Commit PrvState Chal Resp : Type) (rel : Stmt → Wit → Bool)
```

with the same six type parameters, the same `rel`, and the same `commit` / `respond` / `verify` fields. The new one differs in two ways: it is generic in `m` rather than pinned to `ProbComp`, and it trades `sim` / `extract` for `sampleChal`. The first difference is a strict improvement and the one I would want to keep. The second looks like a factoring question rather than a different concept — `sim` and `extract` are data witnessing HVZK and special soundness, so they arguably do not belong in the protocol shape at all, while `sampleChal` (public-coin challenge sampling) does.

So rather than a parallel hierarchy, could this be the `m`-generalization of `SigmaProtocol`, with `sim`/`extract` split out into a mixin that the properties consume? `SigmaProtocol` is load-bearing — Fischlin, the Fiat–Shamir Sigma layer, `IdenSchemeWithAbort` and the Schnorr examples all sit on it — so generalizing it and having Kilian instantiate the shared base seems better than having two protocol records that agree on every field a consumer actually pattern-matches. If the two really are different concepts I would want the docstring to say what the distinction is, since the signatures do not show one.

Related, and more of a direction question than a request here: `PolyFun.Interaction.TwoParty` (`Strategy.lean`, `Compose.lean`) is where two-party protocols are getting roles, strategies, and composition/factorization, with the UC layer above it. A fixed three-message record composes with none of that. If Kilian is meant to be the demonstration that our abstractions support a recognizable protocol proof — which I think is the right ambition for it — it is worth deciding whether new protocol abstractions should be landing in the interaction layer instead, where the composition theory already exists. Not something to resolve in this PR, but worth knowing the intended answer before more consumers attach to `ChallengeVerifyProtocol`.

For the record on scope, since it affects nothing above: what is proved here is `KilianTransformation_perfectlyComplete` and `KilianTransformation_sound` — soundness transfer under perfect batch-position binding. Succinctness, computational Merkle binding, knowledge soundness and Fiat–Shamir are all still open, which seems like the right amount for one PR.

BoltonBailey:
The decode = data comment has to do with something I am not sure how to resolve: A commitment scheme produces prover state, which in principle should itself be able to reconstruct the committed vector. So I think in certain situations, we have to ask, do we want the argument of such-and-such function to contain both the state, or both the state and the data. If it contains just the state, we could reconstruct the data, but the reconstruction procedure might not be very efficient. But if we pass both, the code might not be DRY.

BoltonBailey:
For now I think I will delete decode and assume the prover passes forward this data.

dtumad:
Deep semantic audit across the open PR set (semantics only). The information-theoretic core here is honestly scoped and the proofs are fine; the blocker is that the headline's central hypothesis has no possible model in the setting the PR itself builds:

**BLOCKING: `VectorCommitment.BatchPositionBinding` is *perfect* binding, and no `hashFn` on a finite type with `|α| ≥ 2` can satisfy it for the Merkle instance on any skeleton with ≥ 2 leaves — so `KilianTransformation_sound` currently has no shipped model.** Sketch: `verifyOpen` recomputes `getPutativeRootWithHash` from the leaf value and path. Binding at a *left*-oriented leaf requires `hashFn v b = hashFn v' b' → v = v'` (a collision differing in first argument, padded with an identical path suffix, gives two verifying openings of different values at that position); binding at a *right*-oriented leaf symmetrically requires `→ b = b'`. Any skeleton with at least two leaves has leaves of both orientations, so the two requirements jointly force `hashFn : α × α → α` to be injective — impossible for finite `|α| ≥ 2`. The docstring already concedes the computational statement is future work, but as it stands the PR ships zero inhabitants of its own hypothesis; per the repo's non-vacuity standard the minimum is a witness instance — e.g. an injective pairing `hashFn` over an infinite `α` (`ℕ ⊕ ℕ × ℕ`-style) with the `BatchPositionBinding` proof — and the honest path for the real result is the computational variant carrying a collision-advantage term.

**Semantic hazard to record now, before the computational variant is written:** the motivating instantiation (`m := OracleComp romSpec` for "a hash oracle for the Kilian transform") lands in exactly the semantics #558 repaired away — raw `OracleComp` samples a *fresh independent response per query*, so two equal hash inputs may receive different answers, which invalidated the previous Merkle extractability proof. Your soundness statement is currently safe only because the Merkle instance is pure and the challenge lifts from `ProbComp`. The computational-binding statement must run under `withCacheOverlay` (or the addressed shared-ROM layer from #558–#591). Worth a sentence in the module docstring so nobody writes the naive version. (Good news on the same subject: this PR does *not* collide with the merkle extractability rework — it imports only the pure `Inductive/{Defs,Completeness}` layer, so the rebase past #613/#616 is mechanical.)

Two structural asks:

1. This introduces a third commitment abstraction (`VectorCommitment`) alongside `CommitmentScheme` and `HashCommitment` with no relating theorem — either derive it or add a docstring stating precisely why it can't be derived (the vector/batch shape may well justify it, but say so).
2. Coordinate with #418: both PRs rewrite `SigmaProtocol.lean` incompatibly (this one 73/45, #418 102/44). Cleanest sequencing is #418 (with the lawfulness fixes requested there — its `[LawfulMonadLiftT m SPMF] [LawfulMonadLiftT m SetM] [EvalDistCompatible m]` signature should match what you already require here, which is the right pattern) landing first, then this rebases onto it.

PR #452, opened 2026-06-25
feat(EvalDist): add `evalDist_bind_bind_swap` as root of the swap family

### LINKED ISSUE #453: `bind_bind_swap` family lacks an `evalDist`-level form and carries a spurious `[LawfulMonad m]`
`VCVio/EvalDist/Monad/Basic.lean` proves that swapping two independent draws is probability-preserving, but only at the `probEvent`/`probOutput` levels (`probEvent_bind_bind_swap`, `probOutput_bind_bind_swap`). The distribution-level form is missing:

```lean
𝒟[mx >>= fun a => my >>= fun b => f a b] = 𝒟[my >>= fun b => mx >>= fun a => f a b]
```

This could be a natural root of the family (the two existing lemmas are corollaries by congruence) and the form needed when rewriting under `𝒟[…]`. Its absence seems to have forced a workaround: `Examples/PRFTagReader/Table.lean` carries a private `ProbComp` copy, `evalDist_probComp_bind_comm`.

Separately, the two existing lemmas carry a spurious `[LawfulMonad m]`: their proofs use no monad law of `m`, only `tsum` rearrangement (via `probEvent_bind_eq_tsum`), which needs just `[LawfulMonadLiftT m SPMF]`.

astefano:
Adds the `evalDist`-level form of the independent-draw swap as the root of the family:

```lean
lemma evalDist_bind_bind_swap [LawfulMonadLiftT m SPMF]
    (mx : m α) (my : m β) (f : α → β → m γ) :
    𝒟[mx >>= fun a => my >>= fun b => f a b] =
      𝒟[my >>= fun b => mx >>= fun a => f a b]
```

- `probEvent_bind_bind_swap` and `probOutput_bind_bind_swap` now derive from it by congruence, each a one-line `rw`.
- The proof works on the `SPMF` image and uses no monad law of `m`, so the spurious `[LawfulMonad m]` is dropped from all three lemmas. This is backward-compatible: the full non-test build is green, and `#print axioms` for all three shows only `propext`, `Classical.choice`, `Quot.sound`.

The second commit removes the duplicate: `Examples/PRFTagReader/Table.lean`'s `evalDist_probComp_bind_comm` drops its private `tsum` proof and now specializes `evalDist_bind_bind_swap` in one line (its 13 call sites unchanged). It stays as a thin `ProbComp` wrapper; happy to inline it at the call sites instead if preferred.

Closes #453 

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Examples/PRFTagReader/Table.lean
+++ b/Examples/PRFTagReader/Table.lean
@@ -440,23 +440,13 @@ lemma evalDist_idealCacheMapM_bind_uniformTable {D : Type} [DecidableEq D] [Fini
     exact funext fun r => ih r.2
 
 /-- Two probabilistic samples may be drawn in either order: the output distribution of drawing
-`mx` then `my` and combining is the same as drawing `my` then `mx`. Proven at the distribution
-level by `tsum` rearrangement; the underlying monads need not be commutative as terms. -/
+`mx` then `my` and combining is the same as drawing `my` then `mx`. The `ProbComp` specialization
+of the generic `evalDist_bind_bind_swap`; the underlying monads need not be commutative as terms. -/
 lemma evalDist_probComp_bind_comm {α₁ α₂ β : Type}
     (mx : ProbComp α₁) (my : ProbComp α₂) (F : α₁ → α₂ → ProbComp β) :
     𝒟[mx >>= fun a => my >>= fun b => F a b] =
-      𝒟[my >>= fun b => mx >>= fun a => F a b] := by
-  refine evalDist_ext fun y => ?_
-  rw [probOutput_bind_eq_tsum, probOutput_bind_eq_tsum]
-  have hL : ∀ a, Pr[= a | mx] * Pr[= y | my >>= fun b => F a b] =
-      ∑' b, Pr[= a | mx] * (Pr[= b | my] * Pr[= y | F a b]) := fun a => by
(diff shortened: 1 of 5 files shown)


Examples/PRFTagReader/Table.lean
```diff
@@ -440,23 +440,13 @@ lemma evalDist_idealCacheMapM_bind_uniformTable {D : Type} [DecidableEq D] [Fini
     exact funext fun r => ih r.2
 
 /-- Two probabilistic samples may be drawn in either order: the output distribution of drawing
-`mx` then `my` and combining is the same as drawing `my` then `mx`. Proven at the distribution
-level by `tsum` rearrangement; the underlying monads need not be commutative as terms. -/
+`mx` then `my` and combining is the same as drawing `my` then `mx`. The `ProbComp` specialization
+of the generic `evalDist_bind_bind_swap`; the underlying monads need not be commutative as terms. -/
 lemma evalDist_probComp_bind_comm {α₁ α₂ β : Type}
```
dtumad: Can you actually just delete this lemma and use your `evalDist_bind_bind_swap` wherever it appears? `ProbComp` should have the necessary instance already. 

Other than that I think this looks good to me.

dtumad (approved):
Thanks

Issue #454, opened 2026-06-25
`SampleableType (α × β)` instance carries unneeded `[Fintype]`/`[Inhabited]` constraints


astefano:
The [product instance](https://github.com/Verified-zkEVM/VCV-io/blob/a41b514f16a4a1a9d48ab193797c763956dfe4ed/VCVio/OracleComp/Constructions/SampleableType.lean#L287-L292) requires `[Fintype α] [Fintype β] [Inhabited α] [Inhabited β]`, but they don't seem to be needed anymore:

- `Fintype` only fed the original cardinality-based `probOutput_selectElem_eq` proof (via `probOutput_uniformSample`, `[Fintype]`). A `Fintype`-free proof exists using [`probOutput_seq_map_prod_mk_eq_mul`](https://github.com/Verified-zkEVM/VCV-io/blob/a41b514f16a4a1a9d48ab193797c763956dfe4ed/VCVio/EvalDist/Prod.lean#L88-L91) + the class's own `probOutput_selectElem_eq` per component.
- `Inhabited` fed the old `probFailure_selectElem` field, **deleted in #97**.

The [class itself](https://github.com/Verified-zkEVM/VCV-io/blob/a41b514f16a4a1a9d48ab193797c763956dfe4ed/VCVio/OracleComp/Constructions/SampleableType.lean#L40-L48) doesn't seem to need `Fintype` (its docstring seems to say so). The constraints just block
`SampleableType (α × β)` for abstract `SampleableType` carriers with no `Fintype`.

We could relax to `[SampleableType α] [SampleableType β]` only (same `selectElem`), and optionally add `uniformSample_prod_eq_bind : $ᵗ(α×β) = do a ← $ᵗα; b ← $ᵗβ; pure (a,b)`. 

The cost is that `PRGfromPRF.lean` declares `[Fintype S/O] [Inhabited S/O]` as section vars used only to synthesize `SampleableType (S × O)`; relaxing makes them unused on some decls, tripping the unused-`Fintype` linter. So it needs a targeted signature cleanup there (~20 hyps, some public). 

Would a PR dropping the constraints be of interest?

dtumad:
Solved in #459

PR #459, opened 2026-06-26
feat: fix grind to fail quickly on probability goals

dtumad:
## Summary

`grind` could saturate and time out on routine probability goals. This PR drops the `@[grind]` tags
responsible so `grind` now fails fast (with a cleaner residual goal) instead of looping, while
keeping the information it needs reachable through narrower companions. Adds three benchmark files
that gate `simp`/`grind` over the probability surface.

## The `grind` problem

The support-*characterization* lemmas rewrite a probability/failure fact into an unbounded
quantifier or set over `support`:

- `Pr[ p | mx] = 0 ↔ ∀ x ∈ support mx, ¬ p x`
- `Pr[ p | mx] = 1 ↔ Pr[⊥ | mx] = 0 ∧ ∀ x ∈ support mx, p x`
- `Pr[= x | mx] = 1 ↔ Pr[⊥ | mx] = 0 ∧ support mx = {x}`
- `Pr[⊥ | mx] = 1 ↔ support mx = ∅`, …

Tagged `@[grind]`, these are a **saturation hazard**: as `grind` case-splits the iff it Skolemizes
the support quantifier into fresh witnesses, which the always-on `bind`-expansion lemmas
(`support_bind`, `mem_support_bind_iff`, …) re-expand into more `support`/`Pr[…]` terms with no
finite grounding (`support ($ᵗ α) = Set.univ` is infinite). The hazard is **combinatorial, not
per-lemma** — no single one saturates alone, but tagged *together* they form a re-trigger cycle and
a naive `grind` on a probability goal runs ~25× longer. The hub of the cycle is the
`probEvent_eq_one_iff` family, whose RHS couples the `probEvent`, `probFailure`, and `support`
layers at once (dropping it alone roughly quarters the blow-up).

## Dropped tags

Now `@[simp]`-only (fixed orientation, no case-split — safe): `probEvent_eq_zero_iff(')`,
`probEvent_ne_zero_iff(')`, `probEvent_eq_one_iff(')`, `one_eq_probEvent_iff(')`,
`probOutput_eq_one_iff`, `one_eq_probOutput_iff`, `probFailure_eq_one_iff`, plus
`mem_support_bind_iff` / `mem_finSupport_bind_iff` (untagged).

Kept `@[grind =]`: the directed single-variable membership bridges (`probOutput_eq_zero_iff`,
`probOutput_pos_iff`, `mem_finSupport_iff`) and a few support-quantifier lemmas verified to sit
*outside* the hub cycle (`probEvent_pos_iff(')`, `probOutput_eq_one_iff'`,
`probFailure_bind_eq_zero_iff`).

To keep the dropped information reachable without re-forming the cycle:

- **`Set.Nonempty` companions** (`@[grind =]`): `grind` keeps `Set.Nonempty` atomic, so
  `probFailure_eq_one_iff_not_nonempty`, `probEvent_ne_zero_iff_nonempty`,
  `probEvent_eq_zero_iff_not_nonempty` (+ `support_uniformSample_nonempty`) carry the same facts
  without the saturating quantifier. The `= 1` event companions are deliberately omitted — their
  `Pr[⊥] = 0` conjunct re-couples them to the hub.
- **Monad/functor law normalization**: `bind_pure`, `pure_bind`, `bind_assoc`, `map_pure` are now
  `@[grind =]`, so `grind` collapses computation structure *before* falling into `tsum` expansion.
- If a proof genuinely needs a dropped lemma, re-supply it locally:
  `grind [probEvent_eq_zero_iff]`. Downstream call sites were updated this way (or to explicit
  terms).

## Test gates — "`grind` should solve these"

Three benchmark files lock in the intended `simp`/`grind` behavior so a regression in either
surfaces in isolation rather than deep in a downstream proof. Each fact closes by a single
*terminal* tactic; where both close it, both are kept (mirrors); where only one does, the gap is a
`target(simp)` / `target(grind)` note.

- `VCVioTest/ProbabilityTactics.lean` — broad corpus over concrete `ProbComp`, organised by
  category (outcome / event / failure / support / distribution).
- `VCVioTest/MonadProbability.lean` — the same gate over a generic monad `m` and the concrete
  transformers (`OptionT`, `ExceptT`, `SPMF`, `Id`); surfaces facts `ProbComp` masks, chiefly the
  failure factor.
- `VCVioTest/LongChainPrograms.lean` — programs with 10+ sequential binds; gates that `grind` keeps
  pace with `simp` on the deep failure-probability telescope and on abstract monad-law
  normalization.

## Also

- Docs: a new "`grind` vs `simp` on Probability Goals" section in `docs/agents/probability.md`, and
  an updated gotcha in `docs/agents/gotchas.md`.
- Minor new API filled in along the way: `probOutput_map`, `support_guard`, and the `orElse`
  (`<|>`) probability lemmas for `OptionT (OracleComp spec)`; plus a `QueryCache` cleanup (redundant
  `[spec.DecidableEq]` dropped, a few confluent `@[grind]` tags added).

### DIFF
--- a/VCVio/EvalDist/Monad/Basic.lean
+++ b/VCVio/EvalDist/Monad/Basic.lean
@@ -18,6 +18,14 @@ variable {α β γ : Type u} {m : Type u → Type v} [Monad m]
 
 open ENNReal
 
+/- The monad/functor laws are confluent (terminating) rewrites. Tagging them for `grind` lets it
+normalize a computation's structure (`mx >>= pure = mx`, `pure a >>= f = f a`, reassociation,
+`f <$> pure a = pure (f a)`) *before* it falls into `probOutput`/`tsum` expansion — turning what
+would otherwise be a `grind` explosion on a structured-computation equality into a quick solve. (The
+analogous `bind_pure_comp`/`map_eq_bind` laws are deliberately omitted: their function argument sits
+under a binder that `grind`'s pattern compiler cannot index.) -/
+attribute [grind =] bind_pure pure_bind bind_assoc map_pure
+
 /-! ## Probabilities of `pure` -/
 
 section pure
@@ -115,7 +123,7 @@ lemma support_bind [MonadLiftT m SetM] [LawfulMonadLiftT m SetM] (mx : m α) (my
     support (mx >>= my) = ⋃ x ∈ support mx, support (my x) :=
   monadLift_bind mx my
 
-@[grind =]
+-- untagged: the `∃ x ∈ support` RHS saturates `grind`; `support_bind` is the `simp` form.
 lemma mem_support_bind_iff [MonadLiftT m SetM] [LawfulMonadLiftT m SetM] (mx : m α)
     (my : α → m β) (y : β) :
     y ∈ support (mx >>= my) ↔ ∃ x ∈ support mx, y ∈ support (my x) := by simp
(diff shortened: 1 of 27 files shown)


dtumad:
Also this was noticed while working on it so this closes #454

VCVio/EvalDist/Monad/Basic.lean
```diff
@@ -18,6 +18,14 @@ variable {α β γ : Type u} {m : Type u → Type v} [Monad m]
 
 open ENNReal
 
+/- The monad/functor laws are confluent (terminating) rewrites. Tagging them for `grind` lets it
+normalize a computation's structure (`mx >>= pure = mx`, `pure a >>= f = f a`, reassociation,
+`f <$> pure a = pure (f a)`) *before* it falls into `probOutput`/`tsum` expansion — turning what
+would otherwise be a `grind` explosion on a structured-computation equality into a quick solve. (The
+analogous `bind_pure_comp`/`map_eq_bind` laws are deliberately omitted: their function argument sits
+under a binder that `grind`'s pattern compiler cannot index.) -/
+attribute [grind =] bind_pure pure_bind bind_assoc map_pure
```
dtumad: This is maybe dangerous for downstream consumers? I do kind of feel like base Lean should just have them though really

Issue #396, opened 2026-05-14
Untracked proof obligations in the GPV / Falcon / Rényi chain (gaps from #187, #279)


dtumad:
## Summary

A recent audit of the proof-obligation issue tracker against the current codebase
turned up seven `sorry`s that have no individual ticket. They were introduced by
#187 and #279 and all sit on the same proof spine: Falcon's EUF-CMA security via
GPV hash-and-sign, plus the Rényi-divergence bound that the Falcon analysis relies
on, plus the two Falcon scheme primitives the whole stack quantifies over.

The most important consequence: the "0-sorry" status of `euf_cma_collision_bound`
and `euf_cma_split_bound` (the renamed successors of the obligation tracked in
#188) is transitively false — both theorems are discharged with
`exact ⟨reduction …, forgery_yields_collision …⟩` against sorry-bodied helpers.
**#188 should be treated as still open** until the items below are filled in.

### GPV reductions (witnesses #188 currently leans on)

- [ ] `reduction` — `VCVio/CryptoFoundations/GPVHashAndSign.lean:266`
  (`fun _pk => sorry`, introduced by #187). Collision-finding adversary used by
  `euf_cma_collision_bound`.
- [ ] `programmedPreimageReduction` — `VCVio/CryptoFoundations/GPVHashAndSign.lean:279`
  (`fun _pk _y => sorry`, introduced by #279). Exact-match adversary used by
  `euf_cma_split_bound`.
- [ ] `forgery_yields_collision` — `VCVio/CryptoFoundations/GPVHashAndSign.lean:316`
  (bare `sorry`, introduced by #279). Powers `euf_cma_collision_bound`.
- [ ] `forgery_yields_collision_or_exact_match` —
  `VCVio/CryptoFoundations/GPVHashAndSign.lean:344` (bare `sorry`, introduced by
  #279). Powers `euf_cma_split_bound`.

### Rényi divergence (feeds the Falcon analysis)

- [ ] `etvDist_sq_le_of_renyiDiv` —
  `ToMathlib/Probability/ProbabilityMassFunction/RenyiDivergence.lean:736`
  (bare `sorry`, introduced by #279). The headline Rényi → eTV-distance bound.

### Falcon scheme primitives (quantified over by every Falcon theorem)

- [ ] `keyGenFromSeed` — `LatticeCrypto/Falcon/Scheme.lean:121`
  (`noncomputable def := sorry`, introduced by #187).
[...]

PR #460, opened 2026-06-29
feat(OracleComp): oracle strategies as dynamical systems, oracle machines, and TM-grounded poly-time adversaries

dtumad:
> Depends on **PolyFun#9** (`feat/poly-dynamical-systems`); the lakefile pins its branch HEAD. Kept as a draft until that merges, after which the pin moves to PolyFun `master`.

## Summary

Adds the **coalgebraic dual of `OracleComp`**: an oracle-querying *strategy* modeled as a PolyFun **dynamical system** over `spec.toPFunctor`, plus deterministic/randomized handlers, a closed-loop run + transcript, the headline `iterate = simulateQ` correspondence, denotational query bounds, typed-transcript conversions, and a unification of VCVio's query-log/count carriers with PolyFun's generic trace types. New files: `VCVio/OracleComp/Coinductive/DynSystem.lean`, `VCVio/OracleComp/QueryTracking/Trace.lean`, `Examples/DynamicalSystems/Basic.lean`.

On top of the strategy layer, the PR now also delivers a **foundationally-defined notion of polynomial-time adversary** — polytime as *existence of an implementing oracle strategy whose step functions are computed by concrete Turing machines* — in `VCVio/OracleComp/Coinductive/Machine.lean`, `VCVio/OracleComp/Coinductive/PolyTime.lean`, `VCVio/CryptoFoundations/Asymptotics/PolyTime.lean`, `ToMathlib/Computability/{Encoding,CslibPolyTime,PolyTimeTM}.lean`, and `Examples/DynamicalSystems/PolyTime.lean`.

## The polynomial-functor picture (background)

VCVio already views an oracle signature `OracleSpec ι` as a **polynomial functor** `spec.toPFunctor = ⟨ι, spec⟩`: the positions `p.A = ι` are the queries you may issue, the directions `p.B i = spec.Range i` the possible responses to query `i`. Two dual constructions live on this one functor:

- **Algebra / answer side (inductive).** `OracleComp spec = FreeM spec.toPFunctor` — a finite program that *asks* a sequence of queries and then halts with a value; `simulateQ` folds it against an interpretation.
- **Coalgebra / ask side (this PR).** Following Niu–Spivak (*Polynomial Functors*, Ch. 4), a `p`-**dynamical system** is a state set with `expose : State → p.A` (the position it currently presents) and `update : (s) → p.B (expose s) → State` (how an incoming direction advances it) — equivalently a dependent lens `selfMonomial State ⟹ p`. A **section** `(a) → p.B a` chooses a direction at every position, and `close`/`wrap` glue a system to a section into a `Closed` autonomous machine whose `iterate` is its run.

## The new core: `OracleStrategy` — a potentially unbounded adversarial strategy

```lean
abbrev OracleStrategy (spec : OracleSpec ι) := PFunctor.DynSystem spec.toPFunctor
```

Read on the oracle interface, a dynamical system is exactly a **stateful, adaptive querier**: its `State` is the adversary's working memory (the transcript-so-far, internal coins, whatever it tracks); `expose : State → ι` chooses the **next query** adaptively from the current state; `update : State → spec.Range _ → State` **digests the oracle's answer** and advances.

The crux — and the duality with `OracleComp` — is that **a strategy carries no built-in termination**. Where an `OracleComp` program is an *inductive*, finite syntax tree (it must eventually `pure`), an `OracleStrategy` is a *coalgebra* that can keep exposing queries indefinitely; its full behaviour is the **cofree, potentially infinite trajectory** `PFunctor.DynSystem.trajectory`. So `OracleStrategy` is the natural model for an adversary / environment that runs for an arbitrary, even unbounded, number of rounds — without committing to a finite query count up front.

## What it means in practice

- **Running an adversary against an oracle.** `OracleStrategy.runAgainst h A : Closed` closes the strategy against a handler `h` (just `wrap` along the handler-lens). `Closed.iterate` is the state trajectory; `OracleStrategy.transcript` is the recorded **query/answer log** (`QueryLog spec`), with `@[simp]` recurrences.
- **Randomized oracles → Markov chains.** `ProbHandler spec := QueryImpl spec SPMF` turns the closed loop into a sub-probability **Markov chain on adversary states** (`kleisliStep`/`kleisliIterate`/`transcriptDist`); a deterministic handler embeds as the Dirac case.
- **Programs are finite strategies.** `iterate_advance_eq_simulate` reads a *finite* `OracleComp` program back as its own closed dynamical system and proves the dynamical `iterate` **computes `simulateQ`** — tying the inductive (finite program) and coalgebraic (possibly-unbounded strategy) views together. The probabilistic `simulateQ_eq_advanceK_bind` exhibits `simulateQ` as one coalgebraic Kleisli step.
- **Reductions and products are lens algebra.** `reduce` installs a strategy along a sub-spec coercion `⊂ₒ` (a reduction **is** a `PFunctor.Lens`; composition is free from `wrap_comp`); `pair` targets a product oracle `*`; `juxtapose` the parallel tensor `⊗`.
- **Subsumption, not duplication.** `run_simulateQ_ofFn_withLogging` proves the new transcript *equals* the existing `loggingOracle`/`withLogging` output, so established `QueryLog` lemmas apply; and the operational query bounds become **denotational** bounds on the transcript (`transcript_length_le_of_isTotalQueryBound`, predicate and per-index `countQ` variants).
- **Typed transcripts.** Since `OracleComp = FreeM`, a `PFunctor.FreeM.Path` is a *typed* root-to-leaf transcript; `handlerPath`/`output_handlerPath`/`logOfPath` give the handler-induced path, its leaf (= run value), and its erasure (= flat log); `outputAlong_runAlong` runs a program *along a reduction lens*.

## Handlers as sections; logs as traces

- `OracleHandler spec := PFunctor.Section spec.toPFunctor` is now a first-class **lens** with `ofFn` (build), a `DFunLike` coercion (apply as `h t`), and `toQueryImpl`/`Coe` (drops straight into the deterministic `QueryImpl spec Id` API, e.g. `evalWithAnswerFn h oa`).
- `QueryTracking/Trace.lean` formalizes the latent identification of the tracking carriers with PolyFun's generic traces: `QueryLog spec = PFunctor.TraceList spec.toPFunctor` (the free monoid on query events `PFunctor.Idx spec.toPFunctor`), `QueryCount ι = Control.Trace ℕ ι`, with monoid-valued readouts factoring through the log by `FreeMonoid.lift`.

## Foundationally-defined polynomial-time adversaries

The strategy layer makes a **Turing-machine-grounded definition of "PPT adversary"** possible for the first time. Defining polytime directly on `α → OracleComp spec β` fails structurally: the continuation after each query is an arbitrary semantic Lean function, so a program tree has no bounded syntactic presentation a cost model could attach to. A strategy, by contrast, is a state type plus **pure Lean step functions** — a bounded set of concrete functions that a Turing machine can compute on encoded states. This yields the classic *interactive TM / clocked machine* definition:

> an adversary is polynomial time iff **there exists** an oracle machine implementing it whose step functions are each computed by a Turing machine in polynomial time, and whose readout resolves within polynomially many oracle rounds.

- **Oracle machines** (`Coinductive/Machine.lean`). `OracleMachine spec α β extends OracleStrategy spec` with `init : α → State` and a Moore-style partial readout `output : State → Option β`. Fuelled runs `runD`/`runK` **stop early** at the first `some` readout (essential probabilistically: post-halt sampling would leak sub-probability mass against lossy handlers). The semantic-agreement relation `Implements M oa k` quantifies over **randomized** handlers — deterministic handlers answer repeated queries identically, so `∀ OracleHandler`-agreement provably cannot pin down distributions — with the deterministic form recovered as a Dirac corollary. `IsSimulation` + `implements_of_isSimulation` is the practical proof method: a step-synchronized simulation relation between machine states and program residues, with the fuel supplied by a total query bound.
- **Query bounds ↔ steadiness, both directions.** A query-bounded program *is* a machine — the residual program is the state — steady within its bound against every handler and implementing the program (`OracleComp.toMachine`, `toMachine_steadyBy`, `toMachine_implements`, via the `Eq` simulation and `iterate_advance_eq_of_le`). Conversely a machine unrolled for `k` rounds *is* a program with total query bound `k` and literally the machine's run semantics (`OracleMachine.toComp`, `isTotalQueryBound_toComp`, `simulateQ_toComp`). Steadiness fuel and query bounds are two presentations of the same finiteness.
- **The polytime bundle** (`Coinductive/PolyTime.lean`). `PolyTimeAdversary spec α β` is a security-parameter-indexed family of machines carrying: a polynomial round bound `steps` with stability and steadiness; `Bool`-string encodings of states, inputs, outputs, and the oracle interface; an encoded-state-size invariant `sizeBound` (**provably necessary** — a machine can double its state each step while staying per-step polytime, so per-step witnesses + a round bound alone admit exponential blowup); and per-step machine witnesses `initTM`/`exposeTM`/`updateTM`/`outputTM` under one uniform polynomial `stepTime`. Crucially, **the definition needs no machine composition**: polytime is the conjunction of per-step witnesses, the round bound, and the size invariant — the clocked-machine reading of "PPT". The Prop form `OracleComp.IsPolyTime oa := ∃ D, D.Implements oa ∧ …` is the intended instantiation of `SecurityGame.secureAgainst`'s `isPPT` slot; `IsPolyTime.polyQueries` bridges to the existing `PolyQueries`, and `detTotalTime_le` / `exists_polynomial_detTotalTime_le` prove the **total Turing-machine time of every run polynomial in the security parameter** by pure polynomial arithmetic over the witnesses.
- **TM grounding via Cslib.** New dependency [`leanprover/cslib`](https://github.com/leanprover/cslib) pinned at tag `v4.30.0` (identical Lean toolchain and Mathlib commit to ours), for its sorry-free single-tape model `Turing.SingleTapeTM.PolyTimeComputable` — which comes with **proven polytime composition**, exactly what Mathlib's TM2 model still lists as `proof_wanted`. `Computability.EncPolyTime ea eb f` (`ToMathlib/Computability/CslibPolyTime.lean`) is the encoding-aware wrapper: a machine-computed total string function intertwining `Bool`-string encodings, with `id`/`comp` lifted from Cslib (the monotone-time side condition discharged by normalizing a machine's clock to its own polynomial).
- **Encodings** (`ToMathlib/Computability/Encoding.lean`). Fills Mathlib gaps around `FinEncoding`: `finEncodingOption`, the dependent-pair `finEncodingSigma` (typed query/answer pairs through one alphabet), fixed-width one-hot `boolify` relabeling any finite alphabet into `List Bool`, and `finEncodingOfFinEnum` making every `FinEnum` type (`Unit`, `Bool`, `Fin n`, products, sums, sigmas) encodable for free — so e.g. `coinSpec`'s whole interface encoding is `InterfaceEncoding.ofFinEnum`.
- **A verified base machine** (`ToMathlib/Computability/PolyTimeTM.lean`). `constComputer` computes any constant function in linear time, fully verified against the step semantics: *clear the input moving right, then write the output backwards moving left* — clearing onto an empty left stack keeps the tape canonical and writing back-to-front lands the head on the output's first symbol, so no rewind phases are needed. Gives concrete `EncPolyTime.const` witnesses. Relabeling and projection machines follow the same skeleton and are documented future work, as is the end-to-end single-machine witness per run (clocked iteration on Cslib's composition).
- **Game level** (`CryptoFoundations/Asymptotics/PolyTime.lean`). `SecurityGame.secureAgainstPolyTime := secureAgainst IsPolyTime`, plus a concrete `OneShotGame` former instantiated over both adversary presentations, with the advantage-transfer lemma: an implementing polynomial-time adversary has **exactly** the implemented program's advantage, so security statements move freely between the program and machine layers.
- **Worked examples** (`Examples/DynamicalSystems/PolyTime.lean`). A hand-rolled one-flip machine proved to implement `coinSpec.query ()` through an explicit simulation relation; both program/machine bridges exercised on the two-flip `prog`; free interface encodings; and the query-free polytime certificate (`isPolyTime_pure_of_witnesses`) in honest hypothesis form, with the constant witness already concrete.

## Notes

- **Dependency:** PolyFun#9 (unmerged) — draft until merge, then re-pin to PolyFun `master`. New dependency **cslib @ `v4.30.0`** (tag pins our exact toolchain and Mathlib commit); the single-tape TM module moves paths upstream after v4.31, so the next toolchain bump needs an import fix.
- Full `lake build` green; all new files are `sorry`-free (no `native_decide`); the headline `iterate = simulateQ` correspondence is axiom-clean (`propext`, `Quot.sound`).


quangvdao:
Thanks Devon — this is a very interesting direction, and I am genuinely glad this work is here. I went through the definitions and proofs in detail, compared the generic surface with current PolyFun and cslib main, and looked at the closest literature. My overall conclusion is:

- The coalgebraic OracleStrategy / OracleMachine semantics is strong, well motivated, and worth preserving.
- The separation between a dependently typed interactive protocol and a conventional TM cost model is a good design choice.
- The proposed PolyTimeAdversary / OracleComp.IsPolyTime definition has two structural gaps that currently make it too permissive to serve as VCVio's production PPT predicate.
- I recommend splitting the semantic work from the PPT work, rebasing the former onto current PolyFun, and redesigning the latter before connecting it to SecurityGame.secureAgainstPolyTime.

## 1. The semantic model is compelling

The core type

    State → Σ q, (spec.Range q → State)

is exactly the right shape for a typed adaptive querier: expose chooses the next query from the current state, and update consumes an answer whose type depends on that query. Adding init and a partial Moore-style output gives a natural interactive machine presentation.

This is also a clean way to use dependent types without trying to invent a dependently typed tape TM. The dependent layer enforces the oracle protocol; the actual computability witnesses operate on ordinary finite encodings and conventional single-tape TMs after the dependent query/answer pair is flattened through a Sigma encoding. I think this separation is substantially better than trying to put all the dependent typing into the low-level TM state space.

The inductive/coinductive picture is valuable:

- OracleComp / FreeM is a finite query decision tree.
- OracleStrategy / DynSystem is potentially unbounded reactive behavior.
- OracleMachine adds an initialization and observable stopping condition.
- The toMachine / toComp and iterate / simulateQ correspondences connect the two presentations.

This is directly in the Niu–Spivak polynomial-dynamics lineage. The connection to Libkind–Spivak's [Pattern Runs on Matter](https://arxiv.org/abs/2404.16321) is also real: OracleComp is the finite decision-tree "pattern", while a strategy generates potentially infinite reactive "matter", and the execution correspondence is recognizably a "pattern runs on matter" phenomenon.

One wording qualification: "coalgebraic dual" is good intuition, but should probably be presented as informal. A p-coalgebra is not itself the cofree comonad, and the current handler/iteration construction is not literally the full free-monad/cofree-comonad module action of that paper.

The closest proof-assistant analogues I found are Coq's [Interaction Trees](https://arxiv.org/abs/1906.00046), where events have dependent result types and continuations consume those results, and [CryptHOL](https://www.isa-afp.org/entries/CryptHOL.html), which has probabilistic computations with black-box access to stateful oracles. The latter notably does not attempt to supply the feasibility/PPT layer that this PR is addressing.

## 2. Generic ownership: current PolyFun now subsumes a lot of Machine.lean

PolyFun has moved materially since this PR was started. Current main already has PointedMachine with:

- DynSystem;
- input initialization and partial output;
- sequential composition;
- fuelled conversion to FreeM;
- monad-parametric execution;
- generic simulation/refinement infrastructure.

So I think this PR should be rebased and the generic machine material deduplicated against current PolyFun before we review the final VCVio API.

My proposed boundary is:

**PolyFun**

- DynSystem, trajectories, generic sections/handlers;
- PointedMachine and generic fuelled execution;
- generic reachability, traces, simulation, refinement, and structural composition.

**VCVio**

- the OracleSpec specializations OracleStrategy and OracleHandler;
- ProbHandler/SPMF semantics and early stopping;
- OracleComp bridges, Implements, query bounds and logs;
- crypto games and the eventual PPT predicate.

There is also a generic design choice worth discussing in PolyFun. The current PointedMachine/OracleMachine presentation has a total expose plus output : State → Option β, so even a halted state notionally has a next query and stability is imposed separately. An intrinsic halting coalgebra of the shape

    State → β + Σ q, (Answer q → State)

would make "halt or query" part of the transition type. I do not think this must block the current semantic layer, but it is worth deciding before we grow two parallel machine APIs.

## 3. The standalone TM work is upstreamable, but should be split

I cloned and checked current cslib main. The following look like strong standalone cslib PRs:

1. clearComputer / constComputer and their time/polytime proofs;
2. the finite-prefix/table computer and its time/polytime proofs.

EncPolyTime is also cslib material in principle, but it should not be upstreamed unchanged: it hardcodes Bool and overlaps cslib's open encoding design work in [#590](https://github.com/leanprover/cslib/issues/590), [#591](https://github.com/leanprover/cslib/pull/591), and draft [#595](https://github.com/leanprover/cslib/pull/595). The polynomial-clock normalization also overlaps [#396](https://github.com/leanprover/cslib/pull/396). Some of those PRs are stale/conflicting, so coordination may mean replacing/porting them while preserving credit rather than waiting on them indefinitely.

The generic encoding/list/polynomial lemmas are more naturally Mathlib contributions:

- fixed-width List.flatMap injectivity;
- Nat-polynomial evaluation monotonicity;
- FinEncoding for Option, Sigma and FinEnum;
- a generic finite-alphabet relabelling combinator.

The present boolify one-hot coding works, but it is noncomputable and inefficient. I would prefer a general alphabet relabelling layer followed by an explicit binary encoding.

There is also a current-version port required: this PR pins cslib v4.30.0 and uses the old SingleTapeTuring.Basic path / Turing.SingleTapeTM namespace. Current cslib uses Computability.Machines.Turing.SingleTape.Deterministic and Cslib.Turing.SingleTapeTM.

## 4. Blocking PPT issue: unrestricted per-parameter machine descriptions

PolyTimeAdversary chooses a different M n, four different TMs, and different encodings for every security parameter n. stepTime uniformly bounds the running time of each chosen machine, but nothing bounds:

- the number of TM control states or transition-table size;
- data hard-coded into M n;
- advice/circuit/machine-description size;
- or the complexity of producing the family from n.

This is not standard non-uniform PPT/P-poly. It permits unrestricted advice.

EncPolyTime.ofFintype makes the problem concrete: any function on a fixed finite domain is given a linear-time TM by compiling its entire truth table into finite control. This theorem is correct for a fixed function, but applying it independently at every n can certify an arbitrary parameter-indexed finite function family while charging nothing for exponentially large (or worse) transition tables.

At the extreme, for fₙ : Unit → Bool, the bit fₙ() can encode membership of n in an arbitrary language. Each individual machine has constant runtime because the answer is hard-coded, even if the family is not uniformly computable at all.

A uniform runtime polynomial does not repair this: runtime is being measured after an unbounded machine description has been supplied for free.

I think we need one of:

1. **Uniform PPT:** one fixed oracle TM/evaluator receiving 1ⁿ (plus encoded input and randomness); or
2. **Explicit non-uniform PPT:** a uniform evaluator with polynomial-length advice, or polynomial-size circuit/TM descriptions, together with a polynomial description-size bound.

Until then, OracleComp.IsPolyTime is substantially larger than the class cryptographic security statements intend to quantify over.

## 5. Blocking PPT issue: deterministic handler bounds miss randomized paths

Machine.lean correctly defines ReachableIn over arbitrary typed answer sequences and notes that this is stronger than running against a deterministic OracleHandler: a deterministic handler returns the same answer whenever the same query is repeated.

However, PolyTimeAdversary.steady and encState_length_le quantify only over deterministic handlers, while exec/runK uses ProbHandler and may resample on every occurrence.

Thus a machine can remain small and halt on every constant-answer trajectory, yet grow exponentially or fail to resolve on a varying sequence such as

    false, true, false, true, ...

for repeated calls to the same coin query. Such a path has positive probability under a randomized handler but is invisible to the current state-size and termination invariants.

Implements against every ProbHandler is semantically strong and good, but it does not repair the resource bound: it may constrain final output distribution without bounding the size/cost of intermediate randomized states.

I recommend formulating steadiness/termination and state-size invariants over every ReachableIn state from M.init x — equivalently, every finite typed answer path/random tape — rather than only fixed deterministic handler sections.

This is also closer to standard interactive-machine discipline; e.g. the [IITM model](https://link.springer.com/article/10.1007/s00145-020-09352-1) quantifies resource bounds over arbitrary environments, messages, persistent state and random tapes.

## 6. Other complexity qualifications

- detTotalTime is numerical accounting over four component TM witnesses, not yet a constructed end-to-end oracle TM. The file acknowledges this; I would call it a component-cost bound rather than a complete TM-grounding theorem.
- The bound is for deterministic handlers and requires external polynomial hypotheses for encoded inputs and answers. These conditions are not currently part of OracleComp.IsPolyTime.
- There appears to be an off-by-one in readout accounting: runK checks output initially and after each of steps updates, so it may evaluate output steps + 1 times. detTotalTime charges outputTM only for j in range steps, omitting the final readout.
- Private adversarial randomness via a coin oracle is fine, provided every draw is charged and the resource invariants cover arbitrary answer paths.
- Implements quantifying over randomized handlers is an important strength and should be kept. Deterministic handlers alone cannot distinguish independent resampling from reusing a previous answer.

## Suggested landing plan

1. Rebase on current VCVio/PolyFun/cslib.
2. Split and preserve the semantic DynSystem/OracleStrategy work.
3. Refactor OracleMachine through current PolyFun.PointedMachine, keeping VCV-specific probabilistic semantics and OracleComp implementation theorems here.
4. Upstream the concrete constant/table TM machinery to cslib in small PRs.
5. Coordinate EncPolyTime with cslib's encoding work; send the genuinely generic FinEncoding/list/polynomial pieces to Mathlib.
6. Keep OracleComp.IsPolyTime out of SecurityGame.secureAgainstPolyTime until:
   - uniformity or polynomial advice/description size is enforced;
   - termination and state-size bounds cover all reachable answer paths;
   - input/message-size assumptions are incorporated;
   - final-readout accounting is corrected.
7. Treat an actual compiled interactive-TM theorem as a later second tier if necessary. A first tier of component-wise TM certificates is still useful, provided it is named and documented accurately.

Overall: I am strongly in favor of the semantic direction and would like to see it land. The PPT layer is also aiming at the right problem, but the two gaps above are large enough to invalidate the intended security interpretation in its present form. Splitting the PR should let us merge the genuinely valuable coalgebraic machinery without forcing the complexity model to settle prematurely.

dtumad:
@quangvdao Sorry, I didn't realize this was still open, I've opened #481 which should subsume most of this anyways, and contains a lot less bespoke machinery now that all the `PolyFun` PRs are in. I think a good portion still applies but in particular I think this solves the cheating issue and avoids anything bespoke on the VCV side.

PR #465, opened 2026-07-02
feat(framework): deferred-sampling + relational-coupling engine

alik-eth:
Adds the deferred-sampling and relational-coupling engine that the GPV/#188 and MLDSA/#228 security proofs build on. Purely additive: no existing declaration is changed or removed.

## What's included

- `VCVio/OracleComp/QueryTracking/RandomOracle/DeferredSampling.lean` — expectation algebra for deferred draws: i.i.d. bind-commutation (`evalDist_bind_comm`), the ε-multiplicity kernel (`tsum_probOutput_fresh_mul_count_le`), the `Factorizes` front-tape predicate with `evalDist_step_commute_tape`, and the state-relation transfer lemma `tsum_probOutput_simulateQ_run_mul_of_rel`.
- `VCVio/OracleComp/QueryTracking/RandomOracle/ProbeEps.lean` — hidden-target read games: single-target `q·ε` bound (`probEvent_hiddenReadMany_le`), i.i.d. multi-key `n·q·ε` bound (`probEvent_hiddenReadList_le`), the averaged `E[n]·q·ε` bridge, single-draw deferral (`probEvent_bind_fire_le_of_gen`), and the explicit front-block `drawList`.
- `VCVio/ProgramLogic/Relational/SimulateQ.lean` (additions) — relational-triple monotonicity/implication lemmas, the single-world expected-query-slack accumulator (`probEvent_bad_simulateQ_run_le_expectedQuerySlack`), and the bare-measure averaged-bad primitives (`avgBadM`, `postStepOutM`, `avgBadM_query_bind_eq_tsum_output`, `tsum_tsum_postStepOutM_mul`) used by the aborting-signer coupling.
- `VCVio/ProgramLogic/Relational/ProgrammingOracle.lean` (additions) — TV-distance-to-bad-event bounds for programmed random oracles, including the caching/programming bridge consumed by the GPV reduction.
- `VCVio/EvalDist/Defs/Basic.lean` (additions) — the first-moment bound `probEvent_le_tsum_probOutput_mul_cost` (+ support-restricted variant).

A soundness note for reviewers of the averaged-bad layer: the per-output invariant threading is stated over **un-normalized sub-probability state measures** (`postStepOutM` has no output-conditioning division), which is the formulation that avoids the mass-concentration pathology of a per-output *conditioned* invariant. The packaged PMF-typed telescoping layer that earlier drafts carried was dropped — both downstream consumers telescope directly from the bare-measure primitives, so only the consumed API ships.

## Verification

[...]

PR #466, opened 2026-07-02
feat(Falcon): EUF-CMA of the rejection-loop signer at attempt granularity, down to NTRU-SIS, with a rejecting NTRU witness (#188)

### LINKED ISSUE #188: Proof obligation for `euf_cma_bound` in `VCVio/CryptoFoundations/GPVHashAndSign.lean`
A proof in `VCVio/CryptoFoundations/GPVHashAndSign.lean` contains a `sorry`.

[Link to the sorry on GitHub](https://github.com/Verified-zkEVM/VCV-io/blob/master/VCVio/CryptoFoundations/GPVHashAndSign.lean#L190)

**Code Snippet:**
```lean
2. Each signature is a fresh trapdoor sample, statistically close to ideal Gaussian.
3. A forger must produce a short preimage of a hash value it did not receive from signing,
   which directly yields a preimage-finding adversary.

Reference: GPV08, Theorem 6.1; see also BDF+11 for the QROM extension. -/
theorem euf_cma_bound [DecidableEq Domain] [SampleableType Domain]
```

alik-eth:
Stacked on #478 (`gpv188-generic`); this branch carries #478 verbatim plus the Falcon-only delta below. Review surface: `LatticeCrypto/Falcon/{Scheme,Security,SISBridge,NonVacuity}.lean` and the new generic `VCVio/OracleComp/Constructions/RetryLoop.lean`.

## What changes

**The modelled signer is the rejection loop.** `falconRetryPSF` / `falconRetrySignatureAlg` run `signAttempt` (one `ffSampling` draw plus Falcon's norm check) up to `maxAttempts` times at the hashed target, with a default output on exhaustion (`ProbComp.retryToDefault`). The single-draw scheme `falconSignatureAlg` stays as the one-shot variant.

**The sampler is priced per attempt, not per draw.** Three named hypotheses replace the per-draw transport:

- `attemptTransport`: Falcon's attempt and an idealized attempt are within total-variation distance `ε_step` on honest keys, both sides allowed to reject;
- `attemptRejectBound`: the idealized attempt rejects with probability at most `pRej`;
- `idealAttemptResamples`: the idealized attempt is a rejection sampler for the ideal PSF (`ProbComp.ResamplesTo`).

`RetryLoop.lean` proves the two generic facts these need: per-attempt closeness accumulates as `ε_step / (1 − pRej)` over the loop (`tvDist_retryToDefault_le_div`), and a resampling attempt's loop is within `pRej ^ maxAttempts` of its conditional (`tvDist_retryToDefault_le_pow`). `tvDist_falconRetryPSF_trapdoorSample_le` composes them.

**Headlines.** `euf_cma_security` is the attempt-level split GPV bound for the rejection-loop signer with sampler loss `qSign · (ε_step / (1 − pRej) + pRej ^ maxAttempts)`; `euf_cma_collision_security` discharges its exact-match branch by `idealSamplerGuessBound`; `euf_cma_collision_security_ntruSIS` restates it with the collision term as the keyed NTRU-SIS advantage. The single-draw theorems are renamed `euf_cma_security_oneShot`, `euf_cma_security_oneShot_bytes40`, `euf_cma_security_oneShot_ntruSIS`.

**The one-shot frontier is shown to be vacuous at Falcon's budget, inside the development.** `oneShot_rejection_prob_le_of_samplerTransport` proves that any admissible per-draw budget dominates the norm check's rejection probability on honest keys; `samplerTransport_zero_forces_accept` and `oneShot_samplerLoss_one_le_of_samplerTransport` draw the consequences (`ε_step = 0` forces acceptance with probability one; `qSign · ε_step ≥ 1` once `qSign · p_reject ≥ 1`). `attemptTransport_self_zero` shows `ε_step = 0` is satisfiable at any rejection rate at attempt granularity, and `attemptSamplerLoss_lt_falconScale` pins the attempt-level term at `qSign = 2^64`, `ε_step = 2^-80`, `pRej = 1/2`, `maxAttempts = 128` below `2^-14`.

## The witness (`LatticeCrypto/Falcon/NonVacuity.lean`)

Replaces the `h = 0` instance. Ring degree `n = 2` at `q = 12289`; key `f = −106 + 32x`, `g = 5 + 2x`, `F = −5 + 2x`, `G = −106 − 32x` with `fG − gF = q` checked on the schoolbook `intPolyMul`, `f · h = g` for `h = 8915 + 8488x ≠ 0`, `validKeyPair` proved. Both basis rows have squared norm exactly `q`; the verifier bound is `betaSquared = q`.

The primitives run Falcon's own pipeline (`toFFTTarget` → `Primitives.ffSampling` at depth `0` → `fromFFTPreimage`) with the exact degree-2 packed FFT, and `samplerZ μ σ = round μ + δ` with `δ` uniform on `{−1, 0, 1}`. `toy_trapdoorSample` proves the trapdoor sampler is the Babai round-off of the target perturbed by a uniform point of `{−1, 0, 1}⁴` in basis coordinates; `toy_signAttempt` adds the norm check; `toy_eval_sOf` proves every candidate is a preimage of the target.

| Fact | Declaration | Value |
|---|---|---|
| per-attempt rejection, every target | `toy_attemptRejectBound` | `pRej ≤ 80/81` |
| accepted box points at `c = 0` | `card_accepts_zero` (kernel `decide`) | `9` of `81` |
| one-shot acceptance at `c = 0` | `toy_oneShot_accept_zero` | `1/9` |
| joint hypothesis bundle | `falcon_attempt_hyps_inhabited` | `ε_step = 0`, `pRej = 80/81` |
| headline instantiated, any query-bounded forger | `toy_euf_cma_security` | loss `qSign · (80/81)^maxAttempts` |
| attempt-level loss at Falcon's budget | `toy_samplerLoss_lt_half` | `< 1/2` at `qSign = 2^64`, `maxAttempts = 5265` |
| one-shot budget at the same key | `toy_oneShot_epsStep_ge_half` | every admissible `ε_step ≥ 1/2` |
| one-shot loss at the same key | `toy_oneShot_samplerLoss_one_le` | `qSign · ε_step ≥ 1` for all `qSign ≥ 2` |

The ideal PSF is the attempt's accept-conditional; the new generic `ProbComp.ResamplesTo.uniform_filter` (uniform rejection sampling resamples to the accept-conditional) discharges `idealAttemptResamples`.

## Not claimed

- Falcon+'s fresh salt per retry is not modelled: the GPV wrapper draws the salt once per signature and the loop retries at that target. The fresh-salt accounting ([FGdG+25]'s `C_s` term) is out of scope.
- Verification correctness (`verify` accepting honest signatures) remains out of scope, as the module docstring states.
- `hReg` at the witness is discharged with the ideal sampler's own marginal as domain sampler, which only the basis holder can sample efficiently; at `n = 2` nothing is hard. The witness shows the hypothesis bundle is jointly satisfiable at an instance with the structure of a Falcon key, not at a hard one.
- No concrete bit-security figure. The [FGdG+25] Theorem 1 ingredients and the [Jia+26] tables are removed from the module docstring; a "Relation to the literature" paragraph states what differs.

## Checks

- `lake build VCVio LatticeCrypto Extern HashSig Examples VCVioWidgets` clean; every touched file passes `-Dweak.linter.mathlibStandardSet=true` with no warnings.
- All new declarations are axiom-clean (`propext`, `Classical.choice`, `Quot.sound`); no `native_decide`, no `bv_decide`.
- `scripts/check-pmf-boundary.sh`, `scripts/check-extern-isolation.sh`, `scripts/check-interop-isolation.sh`, `lake exe axiomsweep --check` pass.


VCVio/CryptoFoundations/GPVHashAndSign.lean
```diff
@@ -93,6 +96,41 @@ def Correct (psf : PreimageSampleableFunction PK SK Domain Range) : Prop :=
     psf.eval pk x = t ∧
       psf.isShort x = true
 
+/-- A PSF is correct *at a fixed key pair* `(pk, sk)` if the trapdoor sampler at that key always
```
dtumad: @alik-eth Would you be able to split this into maybe a GPVHashAndSign directory and then split this into a number of different files. Currently it is quite hard for VSCode to keep the whole file loaded, and makes it hard I think for consumers to find the important definitions

dtumad (changes requested):

VCVio/CryptoFoundations/GPVHashAndSign.lean
alik-eth: Done in e5a8fd0e — split into twelve phase modules under `VCVio/CryptoFoundations/GPVHashAndSign/` (`Basic` → `CollisionTelescope` → `Factorization` → `GameRuns` → `TapeFactorization` → `FlagHandlers` → `CombinedHandler` → `GameIdentification` → `VerificationBridge` → `TrapCount` → `EmbedIndex` → `Security`), each 200–2000 lines, following the file's proof-phase section structure. The headline bounds and the reduction definitions a consumer needs are in `Basic` and `Security`. The original module path stays as an umbrella importing all pieces, so imports and fully-qualified names are unchanged (Falcon and the non-vacuity example needed no edits; `#print axioms` on the headlines is unchanged).

alik-eth:
Rebased onto main now that #465 is merged — the diff is down to this PR's own commits. The file split requested above is in `e5a8fd0e` (now `9d0f7dfe` tip after rebase): twelve phase modules under `GPVHashAndSign/` with the original module path kept as an umbrella, all names and consumers unchanged, full CI set green locally.

alexanderlhicks:
I have a local branch with some Flacon stuff on it, let me check what progress I'd made and whether there's anything complementary.

alik-eth:
@alexanderlhicks Happy to coordinate — here's a quick map to make diffing your branch easy.

**What this PR covers:** the generic GPV/PFDH EUF-CMA bounds (`GPVHashAndSign.euf_cma_split_bound` / `euf_cma_collision_bound`, sorry-free) and the Falcon instantiation `Falcon.euf_cma_security` — a conditional reduction terminating at `ntruPSFCollisionProblem` (two distinct short preimages with equal image), plus joint non-vacuity witnesses for both hypothesis sets. Heads-up if your branch touches the old `GPVHashAndSign.lean` monolith: it was just split into twelve modules under `VCVio/CryptoFoundations/GPVHashAndSign/` (the old module path remains as an umbrella, so imports and fully-qualified names are unchanged — conflicts would be textual only).

**Deliberately out of scope here — i.e. the complementary slots:**
- the `ntruPSFCollisionProblem → ntruSISProblem` kernel-vector bridge (`x ≠ x'` ↦ nonzero `x − x'` with `‖·‖₂² ≤ 4·betaSquared`); `ntruSISProblem` ships as a documented dormant target, so this is small and self-contained if you have it;
- signing/keygen **correctness** (`verify_sign_correct` and the retry-loop signer) — removed here as unfinished; it routes through the floating-point inverse-FFT rounding (`FPRBridge`/`ApproxArith`);
- discharging the `hTransport` hypothesis via `SamplerQuality`/`renyiDiv` (the [FGdG+25]/[Jia+26] Rényi finite-precision analysis — the structure is in `Security.lean`, currently unconsumed);
- NTRU key generation (`Concrete/NTRUSolver.lean` still carries sorries on main).

Anything of yours in those slots composes with this PR rather than colliding. If you did touch the same files, happy to coordinate a rebase order.

quangvdao (changes requested):
Thank you for the substantial work here. The generic GPV game-hopping development is valuable, sorry-free, and much easier to review after the split into phase modules. I am requesting changes because the Falcon-facing theorem currently claims more than its assumptions establish. I do **not** expect this PR to prove security of every concrete Falcon implementation detail (compression, floating point, exact sampler implementation, etc.). A meaningful theorem for a reasonably faithful idealized Falcon model—exact real arithmetic and an ideal Gaussian sampler are fine—is a good target.

The main changes I need are:

1. **State the scope as an idealized Falcon/GPV model.** `falconSignatureAlg` is currently the generic `GPVHashAndSign` scheme with the full `(s₁,s₂)` preimage, rather than the concrete `Falcon.Signature`/sign/verify path. This can still be a reasonable abstraction, but the theorem and documentation need to say precisely which Falcon features are modeled and which are idealized. Concrete encoding/compression and floating-point implementation security are not blockers for this PR.

2. **Replace `hTransport` with materially weaker assumptions and prove the adaptive composition.** As written, `hTransport` assumes the central concrete/approximate-sampler-to-ideal-adversary step wholesale: it directly supplies an ideal adversary with the desired advantage inequality, queried-point property, and query bound. The declared `SamplerQuality`/Rényi infrastructure is not used. Please formulate a local sampler approximation hypothesis (or another appropriately weak per-key/per-call condition) and prove how it accumulates across adaptive signing queries and retries. If the Rényi API is not the right route, remove it from this PR rather than presenting it as the justification.

3. **Finish or accurately expose the two remaining cryptographic residuals.** The current result stops at `ntruPSFCollisionProblem`, which is the GPV collision predicate repackaged as a problem, while the standard algebraic NTRU-SIS/kernel-vector bridge is not established. The exact-match/programmed-preimage term is also left unbounded. Please either:
   - prove the collision-to-standard-NTRU relation and an appropriate min-entropy/regularity bound for the exact-match term; or
   - export a deliberately intermediate theorem whose name and documentation expose both residual assumptions without calling it Falcon security “down to NTRU.”

   The latter is acceptable if it gives us a clean, meaningful staging point. The important thing is that the assumptions be strictly weaker than the conclusion we want, rather than restatements of the hard reduction steps.

4. **Formalize the queried-forgery WLOG step.** `ForgesQueriedPoint` currently restricts every supported adversary output, including losing outputs. Please implement the standard transformation that appends the final RO query and account for the extra query, or clearly keep the generic theorem restricted and provide the unrestricted corollary through that transformation.

5. **Reframe the non-vacuity example.** The degree-one, unit-salt, `samplerLoss = 1` witness proves logical inhabitance of the hypothesis type, not meaningful Falcon security. Please label it explicitly as a consistency/inhabitance witness and avoid quantitative “non-vacuity” language.

6. **Small correctness/maintenance items.** The prose mentions a `qSign²/(2|Salt|)` collision term while the theorem uses `(qSign+qHash)²/(2|Salt|)`; please reconcile this. Some generic WriterT/state-projection and query-bound lemmas should eventually live below the GPV namespace, though that extraction can be a follow-up if moving them now creates unnecessary proof churn.

There is also no formal PPT/cost preservation yet. Please mention this limitation briefly rather than trying to solve it here. #460 is dtumad's ongoing work toward the machine/PPT infrastructure; until that lands, reviewers should manually check that the constructed reductions are plausibly efficient and do not use exhaustive/noncomputable steps in the cryptographic path.

Finally, #477 is already merged, so please update this branch to current `main` and rerun CI.

My preferred landing shape is: first the generic GPV theorem with its exact restrictions; then a carefully named idealized Falcon instantiation with the weakest assumptions we can reasonably achieve. Concrete implementation fidelity can remain future work.

alik-eth:
@quangvdao Thank you for the careful review — itemized response:

1. **Idealized-model scope**: agreed; documenting now that `falconSignatureAlg` is the abstract GPV scheme over the full `(s₁,s₂)` preimage with an ideal sampler and exact arithmetic, with an explicit list of what is modeled vs. idealized.
2. **`hTransport` decomposition**: agreed this is the substantive gap. Plan: a local per-call sampler-approximation hypothesis with a proven accumulation across adaptive signing queries and retries, replacing the wholesale assumption. The unused `SamplerQuality`/Rényi declarations come out of this PR until they are actually consumed. This is the main remaining campaign (est. multi-week) and I'd sequence it after item 4.
3. **Residuals**: (a) pushing `LatticeCrypto/Falcon/SISBridge.lean` — the collision→kernel-vector translation is now proven: `ntruSISProblem_isValid_sub` (witness level: nonzero from distinctness, kernel from linearity, `4·betaSquared` from the centered-representative bound), `ntruSISProblemKeyed` + `advantage_le_ntruSISProblemKeyed` (advantage transfer at the honest key distribution), and `euf_cma_security_ntruSIS`. The residual honest-key↔uniform-challenge distance is documented as exactly a decisional-NTRU assumption — the deliberately-staged intermediate you describe. (b) Taking the honest-naming route now; an ideal-Gaussian min-entropy bound for the exact-match term as follow-up.
4. **Queried-forgery WLOG**: in progress — design done (the generic theorem keeps its restriction; an unrestricted corollary goes through the append-final-RO-query transform with `qHash + 1` accounting). Est. 1–2 weeks.
5. **Witness reframing**: agreed — relabeling as consistency/inhabitance witnesses, removing quantitative language.
6. Docstring `qSign²` → `(qSign+qHash)²` reconciliation, PPT-limitation note, and the 4.31 rebase: all in the next push.

**Proposal on landing shape**, per your preference: extract the generic GPV development (the `GPVHashAndSign/` modules + its hypothesis-witness example) into its own PR for prompt landing, and let this PR become the idealized-Falcon instantiation stacked on it, carrying items 2/3b/4. If that works for you I'll do the split with the next push.

quangvdao:
sounds good please do the split. more pre each with self contained slice for review is better

alik-eth:
Will do — extracting the generic GPV development (the `GPVHashAndSign/` modules + the hypothesis-consistency example) into its own self-contained PR, with this one rebased on top carrying the idealized-Falcon layer. Scope/witness/docstring items from the review are already staged and will land with the split.

alik-eth:
Split executed as promised: the generic GPV framework is now #478 (self-contained, builds standalone, headlines kernel-checked axiom-clean), and this PR is rebased on top of it carrying only the idealized-Falcon layer — retitled and body rewritten accordingly. Until #478 merges, the diff here still shows the GPV commits; it shrinks to `LatticeCrypto/Falcon/` after a rebase once it lands.

Also addressed in this round, per the review:
- **Scope section** added to `Falcon/Security.lean` ("an idealized Falcon/GPV model": full `(s₁,s₂)` preimages, exact arithmetic, `hTransport` as a single assumed hypothesis, no cost model).
- **Unused Rényi API dropped** (`SamplerQuality`/`HasUniformSamplerLoss`); the `hTransport` docstring now states the per-call decomposition as intended refinement rather than pointing at dead scaffolding.
- **Witnesses reframed** as logical-consistency (inhabitance) witnesses only — the docstrings now say explicitly they carry no quantitative security content.
- **Birthday bound documented uniformly** as `(qSign + qHash)² / (2 · |Salt|)`.
- Toolchain: branch adapted to Lean v4.31.0, full CI set green, zero non-sorry warnings, all headline axioms re-verified.

alik-eth:
Review point 4 (hForge WLOG) is closed on the base branch: #478 now contains the append-forgery-query compiler and the all-adversaries corollary `euf_cma_split_bound_of_queryBound` (no `ForgesQueriedPoint` hypothesis; details in the #478 comment). This branch is rebased on top — the Falcon layer is unchanged.

alik-eth:
Review point 2 (hTransport decomposition) is done — the "intended refinement" promised in the docstring is now formalized, sorry-free:

- **`samplerTransport ε_step`** — the per-call bound: on honestly generated keys, the concrete and ideal trapdoor samplers are within total-variation `ε_step` at every challenge. Named, inspectable, satisfiable (`samplerTransport_one` proves the `ε_step = 1` witness); its concrete discharge routes through the IEEE/opaque-float layer and is documented as such, per the FPRBridge scope.
- **The adaptive accumulation is proven, not assumed** (in #478's new `SamplerTransport` module): swapping the sampler across the whole EUF-CMA experiment costs `qSign · ε_step`, by charging exactly the signing steps of the game's `simulateQ` seam.
- **`Falcon.euf_cma_security_of_samplerTransport`** — the original theorem's frontier with `hTransport` replaced by the per-call bound; `samplerLoss` becomes the derived `qSign · ε_step`.
- **`Falcon.euf_cma_security_of_samplerTransport_queryBound`** — the same via #478's append-query compiler, dropping `ForgesQueriedPoint` entirely: the Falcon frontier is now honest-key correctness/regularity/never-fail, a query bound, and the named per-call sampler idealization, at the usual `qHash + 1` accounting.
- The toy witness now inhabits the full decomposed frontier end-to-end (`toy_euf_cma_security_of_samplerTransport`), consistency-only framing as before.

Honest accounting note (in the docstrings): the TV-additive `qSign · ε_step` bound is looser than the multiplicative Rényi accumulation of [FGdG+25] for large `qSign` — this is the Rényi term read at total-variation distance. The original `euf_cma_security` with the abstract transport hypothesis remains available and unchanged.

All new declarations `[propext, Classical.choice, Quot.sound]`; full CI set green.

alik-eth:
Review point 3(b) — the exact-match/min-entropy residual — is closed (the 3(a) NTRU-SIS bridge landed earlier via `SISBridge.lean`):

- **Generic** (#478, new `GPVHashAndSign/MinEntropy.lean`): `programmedPreimageAdvantage ≤ trapdoorGuessingProbability` — any programmed-preimage adversary is bounded by the trapdoor sampler's pointwise-maximum mass over honest keys and challenges (the guessing probability `2^(−H∞)`). This is tight: the game's challenge is challenger-drawn uniform and the bound is met with equality by an argmax-guessing adversary.
- **Falcon**: `idealSamplerGuessBound εpp` — the named per-call min-entropy assumption (a genuine distributional property of the ideal sampler, strictly weaker than the discharged conclusion, per the review's criterion), with the trivial witness proven and the literature value documented (GPV08 Lemma 2.10: `2^(−H∞)` of the coset Gaussian, negligible at Falcon parameters). Composed as **`Falcon.euf_cma_collision_security`** on the strongest frontier — the collision-only bound `SIS + (qSign + qHash + 1)·εpp + birthday + qSign·ε_step`, with no `ForgesQueriedPoint`, decomposed transport, and the exact-match existential collapsed away.
- **Recon outcome disclosed**: the current `DiscreteGaussian` module has no pointwise-mass upper bound, so the concrete numeric `εpp` awaits that theory; the two missing lemmas are specified precisely in the docstring rather than assumed silently.
- The toy witness inhabits the full corollary frontier end-to-end; its deterministic sampler makes `εpp = 1` *exactly tight* there (zero min-entropy), which the docstring notes as the honest degenerate case.

With this, every item from the review is addressed: 1 (headline soundness via the split bound), 2 (transport decomposed), 3(a) (SIS bridge), 3(b) (this), 4 (hForge WLOG formalized). All new declarations `[propext, Classical.choice, Quot.sound]`; full CI set green.

LatticeCrypto/Falcon/Security.lean
```diff
@@ -287,25 +230,69 @@ With exact arithmetic (infinite precision), `r_p = 1` and the sampler loss vanis
...
+banked witness `Falcon.Concrete.NonVacuity.falcon_eufcma_hyps_inhabited` certifies their joint
+consistency only at a degenerate instance — degree one, `h = 0`, and a `betaSquared` inflated until
+shortness accepts everything — and therefore exhibits no honest-key NTRU geometry and no
+quantitative content.
+
+The ideal sampler `idealPSF` shares the deterministic `eval`/`isShort` of `falconPSF`
+(`hEval`/`hShort`); `hTransport` carries the finite-precision concrete→ideal gap as the [FGdG+25]
+Rényi term `samplerLoss`, assumed here in the same way MLWE/SIS hardness is assumed. The collision
+branch is discharged by `collisionFindingAdvantage_eq_ntruPSF`. -/
 theorem euf_cma_security
```
dtumad, line 256: This is the version that still assumes `hTransport` wholesale, while `euf_cma_security_of_samplerTransport` derives it from the per-call bound. The short name is what doc search and downstream imports will find first — I would put it on the strongest statement instead.

LatticeCrypto/Falcon/Security.lean
```diff
@@ -287,25 +230,69 @@ With exact arithmetic (infinite precision), `r_p = 1` and the sampler loss vanis
...
+    (hCorrect : ∀ pk sk, (pk, sk) ∈ support hr.gen → idealPSF.CorrectAt pk sk)
+    (hReg : ∃ domainSample : PublicKey p → ProbComp (Rq p.n × Rq p.n),
+      ∀ pk sk, (pk, sk) ∈ support hr.gen →
+        𝒟[(do let s ← domainSample pk; pure (idealPSF.eval pk s, s)
+              : ProbComp (Rq p.n × (Rq p.n × Rq p.n)))] =
+        𝒟[(do let c ← ($ᵗ (Rq p.n)); let s ← idealPSF.trapdoorSample pk sk c; pure (c, s)
+              : ProbComp (Rq p.n × (Rq p.n × Rq p.n)))])
+    (hNeverFail : ∀ pk sk, (pk, sk) ∈ support hr.gen →
+      ∀ c, NeverFail (idealPSF.trapdoorSample pk sk c))
+    -- Finite-precision sampler transport ([FGdG+25] Rényi term): swapping the concrete signing
```
dtumad, line 280: `ε_step` is bounded below by the concrete sampler's rejection mass, independently of precision: `CorrectAt` forces the ideal sampler's support to be short, and `falconPSF.trapdoorSample` has no norm check. See the compiled lower bound in the main comment.

dtumad (commented):
Re-reviewed at `403b28f9`, looking only at the Falcon-specific delta rather than the #478 material underneath it.

The two structural asks from the previous round are answered. The scope is documented as an idealized GPV model rather than the concrete signer, and `hTransport` is no longer the only story: `euf_cma_security_of_samplerTransport` derives that monolithic package from the per-call bound `samplerTransport`, accumulating adaptively across the signing queries, with `euf_cma_security_of_samplerTransport_queryBound` discharging `ForgesQueriedPoint` for one extra hash query. The collision branch building a nonzero norm-bounded NTRU-SIS witness, the exact-match term being priced rather than dropped, and keeping keyed honest-key NTRU-SIS separate from the uniform-challenge assumption all look right to me.

Three things I would fix before this is read as the Falcon security theorem.

**1. The unqualified name is on the weakest theorem.** `euf_cma_security` (`Security.lean:256`) is still the version that takes `hTransport` wholesale — the very assumption the previous round asked to have derived. The derived versions carry the longer names. In doc search and in downstream imports the short name wins, so I would flip that: make the strongest statement the unqualified endpoint (or give it an `idealized`/`gpv` qualifier and retire the short name), and let the wholesale-assumption version read as internal.

**2. `ε_step` is not a precision term, and the docs say it is.** The comment above `hTransport` calls it "Finite-precision sampler transport ([FGdG+25] Rényi term)". But `CorrectAt` (`GPVHashAndSign/Basic.lean:63`) requires *every* supported ideal sample to satisfy `isShort`, while `falconPSF.trapdoorSample` (`Scheme.lean:180-183`) is one-shot with no norm check — `signAttempt` applies the check outside it and returns `none` to retry. So the ideal sampler is short-conditioned and the concrete one is not, and the TV distance between them is bounded below by the concrete sampler's rejection mass, whatever the arithmetic precision:

```lean
theorem samplerTransport_ge_rejection
    (hStep : samplerTransport p prims hr idealPSF ε_step)
    (hShort : ∀ x, idealPSF.isShort x = (falconPSF p prims).isShort x)
    (hCorrect : ∀ pk sk, (pk, sk) ∈ support hr.gen → idealPSF.CorrectAt pk sk)
    (pk : PublicKey p) (sk : SecretKey p) (hk : (pk, sk) ∈ support hr.gen) (c : Rq p.n) :
    Pr[= true | (fun x => !(falconPSF p prims).isShort x) <$>
        (falconPSF p prims).trapdoorSample pk sk c].toReal ≤ ε_step
```

(axiom-clean; the proof is `tvDist_map_le` through `isShort` plus `abs_probOutput_toReal_sub_le_tvDist`, using `CorrectAt` to zero the ideal side). Since the accumulated loss is `qSign · ε_step`, and the per-attempt rejection probability is a constant — that is why the signer has a retry loop at all — this premise cannot be discharged at a precision-sized `ε_step`. The implication is still sound, but what discharging it costs is quite different from what the docs describe. Either let `idealPSF` be the short-conditioned sampler and price the conditioning explicitly (an unconditioned coset-Gaussian point-mass bound then also needs an acceptance-probability factor), or weaken `CorrectAt` so the ideal sampler may also produce long preimages and handle the retry loop at the scheme level.

**3. The module docstring promises the paper's theorem.** This one is inherited rather than introduced here — it is already on `main` (`Security.lean:31`, `:48` there) — but this PR is the one building the layer it purports to describe, so it seems like the moment. The file says "The precise bound follows [FGdG+25] Theorem 1", heads a section "The exact theorem ([FGdG+25] Theorem 1, adapted)" displaying a bound with retry count `C_s`, per-attempt acceptance probability `p` and Rényi orders `a_p, a_u`, and then reproduces the [Jia+26] Table 6 bit-security numbers. What is proved is an additive GPV birthday / min-entropy / TV bound with none of those ingredients. The papers are good motivation and the parameters are worth recording, but as written a reader will take the displayed formula and the table to be what this file establishes.

Minor, on debt accounting: `keyGenFromSeed`, `sign`, and `verify_sign_correct` are removed rather than proved. Retiring them may well be right — they were unused and the old obligations were not the ones you want — but the description and baseline change should classify them as retired debt and point at the concrete-correctness follow-up, so they are not counted as three closures.

quangvdao (changes requested):
Re-reviewed at `403b28f99031ff128229e1c8914805f7606d3b24`, focusing on the Falcon-specific slice above #478.

The PR has made substantial progress: it now states an idealized GPV scope, proves adaptive accumulation from a per-call sampler bound, provides the append-query endpoint, bridges distinct-preimage collisions to a norm-bounded NTRU-SIS witness, and prices the exact-match branch through an explicit guessing-probability assumption. The current build and axiom gates are green. Three blockers remain before this should be presented as the Falcon security theorem:

1. **P1 — `ε_step` is not merely a finite-precision sampler error.** `idealPSF.CorrectAt` forces every supported ideal output to satisfy `isShort`, while `(falconPSF p prims).trapdoorSample` is the one-shot unconditioned sampler; the norm check and retry live outside it in signing. Therefore any `samplerTransport ... ε_step` is bounded below by the concrete sampler's rejection mass, regardless of floating-point precision. The theorem implication is formally sound, but the current documentation and quantitative interpretation as the [FGdG+25] precision term are not. Either compare like-conditioned samplers and price the acceptance factor explicitly, or allow long ideal outputs and model the retry loop at scheme level. Until then, `qSign·ε_step` cannot be read as the advertised precision-only loss.

2. **P1 — the module overview promises a different theorem than the file proves.** It says the precise bound follows [FGdG+25] Theorem 1, displays the Rényi/retry/acceptance formula, and reproduces concrete bit-security numbers. The proved endpoint is an additive GPV collision/min-entropy/TV bound with none of the displayed `C_s`, acceptance, Rényi-order, or numeric-instantiation ingredients. Keep that material explicitly as motivation/deferred comparison, or replace the headline with the additive theorem actually established here.

3. **P2 — make the strongest public endpoint the discoverable one.** The unqualified `euf_cma_security` still takes the monolithic `hTransport` package wholesale. The stronger all-query-bounded theorem is `euf_cma_security_of_samplerTransport_queryBound`, and the collision-only endpoint is stronger again. Rename/reorder or clearly qualify the wholesale theorem so doc search does not lead consumers to the weakest assumption surface.

Landing order also matters: this branch contains the entire #478 stack. Merge #478 first, then rebase this PR so the review/merge surface is the Falcon-only delta. The retired `keyGenFromSeed` / `sign` / `verify_sign_correct` cluster should be recorded as retired correctness debt pointing to its concrete-correctness follow-up, not counted as closed proof obligations.

I am keeping changes requested until the sampler semantics and theorem/documentation contract agree.

dtumad:
Deep semantic audit across the open PR set (semantics only). This extends the 08-17/08-23 review threads with one strengthened finding; the three blockers from those threads remain unaddressed at head `403b28f` (no pushes since 08-11).

**BLOCKING (strengthened): the small-`ε_step` regime of `samplerTransport` is provably empty, so every derived endpoint is trivially true at Falcon's own advertised parameters.** The earlier review noted `samplerTransport_ge_rejection`; the consequence is sharper than a documentation defect:

- `hCorrect` forces `support (idealPSF.trapdoorSample pk sk c) ⊆ {x | isShort x}` (via `CorrectAt` + `hShort`) on every honest key.
- `falconPSF.trapdoorSample` (`Scheme.lean:180–183`) is one-shot `ffSampling` with **no norm check** — the check lives outside it in `signAttempt`.
- Total variation against a distribution with zero mass on `¬isShort` is at least the concrete sampler's own mass there, so any `ε_step` satisfying `samplerTransport` obeys `ε_step ≥ p_reject > 0` — **structurally, independent of floating-point precision**. The `samplerTransport` docstring (lines 395–400) frames the obstruction as `Float` opacity and says `ε_step = 0` "states that the two samplers agree exactly"; that framing is wrong — no FP-analysis improvement can ever push `ε_step` below the per-call rejection probability, because the two sides are conditioned differently.
- The conclusion term is `ENNReal.ofReal (qSign * ε_step)`. At the module docstring's own budget `qSign = 2^64`, the right-hand side is `≥ 1` for any `p_reject ≥ 2^-64`; Falcon's per-attempt rejection probability is a small constant, orders of magnitude above that. So `euf_cma_security_of_samplerTransport`, `…_queryBound`, and `euf_cma_collision_security` hold vacuously in the only regime anyone would instantiate them.

The fix is one of the two already named in review, now with a concrete reason to pick: either (a) state the transport between the *accepted-output* (post-rejection) distributions and pay the acceptance factor explicitly — the shape the Rényi literature actually proves — or (b) weaken `CorrectAt` on the ideal side and model the retry loop at scheme level. (a) composes with what #514 will eventually provide; (b) also repairs the next item.

**BLOCKING (from the earlier threads, still open):**

1. **The modelled scheme is not Falcon's signer.** `falconSignatureAlg` signs with a *single* `trapdoorSample`, no norm check, no retry — so the modelled scheme emits non-verifying signatures with constant probability, and the PR deletes the correctness cluster rather than proving it. An EUF-CMA adversary against real Falcon sees a different signing-oracle distribution, and no transformation is supplied. This is a scheme gap, not a precision gap.
2. **Hypothesis-bundle satisfiability is witnessed only at a degenerate instance.** `Extern/Falcon/NonVacuity.lean` uses `toyPk := { h := 0 }`, where `eval pk (s₁,s₂) = s₁` and the collision problem is solved by `(x,0),(x,1)` — logical consistency only, per `docs/agents/gotchas.md` on relation-pinning. Nothing shows `hCorrect ∧ hReg ∧ hNeverFail` jointly satisfiable anywhere the collision problem is plausibly hard.
3. The unqualified name `euf_cma_security` still sits on the weakest (existential-`hTransport`) theorem; the strong endpoints carry the qualified names. Swap.

**Not blocking, and good:** `SISBridge.ntruSISProblem_isValid_sub` is correct (centered-representative wraparound handled), and keeping decisional NTRU named rather than silently assumed is right.

Coordination: this branch contains #478 verbatim — after #478 lands, rebase to the Falcon-only delta so the review surface is `LatticeCrypto/Falcon/` + `Extern/Falcon/NonVacuity.lean`. The module docstring also still promises the [FGdG+25] Thm 1 ingredients (`C_s`, Rényi ratios, per-attempt acceptance) that appear nowhere in the proved bound, and reproduces [Jia+26] Table 5/6 numbers as if established — trim to what is proved.

alik-eth:
Rebuilt on #478's head (`gpv188-generic`), Falcon-only delta on top; head `80361eed`. Review surface: `LatticeCrypto/Falcon/{Scheme,Security,SISBridge,NonVacuity}.lean` plus the new generic `VCVio/OracleComp/Constructions/RetryLoop.lean`.

**Strengthened blocker (small-`ε_step` regime empty).** The observation is now a theorem in the development instead of a docstring claim: `oneShot_rejection_prob_le_of_samplerTransport` proves `1 − Pr[isShort | falconPSF.trapdoorSample pk sk c] ≤ ε_step` for every per-draw budget admissible on honest keys, `samplerTransport_zero_forces_accept` the `ε_step = 0` case, and `oneShot_samplerLoss_one_le_of_samplerTransport` that `qSign · ε_step ≥ 1` once `qSign · p_reject ≥ 1`. The `samplerTransport` docstring no longer blames `Float` opacity.

The headline moved to attempt granularity, which is (a) and (b) together:

- `falconRetryPSF` / `falconRetrySignatureAlg` model the signer as `signAttempt` — one `ffSampling` draw *plus the norm check* — repeated up to `maxAttempts` times at the hashed target, with a default output on exhaustion (`ProbComp.retryToDefault`).
- `attemptTransport` compares whole attempts, both sides allowed to reject, so it is a transport between post-norm-check distributions; `idealAttemptResamples` says the idealized attempt is a rejection sampler for the ideal PSF (`ProbComp.ResamplesTo`: replacing a rejection by a fresh draw of `idealPSF.trapdoorSample` reproduces its law); `attemptRejectBound` names `pRej`. `CorrectAt` on the ideal side is untouched.
- `RetryLoop.lean` proves the two generic facts: per-attempt closeness accumulates as `ε_step / (1 − pRej)` over the loop (`tvDist_retryToDefault_le_div`) and a resampling attempt's loop is within `pRej ^ maxAttempts` of its conditional (`tvDist_retryToDefault_le_pow`). `tvDist_falconRetryPSF_trapdoorSample_le` composes them, so `euf_cma_security` carries the loss `qSign · (ε_step / (1 − pRej) + pRej ^ maxAttempts)`.
- `ε_step = 0` is satisfiable at any rejection rate (`attemptTransport_self_zero`); `attemptSamplerLoss_lt_falconScale` pins the term at `qSign = 2^64`, `ε_step = 2^-80`, `pRej = 1/2`, `maxAttempts = 128` below `2^-14`.

Not modelled, and said so in the docstring: Falcon+'s fresh salt per retry. The GPV wrapper draws the salt once per signature and the loop retries at that target; the fresh-salt accounting ([FGdG+25]'s `C_s` term) is out of scope here. Verification correctness also remains out of scope.

**Blocker 1 (the modelled scheme).** Addressed by the above: the headline's scheme is `falconRetrySignatureAlg`, whose signing oracle is the norm-checked retry loop. The single-draw `falconSignatureAlg` is kept only for the one-shot theorems.

**Blocker 2 (degenerate witness).** `Extern/Falcon/NonVacuity.lean` is deleted; `LatticeCrypto/Falcon/NonVacuity.lean` replaces it with an instance that has the structure of a Falcon key:

- `n = 2`, `q = 12289`; `f = −106 + 32x`, `g = 5 + 2x`, `F = −5 + 2x`, `G = −106 − 32x` with `fG − gF = q` checked on the schoolbook `intPolyMul` (`toy_ntruEquation`), `f · h = g` for `h = 8915 + 8488x ≠ 0` (`toy_keyRelation`, `toyPk_h_ne_zero`), `validKeyPair` proved. Both basis rows have squared norm exactly `q`; `betaSquared = q`, far below `n · (q/2)²`, so the derived NTRU-SIS target `4q` is far below `q²`.
- The primitives run Falcon's own pipeline (`toFFTTarget` → `Primitives.ffSampling` at depth `0` → `fromFFTPreimage`) with the exact degree-2 packed FFT and `samplerZ μ σ = round μ + δ`, `δ` uniform on `{−1, 0, 1}`. `toy_trapdoorSample` proves the sampler *is* the Babai round-off perturbed by a uniform point of `{−1, 0, 1}⁴` in basis coordinates; `toy_signAttempt` adds the norm check; `toy_eval_sOf` proves every candidate is a preimage of the target.
- It rejects: `toy_attemptRejectBound` gives `pRej ≤ 80/81` at every target (the centre always lands in the ball, `toy_center_accepts`); at `c = 0` exactly `9` of the `81` box points are accepted (`card_accepts_zero`, kernel-checked), so `toy_oneShot_accept_zero : Pr[accept] = 1/9`.
- Joint satisfiability: `falcon_attempt_hyps_inhabited` discharges `hEval ∧ hShort ∧ hCorrect ∧ hNeverFail ∧ hReg ∧ hAttempt (ε_step = 0) ∧ hRej (80/81) ∧ hRes` at once, the ideal PSF being the attempt's accept-conditional (new generic `ProbComp.ResamplesTo.uniform_filter`). `toy_euf_cma_security` instantiates the headline for every query-bounded forger with loss `qSign · (80/81) ^ maxAttempts`; `toy_samplerLoss_lt_half` puts it below `1/2` at `qSign = 2^64`, `maxAttempts = 5265`.
- The same key kills the one-shot frontier: `toy_oneShot_epsStep_ge_half` — any ideal PSF and any `ε_step` admissible for `hShort ∧ hCorrect ∧ hNeverFail ∧ samplerTransport` there has `ε_step ≥ 1/2`, hence `toy_oneShot_samplerLoss_one_le : 1 ≤ qSign · ε_step` for every `qSign ≥ 2`.

What it does not show, in its docstring: `hReg`'s domain sampler is the ideal sampler's own marginal, which only the basis holder samples efficiently, and at `n = 2` nothing is hard. It shows the bundle jointly satisfiable at an instance with the shape of a Falcon key and a real rejection loop, not at a hard instance.

**Blocker 3 (naming).** Swapped: `euf_cma_security`, `euf_cma_collision_security`, `euf_cma_collision_security_ntruSIS` are the attempt-level retry-loop endpoints; the single-draw theorems are `euf_cma_security_oneShot`, `euf_cma_security_oneShot_bytes40`, `euf_cma_security_oneShot_ntruSIS`.

**Docstring.** The [FGdG+25] Theorem 1 ingredients and the [Jia+26] tables are removed from `Security.lean`; a "Relation to the literature" paragraph states what differs (total-variation accumulation of a named per-attempt gap rather than the Rényi accumulation, GPV08's birthday bound rather than `Q_s · (C_s + Q_H) / 2^k`, no bit-security figure).

All new declarations are axiom-clean (`propext`, `Classical.choice`, `Quot.sound`); no `native_decide`. `axiomsweep --check`, the PMF ratchet and both isolation checks pass locally.

alik-eth:
@quangvdao Re-merged on current main (via the updated #478), head 1dcb52c4. Changes since your last review:

- The Falcon headlines are stated for the named GPV reductions, with the domain sampler explicit. In the one-shot forms, the transported ideal-scheme adversary `adv'` is an argument, `hTransport` is its advantage bound, and `hForge'`/`hQ'` are its side conditions. The attempt-level forms bound the reductions of `appendForgeQuery … ⟨adv.main⟩`. The NTRU-SIS corollaries and the toy instantiation follow the same pattern, the toy one with the named sampler `toyDomainSample`.
- Main's fuel-bounded `Falcon.sign` and `verify_sign_correct` from #663 now sit alongside the attempt-level signer.
- The toy NTRU equation is recomputed through `negacyclicMulPure`, unused instance arguments are removed, and the toy witness helpers have docstrings.

Build, style lint and the axiom sweep pass. The Lint job fails only on `usesRetiredProbability`; the policy question on that is open with dtumad on #478.

PR #467, opened 2026-07-02
feat(MLDSA): EUF-CMA via CMA→NMA over the short-secret MLWE model (#228)

alik-eth:
Closes #228.

EUF-CMA security of the proof-level ML-DSA scheme via the Fiat-Shamir-with-aborts CMA-to-NMA reduction (Theorem 3/Lemma 7, CRYPTO 2023), stated over the **corrected short-secret Module-LWE model** per review.

> **Stacked content:** the `Primitives.Laws` repair + concrete witness is #479 and the WithAbort expectation-algebra roots landed in #465; this branch carries #479's files byte-identically — except `Concrete/NonVacuity.lean`, where the witness certificate is re-pointed at the live short-model consumers — and rebases (take-ours on that one file) once it merges.

## The corrected model (review points 1 & 2)

- **`mldsaMLWEShort` / `mldsaMatrixMLWE`** — the MLWE assumption with secrets and errors uniform on the η-bounded box (`sampleShortVec`), seed-based and uniform-matrix forms. The uniform-secret `mldsaMLWE` (information-theoretically zero advantage — the review's catch) is **retired**, along with `honestSamplingSlack` and the old headline chain.
- **Idealized short-key generation** `keygenShort` samples `ρ`, `K`, `(s₁, s₂)` independently; the key-swap hop against it is an **exact identity** (no statistical slack). The material-based relation `validKeyPairShort` replaces the ∃-seed relation on the short path, and **satisfiability certificates are banked**: `keygenShort_generable`, `keygen0_generable` — every relation/pinning hypothesis pair in the headlines is inhabited by a kernel-checked witness.
- **Named idealizations with inspectable content** replace caller-supplied inequalities: `expandAIdealization` (XOF-as-random-matrix; discharges the seed→matrix bridge via the proven `advantage_mldsaMLWEShort_le_matrix` + `matrixLift`) and `expandSReplacement` (PRG reading of ExpandSeed/ExpandS against the correct box marginal). Both docstrings disclose the unrestricted-quantifier caveat pending the cost model (#460).

## Headlines (all `[propext, Classical.choice, Quot.sound]`)

- `nma_security_short` — EUF-NMA ≤ MLWE(+εbridge) + SelfTargetMSIS, at `keygenShort`; needs **no `Primitives.Laws`**.
- `nma_security_short_matrix` — the literature-facing packaging at the uniform-matrix problem (+εA).
- `nma_security_fips` — the FIPS seed-derived keygen connected to the short model at cost `+εPRG` (`expandSReplacement`); the cross-scheme bridge is proven, not assumed.
- `euf_cma_security_of_nma_short` — the CMA composition (bound: MLWE + STMSIS + `cmaToNmaLoss`), plus the re-pointed asymptotic corollary.

## SelfTargetMSIS: point 4 became a soundness fix

Formalizing the requested algebraic bridge **proved the tailored problem was easy**: `SelfTargetMSIS.Problem.isValid` never received the hash preimage, so nothing bound the recomputed commitment to the hashed pair — at sampled parameters acceptance degenerated to two norm gates (`z = 0` + empty hint won outright; this pre-dates the redesign). Fixed at the API: `isValid` now takes the `HashInput`, the experiment threads the cache-consistent preimage, and `mldsaSTMSISShort` carries the binding conjunct. The corrected characterization `mldsaSTMSISShort_isValid_expandA_iff` shows winning now requires a random-oracle fixed point — the genuine self-target task. The full algebraic iff (`stmsisAlgebraicSolution`, `mldsaSTMSISShort_isValid_iff`) states the matrix equation and norm bounds explicitly; the residual distance to the `[A | I]` normal form is documented in-file.

## Honest scoping

- **HVZK obstruction (proved, documented)**: under the material relation, `t₀` is not a function of `pk` (a small `s₂` perturbation preserves `t₁` but shifts `t₀`), so no exact-on-accept simulator covers all short-valid pairs with `t₁`-only public keys — consistent with the literature stating HVZK with full `t` public. The CMA wrapper therefore keeps HVZK as an abstract satisfiable hypothesis; the seed-relation HVZK theorem (`idsWithAbort_hvzk_real`) remains for the FIPS path.
- Scope section documents the idealized proof-level model (point 5); no PPT/cost tracking yet (#460) — the asymptotic statement quantifies numerical sequences only (point 3).
- Module split per point 6: WithAbort/Security → 10 phase modules, GhostBodies → 5.

## Verification

Full CI library set green on v4.31.0; zero non-sorry warnings; every headline and certificate kernel-checked; the `SIS` namespace (Falcon's) byte-unchanged by the API fix.


alik-eth:
Rebased onto main now that #465 is merged — the diff is down to this PR's own commits. Full CI set green locally at the new tip.

quangvdao (changes requested):
Thank you for taking on a very difficult proof chain. There is substantial reusable work here—the Fiat–Shamir-with-aborts hybrids, deferred-sampling machinery, encoding and sampling laws, and ring-norm lemmas are all valuable. I am requesting changes because the current ML-DSA headline relies on a serious distributional misformalization and several assumptions that are too close to assuming the intended reduction.

The blocking issues are:

1. **Correct the MLWE distribution.** `mldsaMLWE` samples both `s₁` and `s₂` uniformly from their entire `RqVec` carriers. Uniform additive error makes `A·s₁+s₂` itself uniform, so this decisional problem has zero distinguishing advantage for an information-theoretic reason. It is not the short-secret/small-error Module-LWE assumption used for ML-DSA.

   Please define the ideal problem using explicit short secret/error samplers matching the intended ML-DSA/literature distribution, and then establish the appropriate bridge to the standard matrix-based MLWE problem. `hMlweBridge` should become a genuine, weaker embedding/assumption with inspectable content, not a caller-supplied inequality that carries the whole reduction.

2. **Redesign `honestSamplingSlack`.** It currently compares deterministic `ExpandSeed/ExpandS` output with independent uniform full-ring vectors. For concrete ML-DSA this TV distance is near one: the real output is supported on η-bounded vectors, while the comparison distribution has full-ring support. Thus a negligible `idealGap` is not plausibly satisfiable, and `idealGap = 1` makes the bound vacuous.

   Please compare against the correct short-secret marginal and model the ideal XOF/random-function step explicitly, or introduce a clearly computational XOF/PRG replacement assumption. Merely labeling the current gap as a ROM assumption is not sufficient when `prims` is a fixed deterministic function.

3. **Fix the asymptotic/non-vacuity claims.** `euf_cma_security_asymptotic_real_satisfiable` only chooses numerical negligible sequences; it does not instantiate `honestSamplingSlack`, MLWE/STMSIS, `hMlweBridge`, the scheme family, or the other security hypotheses. Please remove or rename it to describe only the numerical loss regime. Likewise, `idealPrims` exposes the key-generation seed as public `ρ`; it is useful as a logical consistency witness for `Primitives.Laws`, but not as a faithful or secure ML-DSA instantiation. Document and name it accordingly.

4. **Expose a standard SelfTargetMSIS target.** `mldsaSTMSIS.isValid` is presently defined through the ML-DSA identification verifier. The extraction is useful, but to justify the literature-facing hardness claim we need a bridge to an explicit algebraic SelfTargetMSIS relation with the matrix equation and norm bounds, or the theorem must clearly stop at this tailored intermediate problem.

5. **Scope the theorem to the model actually covered.** The result is for the proof-level `FiatShamirWithAbort (identificationScheme …)`, not yet the full FIPS signing/encoding path. That is acceptable for this stage: concrete FIPS implementation security is not a prerequisite. Please describe it as security of the idealized proof-level ML-DSA model and list the idealizations, rather than “end-to-end ML-DSA.” The objective is a meaningful unforgeability theorem for a reasonably accurate model under genuinely weaker assumptions.

6. **Split the generic engine into reviewable modules.** An 8k-line `WithAbort/Security.lean` plus the 2.7k-line `GhostBodies.lean` is too difficult to maintain and semantically review. Please separate at least retry/cache primitives, game definitions/identification, Sign→Prog, Prog→Trans/deferred sampling, Trans→Sim, managed-NMA bridge, and final assembly. The concrete algebra/law results are also good candidates for an independent PR if that helps preserve progress while the security statement is redesigned.

There is also no formal PPT/cost preservation yet. Please mention this briefly rather than treating it as a blocker here. #460 is dtumad's work toward the machine/PPT layer; until that infrastructure is available, reviewers need to manually verify that each constructed reduction plausibly preserves polynomial time and does not rely on exhaustive/noncomputable computation in the cryptographic path. In particular, avoid calling the current unrestricted-adversary asymptotic theorem a theorem about “poly-time adversaries.”

Finally, #477 is already merged, so please update this branch to current `main` and rerun CI.

The key requested outcome is not more assumptions around the present statement: it is a corrected short-distribution model and theorem statements whose hypotheses are strictly weaker, meaningful, and plausibly instantiable for the idealized ML-DSA scheme.

alik-eth:
@quangvdao Thank you — points 1 and 2 are genuine catches and I want to acknowledge them plainly: with both secret and error uniform over the full ring, `mldsaMLWE` is information-theoretically zero-advantage as a decisional problem, so `hMlweBridge` was silently carrying the entire hardness step; and `honestSamplingSlack` against full-ring uniform vectors makes a small `idealGap` unsatisfiable for any η-bounded instantiation. Neither is fixable by relabeling — agreed the fix is the short-distribution redesign.

Itemized plan:

1. **MLWE distribution**: define explicit short secret/error samplers matching the ML-DSA literature distribution, restate the ideal problem over them, and make `hMlweBridge` a genuine seed→matrix embedding with inspectable content. Main redesign campaign.
2. **`honestSamplingSlack`**: rebuild the keyswap hop comparing against the correct short-secret marginal, with an explicit XOF/PRG-replacement assumption as the computational step (not a ROM label on a deterministic function). Part of the same redesign.
3. **Renames**: doing now — the asymptotic satisfiability witness gets a name describing only the numerical loss regime, and `idealPrims` is documented (and named) as a logical consistency witness for `Primitives.Laws`, explicitly not a faithful instantiation.
4. **SelfTargetMSIS**: will export the explicit algebraic relation with matrix equation and norm bounds plus a bridge from the verifier-tailored intermediate, or clearly stop at the intermediate with honest naming — folded into the redesign.
5. **Scope docs**: doing now — security of the idealized proof-level ML-DSA model, with the idealization list.
6. **Module split**: doing now, along the cut you list (retry/cache primitives, game definitions, Sign→Prog, Prog→Trans/deferred sampling, Trans→Sim, managed-NMA bridge, assembly).

PPT-limitation note and the 4.31 rebase land with the next push.

**Proposal**, per your suggestion: extract the concrete algebra/laws work (`Ring/Norms` additions, `MLDSA/Concrete/{Encoding,Laws,Rounding,Sampling}`, and the `Primitives.Laws` field corrections they witness) into an independent PR so that progress is preserved while the security statement is redesigned here. If that works I'll do the split with the next push, and sequence the redesign campaign after the #466 items unless you'd rather it first.

quangvdao:
yeah do the split. for MLWE definition, I prefer a centralized, generic definition of LWE that works for general modules and error distribution - one umbrella master definition, then specialized into plain / unstructured LWE, module LWE, ring LWE, different error distributions (even hamming weight error as well), etc. 

So in fact what I'm looking for here is a central `NoisyLearningProblem` definition that can specialize to both LWE and LPN type assumptions, then specialize down to MLWE as used in ML-DSA. Put this in the right place, probably in VCVio core.

This is not the original scope of this PR but I think we should have such a self contained PR to aid all future developments.

alik-eth:
Will do — extracting the concrete algebra/laws slice (`Ring/Norms` additions, `MLDSA/Concrete/*`, the `Primitives.Laws` corrections) as its own PR. On the LWE ask: agreed, and it dovetails with review point 1 — I'll draft a self-contained `NoisyLearningProblem` PR in VCVio core (one master definition over general modules with pluggable secret/error samplers, decision and search variants; specializations to plain LWE / MLWE / RLWE / LPN-style Hamming-weight error), and the corrected short-distribution ML-DSA MLWE then lands as an instantiation of it in the redesign here. Renames and scope docs from the review are already staged.

alik-eth:
First response batch pushed (branch force-updated over the update-branch merge; rebased on current main, Lean v4.31.0, full CI set green, zero non-sorry warnings, `MLDSA.euf_cma_security_of_nma` / `nma_security` re-verified `[propext, Classical.choice, Quot.sound]`):

- **Scope section** added to `SecurityNMA.lean` ("an idealized proof-level ML-DSA model": transcript signatures, seed-based problems, `idealGap` semantics, no cost model — these are not poly-time statements).
- **`idealPrims` → `seedRevealingPrims`** renamed throughout the concrete witness, and the Laws instance is now presented as a logical-consistency (inhabitance) witness only, with no quantitative security content. The satisfiability lemma is renamed `asymptotic_loss_regime_satisfiable` and its docstring states the adversary sequence chooses numerical sequences only.
- **Deduped** the two `WithAbort` private copies of the expectation-algebra lemmas that landed at the EvalDist root in #465.

Still in flight from the review, in order: the `WithAbort/Security.lean` module split, the concrete-laws extraction into its own PR, the central `NoisyLearningProblem` definition in VCVio core (design drafted — the existing `LearningWithErrors.Problem` shape lifts cleanly and specializes to LWE/MLWE/LPN and the short-distribution ML-DSA MLWE), and then the short-distribution redesign of `mldsaMLWE` on top of it.

alik-eth:
Second split executed: the `Primitives.Laws` repair + FIPS-shaped concrete witness is now #479 (self-contained off main: the three-field fix, the `honestSamplingSlack` key-swap threading adapted to main's `SecurityNMA`, and the kernel-checked `mldsa_laws_inhabited` certificate). The laws files there are byte-identical to this branch's, so this PR's diff shrinks to the EUF-CMA development once #479 lands and this branch rebases.

Also pushed here: `WithAbort/Security.lean` (8k lines) is now split into a directory of ten phase modules (`Loss` → … → `Assembly`), and `GhostBodies.lean` into five, both with umbrella modules keeping the import paths — content moved verbatim, zero consumer churn, headlines re-verified axiom-clean after each split.

alik-eth:
The central noisy-learning definition is now #480 (`NoisyLearning.Problem` in VCVio core, specializing to LWE/module-LWE/ring-LWE/LPN, with LatticeCrypto's `LearningWithErrors` as a compatibility re-export). The short-distribution `mldsaMLWE` redesign here will instantiate it with η-bounded samplers.

alik-eth:
@quangvdao The redesign you requested is complete and pushed (13 commits; body rewritten). Point by point:

**1 — MLWE distribution corrected.** `mldsaMLWEShort`/`mldsaMatrixMLWE` sample secrets and errors uniform on the η-bounded box; the uniform-secret problem and its chain are retired outright. The bridge to the standard matrix form is a proven reduction (`advantage_mldsaMLWEShort_le_matrix` via the named `matrixLift`), costed by the quantified `expandAIdealization` assumption — `hMlweBridge` is no longer a caller-supplied inequality, and its canonical discharge is packaged as `nma_security_short_matrix`.

**2 — `honestSamplingSlack` redesigned away.** The headline (`nma_security_short`) runs at honestly-sampled short keys, where the key-swap hop is an exact identity — no statistical slack exists to bound. The deterministic FIPS derivation enters only through `nma_security_fips` at cost `εPRG` via `expandSReplacement`, stated against the correct short marginal with the computational reading (and its limits pending #460) disclosed in the docstring. Every relation/pinning hypothesis is backed by a kernel-checked satisfiability certificate (`keygenShort_generable`, `keygen0_generable`) — we adopted this as a standing gate after the review.

**3 — done in the earlier round** (rename + numerical-sequences-only framing); the asymptotic corollary is now re-pointed at the short headline.

**4 — became a soundness fix.** Formalizing your requested algebraic bridge produced a kernel-checked proof that the tailored problem was *easy*: the `SelfTargetMSIS.Problem.isValid` API never saw the hash preimage, so acceptance at sampled parameters was just two norm gates (a `z = 0` forgery won outright — this pre-dates this PR's redesign). We extended the API with the `HashInput`, added the binding conjunct to `mldsaSTMSISShort`, re-proved the extraction (the binding is exactly the cache read-back fact the proof already had), and the corrected characterization shows winning now requires a genuine RO fixed point. The full algebraic iff (`stmsisAlgebraicSolution`) states the matrix equation and norm bounds; the residual distance to the `[A|I]` normal form is documented rather than papered over. Falcon's `SIS` namespace is untouched.

**5 & 6 — scope section and splits** landed earlier in this round.

One finding worth flagging beyond the review's asks: under the material key relation, `t₀` is provably not determined by the public key, so exact-on-accept HVZK simulation is impossible with `t₁`-only public keys (consistent with the literature stating HVZK with full `t` public). The CMA wrapper keeps HVZK abstract and satisfiable; the obstruction and its counterexample construction are documented in-code.

All headlines and certificates check at `[propext, Classical.choice, Quot.sound]`; full CI set green; zero sorries in the touched files.

dtumad (commented):
Follow-up on the current head: the earlier distributional blockers have been addressed in a substantial and mathematically useful way. In particular, the short secret/error samplers now model the intended eta-bounded box rather than uniform full-ring error; the seed-to-matrix and seed-to-short-vector steps are explicit computational idealizations with the unrestricted-adversary caveat stated; and the numerical non-vacuity result is now scoped as a statement about the loss parameters rather than a concrete security instantiation.

The HashInput API change is also an important soundness repair. Threading the cache-consistent preimage into isValid, and then requiring the recomputed ML-DSA commitment to equal the hashed commitment, closes the trivial solution in the earlier tailored problem. The extra 1/(1-p) on the HVZK term is coherent with the definition here as TV distance for each optional signing attempt; it is the restart-loop cost, whereas the CRYPTO 2023 theorem uses an accepted-transcript simulator.

The main mathematical boundary I would keep visible is the remaining gap between mldsaSTMSISShort and the standard SelfTargetMSIS normal form. Definition 4 in the paper and the published EasyCrypt model use a solution y whose final coefficient block is the challenge and require H(([I_m | A] y, mu)) to equal that block. The new algebraic characterization is valuable, but the current ML-DSA relation still goes through UseHint / encoded HighBits and does not yet supply that reduction to the standard matrix relation. Describing the present endpoint as a tailored self-target problem is accurate; a literature-facing standard-STMSIS headline would benefit from a separate explicit reduction.

I would not try to merge the full branch as one unit yet. It is currently conflicted with main, was validated on Lean 4.31 while main is on 4.32, and combines about 11.7k added lines across the generic retry/coupling engine, concrete ML-DSA laws, corrected MLWE modeling, and the final composition. That makes both regression review and attribution of any merge repair unnecessarily difficult.

A natural incremental route, if useful, seems to be:

1. Extract the generic HashInput threading plus the ML-DSA binding conjunct first. This is independently necessary and also unblocks a sound review of #479.
2. Land the corrected short-MLWE definitions and exact short-key swap as a focused model PR.
3. Land the explicit reduction from the tailored verifier relation to standard SelfTargetMSIS once that normal-form bridge is complete.
4. Review the generic CMA-to-NMA engine separately, then compose the final headline on the smaller landed pieces.

This is meant as a suggested way to preserve and land the strongest parts of the branch, rather than a request to reshape the working branch itself. The current head contains real progress over the earlier review; the remaining concern is chiefly the standard-hardness endpoint and the amount of independent semantic content bundled into one merge.

alik-eth:
@dtumad Thanks for the careful reads — and agreed on the decomposition. Bundling ~11.7k lines into one merge was making both regression review and repair attribution harder than they needed to be; splitting into independently-reviewable pieces is the right call.

**Step 1 is up as #503**: the generic `SelfTargetMSIS.isValid` HashInput threading plus the ML-DSA binding conjunct, off current main. It adds a lightweight characterization `mldsaSTMSIS_isValid_eq_true_iff` (acceptance ↔ verifier acceptance ∧ `hashInput.2 = w'`) so the self-target requirement stays visible and hard to drop. Deliberately narrow — just the binding; `LatticeCrypto`/`Extern` green on 4.32, the two touched declarations kernel-clean.

I'll follow your order for the rest: (2) the short-secret MLWE model + exact short-key swap as a focused PR; (3) the explicit reduction from the tailored verifier relation to the standard SelfTargetMSIS normal form — the `H([I_m | A]·y, μ)` form with the challenge as `y`'s final coefficient block — once that bridge is complete; (4) the generic CMA-to-NMA engine reviewed on its own, composing the headline last on the landed pieces. I'll narrow #467 down as those land rather than merging it as a unit.

On the endpoint: agreed the current relation goes through `UseHint`/encoded HighBits and stops at the tailored verifier form; a literature-facing standard-STMSIS headline needs the explicit normal-form reduction, and I'll keep that scoped honestly until (3) lands. Also noted your point that the extra `1/(1-p)` on the HVZK term is the restart-loop cost vs the CRYPTO 2023 accepted-transcript simulator — that framing is in the docstring.

alik-eth:
Step 2 is up as #504: the short-secret MLWE model + the exact NMA key-swap hop, stacked on #503.

- `sampleShortVec` / `keygenShort(1)` (η-box short secrets, idealized keygen), `mldsaMLWEShort` + `mldsaMatrixMLWE` (the corrected short-secret MLWE, fixing the full-ring-uniform zero-advantage issue), `expandAIdealization` + the proven seed→matrix bridge `advantage_mldsaMLWEShort_le_matrix`;
- `nma_keyswap_hop_short` bounds the NMA-game gap by the `mldsaMLWEShort` advantage as an **exact identity** — no `idealGap`/`honestSamplingSlack` term;
- `validKeyPairShort`/`identificationSchemeShort` with the satisfiability certificate `keygenShort_generable`, so the introduced hypotheses are inhabited.

Purely additive (the old `mldsaMLWE`/`nma_security` chain is untouched, retired when the headline moves over); the STMSIS/extraction/headline stays out for steps 3–4. Full CI set green on 4.32; the three load-bearing decls kernel-clean.

Remaining: (3) the explicit tailored→standard-STMSIS normal-form reduction, (4) the CMA-to-NMA engine, then compose the headline on the landed pieces and narrow #467 accordingly.

alik-eth:
Step 4a is up as #507: the generic Fiat-Shamir-with-aborts CMA-to-NMA reduction engine, as its own reviewable unit (no ML-DSA content — pure VCVio machinery, verified to import nothing from LatticeCrypto/Extern).

It brings `euf_cma_to_nma` and its supporting development (the `WithAbort/Security/` phase modules + `GhostBodies/` machinery) onto main, replacing the current 151-line stub — and in doing so **removes two `sorry`-backed placeholder theorems** (`euf_cma_bound`/`_perfectHVZK`, confirmed zero consumers repo-wide) in favour of the proven reduction. `cmaToNmaLoss`/`perKeyLoss` are preserved via the umbrella. `euf_cma_to_nma` kernel-clean; full CI set + test libs green on 4.32.

That leaves: composing the ML-DSA headline (`nma_security_short`/`_fips` on the landed legs #504/#506 + this engine) and the tailored→standard SelfTargetMSIS normal-form reduction (the one genuinely-new piece). The decomposition stack is now #503 (binding) / #504 (short-MLWE) / #506 (STMSIS leg) / #507 (engine), with #479 as the Laws/Extern base.

alik-eth:
The composition capstone is up as #508: the full short-model EUF-NMA/CMA headline assembled on the landed pieces — `nma_security_short` (exact key-swap + reprogramming + STMSIS extraction), `nma_security_fips` (FIPS keygen via `expandSReplacement`), `euf_cma_security_of_nma_short` (CMA composition on the #507 engine's `euf_cma_to_nma`), the matrix corollary, and the joint frontier witness `trivial_euf_cma_security_of_nma_short` (every hypothesis inhabited at abstract params — no `native_decide`). All kernel-clean `[propext, Classical.choice, Quot.sound]`, additive, full CI set green on 4.32.

So the decomposition of the old 11.7k-line #467 is now a reviewable stack: #503 (binding) → #504 (short-MLWE) → #506 (STMSIS leg) → #507 (engine, independent) → #508 (headline), with #479 as the narrowed Laws/Extern base. #467 itself can be closed/narrowed once these land. The one genuinely-new piece — the tailored→standard SelfTargetMSIS normal-form reduction — is in progress separately.

dtumad (commented):
Following up on my 2026-07-27 comment now that I have audited the replacement stack. The security content here is superseded: #506 carries the corrected tailored SelfTargetMSIS relation, #507 the FS-with-aborts CMA-to-NMA engine, and #508 the composed short-model headline. The branch is also conflicting. I do not think it should be reviewed as an umbrella, but two pieces are unique to it and should not be closed with it.

**1. The legacy `MLDSA.euf_cma_security` cleanup.** This branch removes that `sorry` — `LatticeCrypto/MLDSA/Security.lean` is clean here, while it still has the hole at `:366` on `main` and on #508's head. #508 is additive and does not touch that file, so on the current plan the whole ML-DSA stack can land without the axiom baseline shrinking by one declaration. Extracting just the deletion/redirect as a small PR after the new headline lands is what makes the closure real, and it is small enough to review on its own.

**2. The concrete deterministic-laws and non-vacuity work** in `LatticeCrypto/MLDSA/Concrete/Laws.lean` and `Concrete/NonVacuity.lean`. Worth keeping separate from the security headline, since it is answering a different question — that the concrete primitives satisfy the laws the abstract layer assumes — and it is the kind of thing that gets buried in a large security PR.

For that second piece, `NonVacuity.lean:31-33` is already explicit that the `native_decide` certificate for the concrete 256×256 NTT matrix inversion is inherited rather than introduced. That is the right disclosure and I would keep it prominent in the extracted PR, so the trust exception stays visible in the axiom census rather than becoming a property of "the ML-DSA concrete layer" generally.

No need to keep this open as a review target in the meantime — the two extractions above are the whole remaining value as far as I can tell.

alik-eth:
Closing per @dtumad's audit — the security content here is superseded by #506 (merged), #507 and #508, and the branch is conflicting on a July base. Closing it as a review target, not discarding the two pieces he identified as unique to it.

The head is tagged **`archive/pr467-mldsa228`** (`2060596a`) so branch deletion cannot lose it.

## Still owed, and where it stands

**1. The legacy `MLDSA.euf_cma_security` cleanup — ready once #508 lands.**

This branch removes the `sorry` that still sits at `LatticeCrypto/MLDSA/Security.lean:366` on `main` and on #508's head. Since #508 is additive and does not touch that file, the ML-DSA stack could otherwise land with that hole intact and the axiom baseline never shrinking by the one declaration. Worth doing as its own small PR after the new headline merges, exactly as suggested.

**2. The concrete deterministic-laws and non-vacuity work — blocked on a rule that postdates this branch.**

`Concrete/Laws.lean` and `Concrete/NonVacuity.lean` cannot be lifted across as-is. The dependency chain is forced:

```
NonVacuity → Laws → Instance → Sampling → import FFI.Hashing
```

`Instance.concretePrimitives` is built directly from `expandSeed` / `expandA` / `expandS` / `expandMask` / `sampleInBall`, all of which live in `Sampling.lean`, which pulls in SHAKE. `FFI.Hashing` is today `Extern/Hashing.lean`, and `LatticeCrypto` is a proof library that may never import `Extern` — enforced by `scripts/check-extern-isolation.sh`. I checked: no `LatticeCrypto` file on current `main` violates that rule, so extracting this unchanged would be the first breach and CI would reject it. This branch is based on `cbd4144b` from July, before both the `FFI` → `Extern` rename and the isolation rule.

Three ways forward, and the choice is a design call rather than a patch:

1. **Abstract the hash** — parameterise the concrete sampler over a SHAKE function and supply `Extern`'s at the call site. Keeps "the concrete primitives satisfy the abstract laws" inside the proof library, at the cost of a signature change. Same shape as `FloatLike`, which the repo already uses to keep `FPR`/`Float` out of the proof layer.
2. **Rehome** `Instance` / `Laws` / `NonVacuity` to where importing `Extern` is legal. Cheapest, but moves proof content into a library downstream consumers cannot import.
3. Extract only the parts not touching `Sampling` — likely vacuous, since most of these laws are statements *about* those primitives.

@dtumad — one question that decides between (1) and (2). The laws here are output-range and injectivity properties: `concrete_expandMask_bound`, `concrete_sampleInBall_norm`, `concrete_w1Encode_injOn`, and the rounding family. Those look like they hold for *any* byte stream rather than needing SHAKE-specific properties, which would make (1) preserve them unchanged. But I have not confirmed that per-law, and if any of them do lean on something SHAKE-specific then abstracting changes what is being proved and (2) becomes the better answer. Your call on which, and I am happy to do the work once the destination is settled.

Also worth carrying into whichever PR takes this: `NonVacuity.lean:31-33` is explicit that the `native_decide` certificate for the concrete 256×256 NTT matrix inversion is **inherited** via `concrete_transform` rather than introduced, and that the abstract `Laws`-gated statements remain `[propext, Classical.choice, Quot.sound]`. That disclosure should stay prominent so the trust exception remains visible in the axiom census rather than becoming a property of "the ML-DSA concrete layer" generally.

Issue #468, opened 2026-07-05
`FPRBridge.toReal` routes through opaque `Float.ofBits`, blocking the `HasRealSemantics FPR` instance — proposal to decode the `UInt64` bits directly


aryaethn:
Hi @quangvdao — you already flag this in the block comment above the commented-out `HasRealSemantics FPR ieee754_machineEpsilon` instance in `LatticeCrypto/Falcon/Concrete/ApproxArith.lean` ("**Not provable as stated.**", with the `Float.ofBits` / `Float.toRat0` opacity reasoning and the two suggested remedies). This issue isn't reporting that as news — it's to (a) pin down *exactly* which obligations are blocked vs. which aren't, and (b) propose a concrete, low-blast-radius fix along your remedy #2, with a small verified proof-of-concept, so you can green-light the design direction before I do the work.

Relates to #229 (`add_error`), #230 (`mul_error`), #231 (`div_error`), #232 (`sqrt_error`), #233 (`expm_p63_error`).

### Diagnosis (what I re-verified)

`FPRBridge.toReal x := ((Float.ofBits x).toRat0 : ℝ)`. The whole path bottoms out in `@[extern] opaque` core primitives with no kernel reduction and no formal spec: `Float.toRat0 → Float.toRat? → Float.toRatParts`, and `toRatParts` calls `Float.isFinite` and `Float.frExp`, both opaque. Core's `floatSpec` is literally `opaque … := { float := Unit, … }`. The Lean reference manual is explicit that this is by design: floats "are not encoded in Lean's logic … represented by an opaque type … the kernel is not capable of computing with or reasoning about floating-point values without additional axioms … equality of floating-point numbers is not decidable." Batteries' own `toRatParts'` carries a `TODO … only relevant if Float is not opaque`. Nothing in Mathlib, Batteries, or this repo bridges the gap.

Splitting the instance's obligations by what actually happens:

[...]

Issue #229, opened 2026-04-01
Proof obligation for `add_error` in `LatticeCrypto/Falcon/Concrete/FPRBridge.lean`


quangvdao:
A proof in `LatticeCrypto/Falcon/Concrete/FPRBridge.lean` contains a `sorry`.

**Goal:** Replace the `sorry` with a complete proof.

[Link to the sorry on GitHub](https://github.com/Verified-zkEVM/VCV-io/blob/master/LatticeCrypto/Falcon/Concrete/FPRBridge.lean#L75)

**Code Snippet:**
```lean
  nttOps := hn ▸ Falcon.Concrete.concreteNTTRingOps p.logn

/-! ## Per-operation error bounds -/

/-- Relative error bound for `FPR.add`. -/
theorem add_error (a b : FPR) :
```

quangvdao:
Proved in #514 (bit-level decode of `FPR` to ℝ plus all six error bounds), with the expm bound finished by kernel-checked Chebyshev certificates in #776. None of these declarations is in the `sorry` baseline (`scripts/axiom_baseline.json`).

`add_error` is at `Extern/Falcon/FPR/Add.lean:1083`. The proved statement adds domain hypotheses: `FPR.IsNormalOrZero` operands, and for the binary operations an exact result in `FPR.InNormalMagnitudeRange`. The unrestricted statement in this issue is false (overflow to the decoder's `0`, subnormal differences); `Extern/Falcon/FPRBridge.lean` records the counterexamples and `Extern.Falcon.FPR.Witnesses` shows the restricted statements are not vacuous.

PR #471, opened 2026-07-08
Falcon faithfulness & soundness remediation — concrete keygen, verify_sign_correct, verify bridge

alexanderlhicks:
**Draft — opened for visibility/coordination, not for merge yet.** cc @alik-eth (#466). This is the complementary half of the Falcon work: where #466 proves EUF-CMA via the abstract GPV → NTRU-collision route and **deletes** the sorry-backed correctness cluster (`sign` / `keyGenFromSeed` / `verify_sign_correct`), this branch **builds out and proves that concrete correctness layer** and audits byte-level faithfulness. The two are largely orthogonal and should inform each other — flagging early so we don't diverge or duplicate.

Canonical working doc: [`docs/agents/falcon-review.md`](docs/agents/falcon-review.md) (session log + findings). Paired re-runnable adversarial-review harness: `scripts/falcon_review.mjs`.

## Scope & baselines

Three baselines judged separately (no single authoritative spec): **Falcon v1.2**, **FN-DSA / FIPS 206**, **Falcon+ / c-fn-dsa**. The repo targets **c-fn-dsa / FN-DSA** (FFI uses `FNDSA_HASH_ID_RAW`); divergences from v1.2 (pk-bound hash, fresh salt per retry, L2-only) are intended and documented (matrix M1–M9 in the doc).

## Proven (standard-axioms-only: `propext, Classical.choice, Quot.sound`)

- **`verify_sign_correct`** (`Security.lean`) — a non-aborting signature always verifies, conditional only on the abstract `h_laws.compress_decompress` law (not the concrete codec).
- **`toRq_rqToIntPolyCentered = id`** (F8, `Scheme.lean`).
- **Verify bridge `concrete_verify_eq_verify`** (`Concrete/FPRBridge.lean`) — the byte-level verifier provably equals the spec verifier with **no semantic hypotheses** (B6 U32-kernel equalities + both codec round-trips `sigDecode`/`pkDecode` discharged inline; remaining side-conditions are purely structural/numeric).
- **`sign`** implemented as a fuel-bounded `Option` rejection loop (`Scheme.lean`).
- **B9/ENC-3 fix**: `decompress` rejects non-canonical-length signatures (over-length malleability), the one genuine wire bug found.
- **KG-quickwin**: `NTRUSolver` now uses the real `Concrete/FXR.lean` + `PolyBigInt.lean` (deleted 9 stub `sorry`s); `check_ortho_norm` is sorry-free and executable.

## Open (WIP — why this is a draft)

- **`keyGenFromSeed`** (`Scheme.lean`) — `sorry`.
- **NTRU solver ascent**: `solve_NTRU_intermediate` / `solve_NTRU_depth0` — `sorry`. A full port of c-fn-dsa `solve_NTRU_depth0` was written but **fails end-to-end validation** (reverted to `sorry`; implementation + diagnostics preserved in [`docs/agents/falcon-ntru-depth0-wip.lean`](docs/agents/falcon-ntru-depth0-wip.lean)). Prime suspect: `solve_NTRU_deepest`'s deeper-`(F,G)` convention, never runtime-validated. Next step: differential-trace against the c-fn-dsa backend.
- **`euf_cma_security`** + the generic GPV chain (`VCVio/CryptoFoundations/GPVHashAndSign.lean` ×4) — **this is exactly what #466 addresses**; coordination point.

**Sorry count: 14** (9 Falcon + 4 GPV + 1 ToMathlib/RenyiDivergence). Full `LatticeCrypto` build green at v4.30.0. `#print axioms` confirmed standard-only on every proven result above.


alexanderlhicks:
@alik-eth I'd started doing some work on Falcon after someone asked me about it but it didn't seem necessary in the end (cc https://github.com/kobigurk ?) so I left it unfinished; since you have an active PR looking at this, feel free to re-use anything valuable here (add me as co-author to any PR I guess). 
I primarily followed https://falcon-sign.info/ rather than the NIST submission, assuming this would be a more up to date reference, with an intention to experiment with FFI into the reference C code but didn't reach that stage.

alik-eth:
@alexanderlhicks This is exactly the missing half — nice catch on ENC-3, that's a genuine wire bug. Mapped the interaction with #466 precisely:

**Conflict surface is just two files** — `Falcon/Scheme.lean` and `Falcon/Security.lean` — and it's decl-level clean: #466 *removes* the sorry-backed `sign`/`keyGenFromSeed`/`verify_sign_correct` and rewrites `Security.lean` around the EUF-CMA chain; your branch re-adds those same decls properly proven. Everything else here (`Concrete/*`, `Primitives`, FFI, the review harness) is disjoint from #466.

**Proposed sequencing:** #466 lands first (it's deep in review and unblocks the sorry-free GPV chain — 4 of your 14 sorries disappear on rebase), then this rebases on top and your correctness cluster slots back in additively. One suggestion for the rebase: consider landing `verify_sign_correct` + the fuel-bounded `sign` in their own module (e.g. `Falcon/Correctness.lean`, or next to your verify bridge in `Concrete/`) so `Security.lean` stays security-only — keeps future churn between our branches near zero. Happy to help with that rebase; I know both trees well.

**Synergy after both land:** your correctness layer is the natural discharge route for part of #466's hypothesis surface (`hCorrect`/`hNeverFail` on honest keys, eventually `hTransport` via `SamplerQuality`/`renyiDiv`). Also: does your local work cover the `ntruPSFCollisionProblem → ntruSISProblem` kernel-vector bridge? It's small and self-contained; if yours doesn't, I'll pick it up after your rebase so we don't cross in `Security.lean`.

alexanderlhicks:
> Proposed sequencing: https://github.com/Verified-zkEVM/VCV-io/pull/466 lands first

Agreed. I'll aim to rebase asap as I'll be bus with other priorities later this month and want to ensure my partial work is useful rather than a blocker. 

> One suggestion for the rebase: consider landing verify_sign_correct + the fuel-bounded sign in their own module (e.g. Falcon/Correctness.lean, or next to your verify bridge in Concrete/) so Security.lean stays security-only 

Yep, can do.

> ntruPSFCollisionProblem → ntruSISProblem

I only have the definition, stated as 
```

/-- The NTRU-SIS problem: given `h ∈ R_q` (the Falcon public key), find short
`(s₁, s₂) ∈ R_q²` satisfying `s₁ + s₂ · h = 0 mod q` with
`‖(s₁, s₂)‖₂² ≤ ⌊β²⌋`.

This is the lattice problem underlying Falcon's security. It is an instance of
the generic SIS problem where the matrix is the single-row matrix `[I | h]`
over the cyclotomic ring `R_q = ℤ_q[x]/(x^n + 1)`. -/
noncomputable def ntruSISProblem [SampleableType (Rq p.n)] :
    SIS.Problem (Rq p.n) (Rq p.n × Rq p.n) where
  sampleChallenge := $ᵗ (Rq p.n)
  isValid h x :=
    decide (x ≠ (0, 0)) &&
    decide (pairL2NormSq x.1 x.2 ≤ p.betaSquared) &&
    decide (x.1 + negacyclicMul x.2 h = 0)
```

I may have a bit of time to look at it further but if you don't see anything by sometime next week you can assume I've been prioritising other work and feel free to pick it up from there.  (edit: started on this)

alexanderlhicks:
Rebased this onto #466. 

This adds
  - Correctness split: verify_sign_correct moved into LatticeCrypto/Falcon/Correctness.lean, so Security.lean stays security-only (per @dtumad's file-organization steer / your suggestion).
  - collision → keyed-SIS bridge: shipped the two follow-ups from the design note above —
    - ntruPSFCollision_to_keyedSIS_valid: a valid Falcon-PSF collision differences to a valid keyed NTRU-SIS solution at bound 4·betaSquared (kernel by negacyclicMul_sub_left; norm by new centered-residue triangle lemmas in Ring/Norms.lean — the valMinAbs min-representative inequality, so no no-wraparound hypothesis is needed);
    - ntruPSFCollision_advantage_le_keyedSIS: the reduction adversary wins with ≥ the collision advantage;
    - euf_cma_security_ntruSIS: your euf_cma_security, composed with the bridge, so the bound now terminates at ntruKeyedSISProblem (4·betaSquared) — the standard [I | h] kernel-vector target.
    - The SIS problem is parameterized (ntruSISProblemWithSample by sampler + bound); the idealized uniform-h/betaSquared ntruSISProblem is kept separate. The uniform-h step (NTRU decisional) is the remaining piece, deliberately not assumed here.

The sorry count is 8 (down from 14): FPRBridge ×5 (FPR error bounds behind the disabled HasRealSemantics), NTRUSolver ×2 (the NTRU-solver ascent), and the ToMathlib Rényi sorry — your GPV chain resolved the other six. 

Credited you as co-author on the bridge commit since it builds directly on your euf_cma_security / ntruPSFCollisionProblem.

Review is encouraged as always of course, but I think this is a good handover point otherwise.

aryaethn:
Coordination note on the `Concrete/FPRBridge.lean` real-semantics layer — this touches the same file, so flagging early per your "don't diverge or duplicate."

Your `concrete_verify_eq_verify` bridge discharges the *verify* path with no semantic hypotheses, which is great — and it deliberately stops short of the **real-semantics arithmetic layer** in that same file: `toReal` and the `add`/`mul`/`div`/`sqrt`/`expm_p63_error` bounds (#229–233), plus the commented-out `HasRealSemantics FPR` instance in `ApproxArith.lean`. That's what #468 is about.

The blocker there is that `def toReal x := (Float.ofBits x).toRat0` routes through opaque `@[extern]` `Float`, invisible to the kernel — which the `ApproxArith.lean` instance docstring already documents ("**Not provable as stated.**"), proposing two fixes. **Option 2 — "a verified pure-Lean IEEE-754 decoder that replaces the opaque `Float` path" — is exactly the `toRatBits` decoder I proposed in #468**, and I have it working: decoding the `UInt64` bits directly (sign / biased-exp / mantissa → ℚ → ℝ) makes `interp_zero`, `interp_one`, and `neg_exact` **kernel-provable with no extra axioms** (`#print axioms` = `[propext, Classical.choice, Quot.sound]`); `sub_error` then follows from `add_error` + `neg_exact`. The four core rounding-error bounds stay genuine work (a formalization of Pornin's integer-only emulation), but become *well-posed* instead of inaccessible.

Blast radius is a drop-in swap of the `def toReal` body — the five theorem *statements* are textually unchanged, and `toReal` is `noncomputable` so nothing evaluates it. Since you're live in this file, I'd sequence a small PR (decoder swap + `interp_zero`/`interp_one`/`neg_exact` + uncomment the `HasRealSemantics FPR` instance, leaving the four error bounds as tracked #229–232 sorries) behind or alongside #471 — whatever avoids a collision. Does that split work for you? cc @quangvdao.

---
*Written by aryaethn's Claude Code Agent (Approved by aryaethn)*

alik-eth:
@alexanderlhicks Heads-up for your planned rebase: #466's Falcon surface moved substantially this round, all additive to `Security.lean`'s security-only scope (which should keep your correctness cluster conflict-free as discussed):

- `SISBridge.lean` — the collision→NTRU-SIS kernel-vector bridge (`euf_cma_security_ntruSIS`).
- `samplerTransport` + `euf_cma_security_of_samplerTransport(_queryBound)` — the finite-precision transport hypothesis decomposed into a per-call sampler bound with a proven `qSign · ε` accumulation; the `_queryBound` variant also drops `ForgesQueriedPoint` via the append-query compiler that landed in #478.
- `idealSamplerGuessBound` + `euf_cma_collision_security` — the exact-match term discharged under a named min-entropy assumption (1-D discrete-Gaussian mass bounds backing it are #485, off main).

`Scheme.lean` is untouched beyond what you saw at review time. If any of this crosses wires with the faithfulness work, happy to coordinate sequencing again.

dtumad:
I took another pass over this in the context of current `main`, #466/#478, the Falcon v1.2 algorithms, and the current c-fn-dsa tree. The branch contains several useful pieces, but I think its own “visibility/coordination, not for merge yet” description is still the right disposition.

A local merge with current `main` was mechanically clean, but the resulting review surface is still about 15k added lines and includes an older copy of the generic GPV work that has since diverged. The remaining keygen/NTRU holes are also substantive rather than merge-conflict cleanup. I would therefore avoid trying to remediate or merge #471 as one unit.

The clearest reusable slice was the exact-length signature decoder change. The old path accepted an appended byte because `sigDecode` retained all trailing bytes while `decompress` consumed only `dlen`; both Falcon v1.2 Algorithm 18 and c-fn-dsa’s fixed-size verification path support exact framing. I extracted that as #501, kept Alexander as the commit author, added a rejection theorem and regression check, and adapted it to the current `Extern` layout.

A few other parts look promising as later, smaller slices:

- The abstract `verify_sign_correct` argument looks coherent for the theorem it actually states: successful support membership gives a sampled pair that passed the norm and compression gates; the constructed `s₁ := c - s₂·h` gives exact evaluation; and the codec law recovers `s₂`. I would keep its present qualification prominent: this is conditional abstract correctness, not validation of key generation or the sampler distribution.
- The collision-to-keyed-NTRU-SIS map has the expected algebra and `4·β²` bound: subtracting equal PSF images gives a kernel vector, while the squared triangle bound gives the factor four. It seems more natural to restack that only after the generic GPV work in #478/#466 has settled.
- The codec round trips and concrete-kernel equalities are potentially valuable, but would be easier to trust and maintain as focused PRs against the current `Extern` files and a pinned reference implementation revision.

One baseline detail is now especially worth making explicit. NIST still describes FIPS 206 as in development, and c-fn-dsa’s README was updated on 2026-07-22 to say that no FN-DSA draft exists yet and that the code is the author’s current best guess. That update also changed wire and hashing details after this PR’s 2026-07-10 head. So I would describe this branch as targeting a pinned c-fn-dsa/Falcon+ snapshot, or re-run the faithfulness matrix against current c-fn-dsa, rather than treating “FN-DSA faithful” as a stable claim yet.

A low-friction extraction order that seems to preserve the work well would be: #501 first; then the abstract sign/correctness slice; then codec/kernel bridges after pinning the implementation snapshot; then the collision→SIS bridge after GPV; and keygen/NTRU ascent last, once its differential validation is available. This is meant as a suggested way to make the contributions reviewable, not a request to restructure the WIP branch itself.

dtumad (commented):
I went through this while auditing the open Falcon work. There is real content here that should not get lost, but I do not think it lands as one branch — it currently conflicts with the active Falcon stack (#466/#478/#514), and its own body documents a WIP port. My suggestion is to split it into the clusters below and land them independently.

Worth keeping, roughly in the order I would extract them:

1. **Abstract signing and correctness.** The repaired leaf sampling, the reconstruction `s₁ := c - s₂·h`, the fuel-bounded rejection loop, and the conditional `verify_sign_correct` in `Correctness.lean`. This is the most self-contained piece and the one with the clearest statement.
2. **Codec/kernel bridge.** The concrete roundtrips, the integer-kernel correspondence and `concrete_verify_eq_verify`. The overlong-signature acceptance fix from this line of work has already landed separately, which is a good sign for the rest of it.
3. **NTRU implementation plumbing.** The low-level `FXR`/`PolyBigInt` implementations plus the differential traces that go with them.
4. **The negative result.** The documented failed `depth0` port — the diagnostic findings pointer and "kept `sorry` until correct" at `NTRUSolver.lean:386-390` — is exactly the kind of evidence worth preserving rather than quietly dropping. A differential validation that failed tells the next person more than an untested port would.

Two scope points to keep straight while splitting.

`verify_sign_correct` still discharges through `Primitives.Laws.compress_decompress`, and there is no concrete aggregate laws instance, so this is not yet end-to-end concrete Falcon correctness — it is correctness relative to assumed primitive laws. The byte-level verifier bridge and the proof-level GPV security theorem also remain two separate worlds; neither implies the other yet.

The two ascent holes (`solve_NTRU_depth0`, and the sibling at `NTRUSolver.lean:366`) are `sorry` at the definition level rather than at the proof level, so anything downstream of them is not merely unproved but uncomputable. I would leave them as explicitly named obligations rather than reintroducing unvalidated code to fill the shape.

One thing that has moved since this branch was opened: `LatticeCrypto/Falcon/Concrete/FPRBridge.lean` here still has its `interp_zero`/`interp_one`/`neg_exact` fields as `sorry`, with `ApproxArith.lean:241` recording them as "not provable as stated". #514 proves that family of bounds against a pure bit-field decoder, so that cluster is superseded rather than pending — worth dropping from the extraction list to avoid re-litigating it.

dtumad:
Deep semantic audit across the open PR set (semantics only). Extends my 08-17 comment (4-way split request, `FPRBridge` superseded by #514); no pushes since 07-10. **Verdict: do not merge as one branch — one new blocker below is fatal to the security cluster as bundled, and the remediation path is the split.**

**BLOCKING (new): `SamplerQuality` + `HasUniformSamplerLoss` are jointly satisfiable only at `samplerLoss = ⊤`.** At head, `SamplerQuality` bundles

- `hOrder : 1 < renyiOrder`,
- `quality : ∀ c, renyiDiv renyiOrder ((falconPSF p prims).trapdoorSample pk sk c) (idealSampler c) ≤ bound`,
- `idealCorrect : ∀ c, ∀ x ∈ support (idealSampler c), … ∧ isShort x = true`.

`idealCorrect` puts the ideal support inside `isShort`; the concrete one-shot sampler has positive mass outside it (that is exactly why this PR's own `sign` is fuel-bounded with a retry). For any such point `x` with `p x > 0`, `q x = 0`, and `a > 1`, the summand of `renyiMGF a p q = ∑' x, p x ^ a * q x ^ (1−a)` is `p x ^ a * 0 ^ (1−a) = ⊤` in `ℝ≥0∞` (negative exponent at zero), so `renyiDiv = ⊤` and `quality` forces `bound = ⊤`. `HasUniformSamplerLoss samplerLoss` then holds only at `samplerLoss = ⊤`, where the conclusion `… + samplerLoss` is trivially true. Note the `hOrder` guard is what makes this unavoidable rather than accidental — it correctly closes the `a ≤ 1` junk-value escape, which leaves `⊤` as the only value. The Rényi argument in the literature is stated for the *accepted-output* distribution (post-rejection), which restores absolute continuity; as bundled here it can never be instantiated below `⊤`. Same defect class as #466's `ε_step ≥ p_reject`, but exactly `∞` — and #466 already dropped this Rényi scaffolding on review instruction, so it should not return in this shape.

**BLOCKING (sharpened): `verify_sign_correct` is proved by redefining the scheme so that it is nearly a tautology.** At head, `fromFFTPreimage` computes `s₂ := −toRq (ifftRound v₁FFT)` and then **defines** `s₁ := c − s₂·h` — i.e. `s₁` is what the verifier recomputes, making `eval pk (trapdoorSample …) = c` definitional. The real obligation — that the FP inverse-FFT rounding lands on an actual lattice point, so that the implementation's `s₁ = c − v₀` coincides with `c − s₂·h` — is not proved but *erased from the model*. Consequence: the modelled `isShort` check runs on a different vector than the implementation's whenever FP error is non-negligible, so the modelled acceptance region diverges from the C code's. If the project wants `s₁`-recomputation as a modelling decision it must be argued (against Falcon spec Alg 16 and what `sign_core` actually norm-checks) and documented as such, not silently substituted in a PR titled "faithfulness remediation".

**BLOCKING (mechanical): 10 duplicated framework files.** The `VCVio/CryptoFoundations/GPVHashAndSign/*` files here are a byte-identical snapshot of #478's surface (minus `AppendQuery`/`MinEntropy`/`SamplerTransport`), and `Falcon.euf_cma_security` is textually identical to #466's. Drop them all; this PR's genuinely unique value is the concrete/codec layer.

Path forward per the earlier 4-way split, updated: (1) abstract signing + correctness (after the `s₁` decision above); (2) codec/kernel bridge; (3) **drop** the `SamplerQuality`/`HasUniformSamplerLoss` bundle entirely (unsatisfiable at any finite bound) and drop the `FPRBridge` cluster (superseded by #514); (4) the documented negative `depth0` result as its own PR. Also please move `docs/agents/falcon-review.md`, `scripts/falcon_review.mjs`, and the two `*-wip.lean` session logs out of the library material.

alik-eth:
Proceeding with the split rather than remediating this branch; this comment records what goes where.

**Dropped, not carried into any slice.** The `SamplerQuality`/`HasUniformSamplerLoss` bundle (agreed: `idealCorrect` puts the ideal support inside `isShort`, the one-shot sampler has mass outside it, and `renyiDiv` is `⊤` there, so `HasUniformSamplerLoss` holds only at `⊤`). The ten duplicated `GPVHashAndSign/*` files and the copy of `Falcon.euf_cma_security`. The `FPRBridge` float-bound cluster: this branch never added to it, it only inherits `main`'s July version, and #514 proves those bounds; nothing from it is extracted. `docs/agents/falcon-review.md`, `scripts/falcon_review.mjs`, and the two `*-wip.lean` logs stay out of library material.

**Superseded by other PRs.** The retry-loop signer and the collision-to-NTRU-SIS algebra (`centeredRepr_sub_natAbs_le`, `pairL2NormSq_sub_le_two_mul_add`) now live in #466 (`falconRetrySignatureAlg`, `SISBridge.lean`), so the "abstract signing" slice is not re-extracted from here. #501 already took the exact-length decoder.

**Extracted now: the codec/kernel bridge** — PR `#638`. This branch's own edit to `FPRBridge.lean` was the discharge of `concrete_verify_eq_verify`'s four side conditions, and that is the slice: `sigDecode_sigEncode`, `publicKeyBytes_extract`, `pkDecode_pkEncode` (for `4 ∣ n`), `negacyclicMulU32_eq_negacyclicMul`, `pairL2NormSqU32_eq_pairL2NormSq` (under the `UInt64` no-overflow bound), and the discharged bridge `concrete_verify_eq_verify_of_logn` for `2 ≤ logn ≤ 10`. Ported onto current `main`, module system and all, in a new `Extern/Falcon/VerifyBridge.lean` so it does not touch #514's file. Both sides of each equation are this repository's definitions, so no c-fn-dsa revision is pinned or claimed.

**Remaining, in order.** (1) `verify_sign_correct` as conditional abstract correctness, only after the `s₁` question is settled the way you describe: keep the implementation's `s₁ = c − v₀` and state the lattice-point condition on the rounded inverse FFT as an explicit hypothesis rather than defining it away. (2) The documented negative `depth0` result as its own small PR. (3) Keygen/NTRU ascent last, once its differential validation exists. When (1)–(3) have their own PRs this one will be closed as a review target with an archive tag, the way #467 was.

alik-eth:
Slice (1) is up as #640: `Falcon.sign` defined as the fuel-bounded fresh-salt loop, and `verify_sign_correct` proved with `fromFFTPreimage` untouched and the lattice-point condition on the rounded inverse FFT (`hpreimage`) as an explicit hypothesis, alongside the codec round-trip law. Based on `main`, independent of #638.

quangvdao:
Review-train disposition at unchanged head `0538916379dd4c7f99821cec24b5395796017625`: **hold this legacy aggregate; do not merge it as a unit.** GitHub marks it non-draft, but the description explicitly says it is for coordination rather than merging, and the latest author handoff adopts extraction rather than repairing the old branch.

The relevant review work is therefore on the replacements: #638 for the repository-internal codec/kernel verifier bridge, #640 for conditional signing correctness with the rounded-FFT preimage condition explicit, and the later concrete-arithmetic slices #662/#663. #466/#478 remain separate security/framework work. The exact-length decoder extraction #501 is already merged. This is a routing/status assessment, not an approval of those replacement PRs.

The existing substantive concerns still make a blanket merge inappropriate: the aggregate carries an older duplicated GPV/security surface, unresolved key-generation/NTRU work, and modelling/claim issues already discussed above. In particular, a finite Rényi transport claim requires checking absolute continuity of the actual distributions; positive concrete mass outside the ideal support would invalidate a finite bound. Conditional abstract signing correctness must not be presented as a concrete FFT/sampler faithfulness theorem.

No source changes, close/archive operation, or merge request made here. I inspected the contract and review/author handoff, not a fresh line-by-line audit or rebuild of this July aggregate. The current train leaves this PR open pending maintainer direction about taking over the replacement slices.

PR #475, opened 2026-07-10
feat(CryptoFoundations): add round-by-round knowledge games

eliasjudin:
Formalizes round-indexed event games and the extensional doomed-set, uniform-challenge, direct-extractor, and terminal-rejection clauses of round-by-round knowledge in Lean 4.

Polynomial-time extractor admissibility and asymptotic negligibility remain incomplete in this concrete API.

Co-authored-by: Aristotle (Harmonic) <aristotle-harmonic@harmonic.fun>


VCVio/CryptoFoundations/RoundByRound.lean
```diff
@@ -0,0 +1,285 @@
...
+  sample : (round : Round) → Context round → ProbComp (Result round)
+  /-- The event measured in a context and round. -/
+  event : (round : Round) → Context round → Result round → Prop
+
+namespace GameFamily
+
+variable {Round : Type u} {Context : Round → Type v}
+
+/-- The failure-based experiment that succeeds exactly when the indexed event occurs. -/
+def experiment (games : GameFamily Round Context)
```
alexanderlhicks, line 69: **[AI-generated]** `def … := by classical exact {…}` appears to be the first tactic-block `def` in the repo (the existing idiom for classical instances in definitions is term-level, e.g. `letI := Classical.dec _` in `VCVio/ProgramLogic/Relational/QuantitativeDefs.lean:38`). Term mode keeps the definition's unfolding transparent, which would also let `experiment_advantage` below avoid its `change` steps. Since this file already uses term-level `letI` in `sampleChallenge`, the same style works here:

```lean
def experiment (games : GameFamily Round Context)
    (round : Round) (context : Context round) : SecExp (OptionT ProbComp) :=
  letI : DecidablePred (games.event round context) := Classical.decPred _
  { toSPMFSemantics := SPMFSemantics.ofMonadLift (OptionT ProbComp)
    main := do
      let result ← OptionT.lift (games.sample round context)
      guard (games.event round context result) }
```

Easy to validate: swap and rebuild.

VCVio/CryptoFoundations/RoundByRound.lean
```diff
@@ -0,0 +1,285 @@
...
+  classical
+  exact
+    { toSPMFSemantics := SPMFSemantics.ofMonadLift (OptionT ProbComp)
+      main := do
+        let result ← OptionT.lift (games.sample round context)
+        guard (games.event round context result) }
+
+/-- The advantage of the indexed experiment is the probability of its event. -/
+@[simp]
+theorem experiment_advantage (games : GameFamily Round Context)
```
alexanderlhicks, line 79: **[AI-generated]** The proof is correct (machine-checked), but the two `change` steps tie it to definitional unfolding through `SecExp.advantage` / `ofMonadLift` / the tactic-defined `experiment`, which makes it fragile under refactors. The library already has the transparent route:

- `SPMFSemantics.ofMonadLift_probFailure` (`VCVio/EvalDist/Defs/Semantics.lean:145`) replaces the first `change`;
- `probOutput_bind_guard_eq_probEvent` (`VCVio/OracleComp/EvalDist.lean:394`) is exactly the "experiment succeeds iff event holds" bridge for `OptionT (OracleComp spec)` — it is currently `private`, so consider publicizing it (or adding a `probFailure_bind_guard_eq_probEvent` sibling next to it, alongside the existing `@[simp]` `probOutput_guard`/`probFailure_guard`) and reducing this proof to a couple of rewrites.

That would also serve the two downstream `experiment_advantage` lemmas, and future SecExp-over-guard constructions get the bridge for free.

VCVio/CryptoFoundations/RoundByRound.lean
```diff
@@ -0,0 +1,285 @@
...
+        Pr[games.event round context | games.sample round context] ≤ error round := by
+  simp only [IsBounded, experiment_advantage]
+
+end GameFamily
+
+/-- Data defining a family of knowledge-transition events.
+
+The context at a round is fixed before `sampleChallenge` draws the fresh challenge. The two witness
+types represent knowledge immediately before and after that challenge. -/
+structure KnowledgeTransitionFamily (Round : Type u) (Context : Round → Type v) where
```
alexanderlhicks, line 113: **[AI-generated]** `KnowledgeTransitionFamily` is not referenced by anything else in the PR: `KnowledgeExtractionFamily.toGameFamily` maps directly to `GameFamily`, and no lemma relates the two specializations. The shape itself is validated by real usage — e.g. ArkLib's proven γ-round bound `gamma_transition_prob_le` (`feat/abf26-plan`, `ArkLib/ProofSystem/ToyProblem/SoundnessBounds.lean` ~L1064) is precisely a `badEvent`-style statement (no-witness pre-state, ∃ post-challenge live message, bound over a uniform challenge) — so the structure is worth having. Suggestion: make that concrete inside this PR by adding the lemma connecting the two layers (e.g. how a `KnowledgeExtractionFamily` induces a `KnowledgeTransitionFamily` whose bounded bad-transition probability corresponds to the extraction clause — the analogue of ArkLib's one-shot ↔ knowledge-state-function bridge), or defer this structure to the follow-up PR that uses it.

alexanderlhicks (commented):
> [!NOTE]
> **This review is AI-generated** (Claude Code), run and posted at my request. Every claim below was machine-validated against the PR head `d26325af` (local build, `#print axioms`, direct reading of the cited sources); line references are exact so each point can be independently checked.

Thanks for this — the formalization checks out on the correctness axis, and the design choice at its core (per-fixed-context, worst-case quantification with a uniform fresh challenge) is the right "paper-strength" layer. Summary of what was validated, followed by suggestions. Code-level points are inline.

## Validated ✅

- **Build**: `lake build VCVio.CryptoFoundations.RoundByRound` at the PR head is clean (no errors/warnings, no `sorry`). Note that CI on this PR only ran the summarize job — worth triggering the Lean build workflow before merge.
- **Axioms**: `#print axioms` on all public declarations reports standard axioms only (`propext`, `Classical.choice`, `Quot.sound`).
- **Faithfulness to BGTZ Definition 4.2** ([ePrint 2023/1256](https://eprint.iacr.org/2023/1256)), clause by clause: `InitialCondition` ↔ "for **all** inputs x (not necessarily in L_R), (x) ∈ D"; `TerminalCondition` ↔ doomed terminal transcript ⇒ reject any final message; `ExtractionCondition` ↔ "Pr_{c←$C_i}[(x,τ,m,c) ∉ D] > ε_i ⇒ Ext(x,τ,m) outputs a valid witness", with the strict inequality direction and the per-fixed-message trigger both matching (the per-message reading is also the one the paper's own §6 proofs use).
- **Uniformity and failure mass**: `SampleableType` carries the equal-probability law (`probOutput_selectElem_eq`) and `probFailure_uniformSample = 0`, so `Pr[escapeEvent … | sampleChallenge …]` is exactly the paper's uniform-challenge probability. `experiment_advantage` is also correct in general: `guard` routes both "event false" and any failure mass of `sample` into `⊥`, so `advantage = Pr[event]` holds unconditionally.
- **Instantiability**: the API admits a complete instantiation — a toy one-round protocol satisfies `ExtensionalConditions` with zero error in a short proof (collapsed below), and the probability obligations are `simp`-tractable.
- One observation, not a defect: the conditions quantify over **all** contexts, including ones unreachable via `extend` from any `initialContext`. This is strictly stronger than Def 4.2 but harmless, since `doomed` can always be restricted to reachable contexts (closed under `extend`); the `IsBounded` docstring's subtype remark already gestures at this.

<details><summary>Instantiation smoke test (compiles against this PR)</summary>

```lean
import VCVio.CryptoFoundations.RoundByRound

open RoundByRound OracleComp

noncomputable def toy : KnowledgeExtractionFamily 1 where
  Statement := Bool
  Witness := Unit
  Context := fun _ => Bool
  Message := fun _ => Unit
  Challenge := fun _ => Bool
  FinalMessage := Unit
  challengeSampleable := fun _ => inferInstance
  statement := fun _ x => x
  initialContext := id
  extend := fun _ x _ _ => x
  statement_initial := fun _ => rfl
  statement_extend := fun _ _ _ _ => rfl
  doomed := fun stage x => stage.val = 0 ∨ x = false
  extract := fun _ _ _ => ()
  relation := fun x _ => x = true
  rejects := fun x _ => x = false

example : toy.ExtensionalConditions (fun _ => 0) := by
  refine ⟨fun input => Or.inl rfl, ?_, ?_⟩
  · rintro x m (h | h)
    · simp [Fin.last] at h
    · exact h
  · intro round x m hdoomed hlt
    by_cases hx : x = true
    · exact hx
    · exfalso
      simp only [Bool.not_eq_true] at hx
      have hzero : Pr[toy.escapeEvent round x m | toy.sampleChallenge round] = 0 := by
        simp [KnowledgeExtractionFamily.escapeEvent, toy, hx]
      rw [hzero] at hlt
      exact lt_irrefl 0 hlt
```
</details>

## Main suggestion: connect the module to its waiting consumers

The quantification style chosen here already has proven consumers on the ArkLib side (ArkLib depends on VCV-io, so this module becomes visible to it). On `Verified-zkEVM/ArkLib` branch `feat/abf26-plan`, the ABF26 toy-problem development proves per-round facts in **exactly this per-fixed-prefix shape** (see the "Quantification note" in `ArkLib/ProofSystem/ToyProblem/Spec/General.lean` ~L1196, and `gamma_transition_prob_le` in `ToyProblem/SoundnessBounds.lean` ~L1064, which is literally a `badEvent`-style bound), then lifts them into ArkLib's averaged `rbrKnowledgeSoundness` game via mixture lemmas (`probEvent_simulateQ_addLift_getChallenge_bind_le`, `ArkLib/ToVCVio/OracleComp/RbrGame.lean:81`); branch `feat/generic-ring-switch` has the analogous `probEvent_challengeRound_le` (`Security/ChallengeRound.lean:49`). Both pipelines consume bare `∀ prefix, Pr[fun c ↦ E … | $ᵗ _] ≤ ε` facts directly.

Two actionable follow-ups that would make this PR immediately load-bearing:

1. **Add the `probEvent`-over-bind mixture lemmas to VCV-io** — `RbrGame.lean` explicitly docstring-flags `probEvent_bind_le_probEvent` (:177), `probEvent_bind_le_probEvent_add` (:199), and `probEvent_bind_le_probEvent_convex` (:233) as "upstream VCV-io candidates". They are the generic bridge from this module's per-context conditions to full averaged game statements, and they have two independent consumers waiting.
2. **Ship one in-repo consumer** for the knowledge layer — most naturally a `SigmaProtocol` adapter (`rounds := 1` instance of `KnowledgeExtractionFamily`), or one of the BGTZ implication theorems (e.g. RBR knowledge ⇒ special soundness; the analogous ArkLib statements exist but are currently sorried in `feat/abf26-plan`'s `Security/Implications.lean`). Either would validate the abstraction against a use case and settle the API.

VCVio/CryptoFoundations/RoundByRound.lean
```diff
@@ -0,0 +1,285 @@
...
+  /-- Statement-and-transcript contexts at every protocol stage. -/
+  Context : Fin (rounds + 1) → Type
+  /-- Prover messages fixed before the verifier samples the challenge at each round. -/
+  Message : Fin rounds → Type
+  /-- Fresh verifier challenges at each round. -/
+  Challenge : Fin rounds → Type
+  /-- Final prover messages supplied after all interaction rounds. -/
+  FinalMessage : Type
+  /-- Canonical uniform-sampling data for each verifier challenge type. -/
+  challengeSampleable : (round : Fin rounds) → SampleableType (Challenge round)
```
alexanderlhicks, line 204: **[AI-generated]** Design note on bundling `SampleableType` as a field: `sampleChallenge` then runs under `letI := games.challengeSampleable round`, which for a user whose challenge types already carry ambient instances (the `[∀ i, SampleableType (pSpec.Challenge i)]` style used elsewhere, e.g. in ArkLib) is defeq to — but not syntactically — the ambient instance. That mismatch can surface as friction in `rw`/`simp` chains against `$ᵗ`-lemmas stated with the ambient instance. It's handled correctly here because the API's own lemmas are stated through `games.sampleChallenge` (good); just flagging that if downstream friction appears, lifting `Challenge` to a structure parameter with instance binders is the standard alternative.

VCVio/CryptoFoundations/RoundByRound.lean
```diff
@@ -0,0 +1,285 @@
...
+  /-- Projects the common statement from a context at any stage. -/
+  statement : (stage : Fin (rounds + 1)) → Context stage → Statement
+  /-- The protocol context containing only the input statement. -/
+  initialContext : Statement → Context 0
+  /-- Appends the prover message and verifier challenge of an interaction round. -/
+  extend : (round : Fin rounds) →
```
alexanderlhicks, line 210: **[AI-generated]** `extend` hard-wires exactly one prover message followed by one challenge per round. Strictly alternating protocols fit directly, but consecutive prover messages or a challenge-first round (both common in practice; ArkLib's `ProtocolSpec` allows arbitrary direction sequences for this reason) only fit via `Unit`-padding of `Message`/`Challenge`. Since BGTZ's IOP model is round-based this is a faithful reading of Def 4.2 — suggestion is just to document it: one sentence in the structure docstring noting the alternation assumption and the `Unit`-padding encoding for other shapes would save future users a detour.

VCVio/CryptoFoundations/RoundByRound.lean
```diff
@@ -0,0 +1,285 @@
...
+/-- Every final prover message is rejected from a doomed terminal context. -/
+def TerminalCondition (games : KnowledgeExtractionFamily rounds) : Prop :=
+  ∀ (context : games.Context (Fin.last rounds)) (message : games.FinalMessage),
+    games.doomed (Fin.last rounds) context → games.rejects context message
+
+/-- The extensional extraction clause with a separate error for each round.
+
+For every fixed context and prover message whose preceding state is doomed, an escape probability
+strictly greater than the round's error requires the directly extracted witness to be valid. -/
+def ExtractionCondition (games : KnowledgeExtractionFamily rounds)
```
alexanderlhicks, line 290: **[AI-generated]** Two notes, one positive and one docstring sharpening. Positive: phrasing ε as a strict trigger threshold (`error round < Pr[escape] → …`) rather than a bound matches Def 4.2 exactly — ε appears in the paper only as the extraction trigger, and the inequality direction is right. Sharpening: since `extract` carries no efficiency requirement yet, `ExtractionCondition` **as a hypothesis** is equivalent to plain witness existence at every triggering context (any inhabitant of `∃ w, relation stmt w` can be repackaged into an `extract` field by choice). The module docstring already says poly-time is omitted; consider stating this consequence explicitly — that until the admissibility layer lands, the condition pins down soundness-strength content and "knowledge" is not yet claimed — so downstream users don't over-read the guarantee. (As *data* the extractor is constructive, which is the right call for the eventual efficient version.)

alexanderlhicks:
**[AI-generated]** Addendum to the review above: the polynomial-time extractor admissibility that this PR (correctly, and explicitly) defers now has a natural in-repo formulation path via #460, which introduces TM-grounded polytime infrastructure — in particular `Computability.EncPolyTime ea eb f`, an encoding-aware polytime witness for **pure** functions (`ToMathlib/Computability/CslibPolyTime.lean` there), plus security-parameter-indexed `PolyTimeAdversary` / `OracleComp.IsPolyTime`.

Since `extract : (round : Fin rounds) → Context round.castSucc → Message round → Witness` is a pure function, the missing BGTZ clause ("Ext is poly-time in |x|") would take the concrete shape: index the family by a security parameter, equip `Context`/`Message`/`Witness` with `Bool`-string encodings (#460's `finEncodingOfFinEnum` covers the finite cases for free), and require an `EncPolyTime` witness for each (uncurried) `extract round`; asymptotic negligibility of `error` then lands in the existing `Asymptotics/Negligible` algebra. Might be worth a `TODO` in the module docstring pointing at that route, so the deferred layer is tracked against #460's landing rather than left open-ended.

alexanderlhicks:
@quangvdao you may have an opinion on this one.

alexanderlhicks:
**Downstream-consumer report from ArkLib.** We have a round-by-round layer in ArkLib that overlaps this PR, and I read the 401 lines here closely to work out how the two relate. Posting the findings since they bear on the API shape. No action implied for this PR unless you want it — mainly I want to record one exact correspondence and one non-correspondence, because the latter is easy to assume and is false.

### The generic transition family exactly packages ArkLib's inner worst-case obligation

`KnowledgeTransitionFamily.IsBounded` is, under the evident translation, the body of ArkLib's `Verifier.rbrKnowledgeSoundnessWorstCase`:

| #475 | ArkLib |
|---|---|
| `Round` | `pSpec.ChallengeIdx` |
| `Context i` | `StmtIn × Transcript i.1.castSucc pSpec` |
| `WitnessBefore i` / `WitnessAfter i` | `WitMid i.1.castSucc` / `WitMid i.1.succ` |
| `extractBefore` | `extractor.extractMid` |
| `preState` / `postState` | the knowledge state function at `i.1.castSucc` / `i.1.succ` |
| `sampleChallenge` | `$ᵗ (pSpec.Challenge i)` |

With that, your

```lean
badEvent = ∃ witnessAfter, ¬ preState (extractBefore … witnessAfter) ∧ postState … witnessAfter
```

is ArkLib's worst-case bad event definitionally, and the boundedness statements differ only by rearranging the context-pair quantifiers. Two caveats on that correspondence, so it isn't overstated: ArkLib's definition additionally existentially packages `WitMid`, the extractor, and the state function, so `IsBounded` corresponds to the inner obligation rather than the whole predicate; and `KnowledgeTransitionFamily` is data-only (seven fields, no axioms) whereas ArkLib's `KnowledgeStateFunction` carries three obligations, so an adapter transfers the bound but none of the structure.

### `ExtractionCondition` is *not* ArkLib's one-shot notion

Worth stating explicitly, because the names invite the assumption and I made it myself before checking. ArkLib's `rbrKnowledgeSoundnessOneShot` quantifies over the prover and input witness, **samples the transcript prefix by running the prover**, records a query log, and **feeds that log to the extractor**. `ExtractionCondition` fixes context and message before sampling the challenge. These are different notions; an equivalence would need a new worst-case one-shot definition on our side plus encoding work. I'd suggest not implying such a correspondence in the docs.

### Scope observation: no averaged, prover-sampled layer

Everything here is worst-case-per-context — `IsBounded` quantifies `∀ round, ∀ context` and takes the probability over the fresh challenge only. Protocol proofs also need the statement about the prover-sampled game. That's a scope boundary rather than a defect, and worth naming in the module docstring so consumers know where the layer stops.

Separately, we have a mixture theorem in ArkLib that crosses exactly that boundary: if for every *fixed* prefix the event over a fresh uniform challenge is `≤ ε`, then the whole game — initial state sampled, adversarial prefix produced under `simulateQ`, challenge drawn — is `≤ ε`. It's generic apart from our challenge oracle, which is itself only `fun q => $ᵗ _`, and `QueryImpl.addLift` is already yours, so it generalises cleanly and would likely be useful in VCVio on its own account.

To be clear about the relationship, since I initially overstated it: **that theorem does not depend on this PR and is not a missing piece of it.** It's a `simulateQ`/`addLift` lemma; your `sample : Context → ProbComp` never mentions oracle simulation, so it composes with this layer only through a translation step, not through its API. I'd rather offer it as an independent PR judged on its own merits than couple the two.

### Smaller notes

- **Law-freeness.** Because the generic transition family has no axioms, the structural obligations (initial doom, transition monotonicity, terminal rejection) live only in `KnowledgeExtractionFamily`, which is alternation-only. A consumer with a *lawful general* state function fits the generic layer but has nowhere to put its laws. Either a lawful variant, or lifting the initial/terminal conditions to the generic layer, would close that.
- **Point consumers at the right layer.** The generic layer is the integration point for arbitrary scheduling; `KnowledgeExtractionFamily`'s strict alternation won't serve a protocol spec with an arbitrary per-round direction without `Unit`-padding gymnastics. One line in the module docstring would save implementors reaching for the source-shaped one first.
- **Citation scope.** The BGTZ Def 3.12 disclaimers (no polynomial-time extractor, no negligible-error asymptotics) are correctly stated in the module header, but not in the per-definition docstrings — which is where readers land via go-to-definition.
- **Namespace check, clear.** The root `RoundByRound` namespace does not collide with ArkLib's `Extractor.RoundByRound`: unqualified references still resolve correctly (verified in Lean), and ArkLib never `open`s it bare.

For context on the downstream side: ArkLib's worst-case definitions and the worst-case⇒averaged bridges are in Verified-zkEVM/ArkLib#668.

*Caveat: I read this PR's source closely but have not built it locally; I'm relying on CI for proof checking.*

alexanderlhicks:
**Follow-up to the note above: we restructured on the ArkLib side, and settled the plan. Nothing here asks this PR to wait.**

We split the ArkLib file I mentioned (Verified-zkEVM/ArkLib#668) along its upstreaming boundary, so the part destined for VCV-io is now isolated and depends on nothing but VCV-io:

- `ArkLib/ToVCVio/OracleComp/QueryTracking/LoggingOracle.lean` — the two `loggingOracle` lemmas, at a path mirroring your `OracleComp/QueryTracking/LoggingOracle.lean`. Upstreaming is a straight file move.
- The `ProtocolSpec`-shaped mixture bounds moved to ArkLib core, where they'll stay until generalised (challenge oracle → arbitrary uniform-answer `QueryImpl`; `QueryImpl.addLift` is already yours).

### Plan, and what it means for this PR

We considered building an adapter from ArkLib's round-by-round definitions onto `KnowledgeTransitionFamily` and **decided against it**, at least for now. Since that family is law-free data, an adapter would transfer the probability bound but none of ArkLib's structural obligations, and the theorems it unlocks are `simp`-level restatements — not worth a dependency from ArkLib's core security definitions onto an in-flight upstream API. So there's no pending ArkLib requirement on this PR's API, and no reason to hold it for us.

We'll upstream the generic operational machinery separately, judged on its own merits. I checked for collisions: no overlap with this PR's content, and its only other touched file (`VCVio/OracleComp/EvalDist.lean` — generalising `probOutput_punit_eq_sub_probFailure`, de-privatising `probOutput_bind_guard_eq_probEvent`) doesn't intersect what we'd add either. Order of landing doesn't matter.

### The two suggestions from the earlier note still stand

Both are about this PR's own shape rather than our needs:

1. **Law-freeness of the generic layer.** The structural obligations live only in `KnowledgeExtractionFamily`, which is alternation-only, so a consumer with a lawful *general* state function fits the generic layer but has nowhere to put its laws. A lawful variant, or lifting the initial/terminal conditions to the generic layer, would close that.
2. **Is the source-shaped layer worth carrying here?** `KnowledgeExtractionFamily` is the least reusable part — strictly alternating, `Unit`-padded for other schedules, one-shot `extract`, doomed-context subtype — while the generic `GameFamily`/`KnowledgeTransitionFamily` kernel is where the reuse is. Keeping both is fine if the BGTZ instantiation stays clearly separable from the kernel; splitting it out would let the kernel merge on its own merits. Maintainer's call, not a blocker.

Plus the two documentation points: the BGTZ Def 3.12 scope disclaimers are in the module header but not the per-definition docstrings (where readers land via go-to-definition), and one line saying the generic layer is the integration point for non-alternating schedules would save implementors reaching for the source-shaped one first.

alexanderlhicks:
**Validation pass: built locally, axiom-probed, checked clause-by-clause against BGTZ. Verdict is merge, with one docstring change I'd suggest first.** Correcting my own earlier caveat — the note above this said I was relying on CI; I'm not any more.

- `lake build` clean (2988 jobs) on `a752d1362`. Zero `sorry` in either added file.
- `#print axioms` on all 41 added declarations: 37 on exactly `[propext, Classical.choice, Quot.sound]`, 4 (`GameFamily`, `KnowledgeTransitionFamily`, `.badEvent`, `.toGameFamily`) on none. No `sorryAx`, no custom axiom.
- `EvalDist.lean` checked for downstream breakage: `probOutput_eq_sub_probFailure_of_unit` keeps its exact signature, its four `SecExp.lean` call sites are untouched, and there was no pre-existing general form to duplicate. Nice touch pinning `(spec := spec)` in the test — that's a real guard against the wrapper losing its implicit-argument shape.

### One deviation from Def 3.12 that isn't in the disclosure list

The three disclosed deviations (extensional only, `Unit`-padded alternation, totalized `extract`) are the only ones I found in the *stated* list — but there's a fourth in the extraction clause. Literal Def 3.12 puts the message quantifier **inside** the hypothesis:

> If `(x,τ) ∈ PartTr(i−1)`, `(x,τ) ∈ D`, and **for all `m ∈ M_i`** we have `Pr_c[(x,τ,m,c) ∉ D] > ε_i`, then `Ext(x,τ,m)` outputs a witness…

so `(∀m. P m) → (∀m. Q m)`. `ExtractionCondition` pulls it outside — `∀m. (P m → Q m)`, a per-message obligation, which is **strictly stronger**.

I think your reading is the right one, and the paper agrees with you twice over: the clause goes on to name `(x,τ,m)` — a specific message — an "RBR extractable partial transcript", which would be vacuous under the ∀m-in-hypothesis reading; and the Thm 1.2 proof uses per-`m` outright, concluding that `(x, τ, EndNode(τ))` is extractable from *that one message's* escape probability. The `∀m` looks transplanted from Def 3.11 clause 3, where it genuinely belongs to the conclusion.

So: not a defect, and it errs in the safe direction. But it means "gives the extensional clauses of Definition 3.12" is slightly overstated, and a reader comparing against the paper will hit this. **Suggest one sentence on `ExtractionCondition`'s docstring** recording that the quantifier is read per-message in line with the source's own Thm 1.2 proof, and that this is stronger than the literal text.

(Also: the polytime/negligibility caveat is now on `ExtensionalConditions` — thanks. It's still absent from `ExtractionCondition` itself, which is the other place a reader lands.)

### On the one-round test — it's doing more than "smoke test" suggests

I'd missed this file existed when I planned the review, and it covers most of what I meant to check. Three things it doesn't reach, which I've compiled locally and would offer as a follow-up PR rather than pushing here:

1. `ExtensionalConditions` is never bundled. Both clauses are proven for the same family, so the interesting test — Initial (*every* input doomed) and Terminal (doomed ⇒ rejects) pulling in opposite directions — does pass; only the conjunction is unstated.
2. `experiment_advantage` is never evaluated concretely. The tests establish `Pr[escape | sample] = 1/2`, which is the lemma's RHS; nothing confirms the `SecExp`/`OptionT` routing delivers it. It does — I checked.
3. **The trigger never fires anywhere.** In `oneRound` the extractor always fails, so `ExtractionCondition (1/2)` holds only because `error < Pr[escape]` stays silent. No instance exercises extraction *succeeding* under a firing trigger — nor the `relation`-holds branch of `extractionCondition_iff_isBounded`'s `by_cases`.

   `{ oneRound with relation := fun _ w => w = false }` fixes it in one line: at `error = 0` the trigger genuinely fires (`0 < 1/2`) and the condition holds *because extraction succeeds*. Against your existing `oneRound_not_extractionCondition_zero`, that brackets the definition from both sides and pins down that it measures the extractor, not the trigger.

Happy to open that as a separate PR once this lands.

PR #476, opened 2026-07-10
feat(UAKE): UAKE security notion

protoben:
This PR adds a model of unilaterally authenticated key exchange, a la [Dodis & Fiore 2017](https://eprint.iacr.org/2017/109). I've put it into `CryptoFoundations/AKE`, since it seems likely that multiple AKE definitions would be useful. This is just a simple and well accepted one.

Some caveats about the model are in the doc comment at the top of `UAKE/Defs.lean`.


alexanderlhicks:
> [Dodis & Fiore 2017](https://eprint.iacr.org/2014/144).

Wrong link :sweat_smile:

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
```diff
@@ -0,0 +1,252 @@
...
+open OracleSpec OracleComp
+
+namespace AKE.UAKE
+
+variable {K UK TK W : Type} {m : Type → Type}
+
+/-- A UAKE scheme with a fixed number of rounds. The keyed (authenticated) party
+  is T; the unkeyed (unauthenticated) party is U. -/
+structure Scheme (m : Type → Type) (K UK TK W : Type) where
+  rounds : ℕ
```
alexanderlhicks, line 61: 🤖 *AI-generated comment*

`rounds` is load-bearing but its unit is undocumented: `isPingPong`'s leader flag (`proto.rounds % 2 == 1`) and `CorrectExp`'s fuel (`proto.rounds + 1`) both assume it means **the total number of protocol messages** (so U→T then T→U is `rounds = 2`), not round-trips. Misdeclaring it fails *safe* (the honest relay stops being recognized as ping-pong, so a correct scheme becomes unprovable rather than a theorem unsound) — but silently. A docstring would prevent that, e.g.:

```lean
  /-- Total number of protocol messages exchanged (not round-trips). Under the
    DF'17 assumption that T speaks last, `rounds` odd means T also speaks first;
    the ping-pong predicate (`isPingPong`) relies on this parity. -/
  rounds : ℕ
```

Easy to validate: trace a 2-round honest relay through `opImpl` with the flag flipped — the timestamp interleaving comes out in the wrong order and `Matching` fails.

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
```diff
@@ -0,0 +1,252 @@
...
+  The protocol is correct if both (honest) parties output the same key (or
+  either party outputs ⊥). -/
+def CorrectExp [DecidableEq K] [Monad m] (proto : Scheme m K UK TK W) : m Bool := do
+  let (uk, tk) ← proto.setup
+  let (uOut, tOut) ← proto.U.runHonest proto.T uk tk (proto.rounds + 1)
+  return decide (uOut.join = none ∨ tOut.join = none ∨ uOut.join = tOut.join)
+
+def PerfectlyCorrect [DecidableEq K] [Monad m] [MonadLiftT m SPMF]
+    (proto : Scheme m K UK TK W) : Prop :=
+  Pr[= true | CorrectExp proto] = 1
```
alexanderlhicks: 🤖 *AI-generated comment*

Main suggestion of this review, applying to `PerfectlyCorrect`, `oracleImpl`, `Exp`, and `advantage` together: the game layer here deliberately avoids global lift instances in favor of the bundled `ProbCompRuntime m`. From `SecExp.lean`'s module docstring: “*Each experiment carries bundled subprobabilistic semantics, so the experiment can be interpreted through an internal semantic monad instead of requiring a global `MonadLiftT _ SPMF` instance on the ambient monad.*” The peer schemes all follow it — compare `PerfectlyCorrect (runtime : ProbCompRuntime m)` in `KeyEncapMech.lean`, `DataEncapMech.lean`, and `AsymmEncAlg/Defs.lean`, and `IND_CCA_Game`'s use of `runtime.liftProbComp ($ᵗ Bool)` / `runtime.evalDist` in `AsymmEncAlg/INDCCA.lean`.

Concretely: take `runtime : ProbCompRuntime m` here and in `Exp`/`advantage`, replace `[MonadLiftT ProbComp m]` uses with `runtime.liftProbComp`, and `[MonadLiftT m SPMF]`+`Pr[…]` with `runtime.evalDist`. That keeps UAKE composable with the reduction lemmas already stated against the runtime pattern — which matters most for foundational definitions like these.

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
```diff
@@ -0,0 +1,252 @@
...
+      | some t =>
+        match t.key with
+        | some _ => pure (.inr ())
+        | none => do
+          match ← (proto.T.step t.state w : m _) with
+          | .reject => pure (.inr ())
+          | .acceptAndSend st' w' done =>
+              let (tr1, c1) := recordOne t.transcript w env.clock
+              let (tr2, c2) := recordOne tr1 w' c1
+              let key ← if done then (proto.T.output st' : m _) else pure none
```
alexanderlhicks, line 172: 🤖 *AI-generated comment*

Corner case worth a note in `TSession.key`'s docstring: if `proto.T.output st'` returns `none` on a `done` step, then `key = none` is recorded, which the rest of the oracle reads as “session not completed” — the session stays steppable through the `t.key` gate above, and `revealT` returns `none`. This is only reachable for protocols violating the outputs-at-completion discipline, and it errs in the adversary-favoring (safe) direction, so no code change is needed — but the docstring currently states `none` ⟺ “session has not yet completed”, which this corner contradicts.

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
```diff
@@ -0,0 +1,252 @@
...
+  pingPong (proto.rounds % 2 == 1)
+    ((env.tSessions.filter (·.revealed)).map (·.transcript)) cr.challengeTr
+
+def finalize [DecidableEq W] [Monad m] [MonadLiftT ProbComp m]
+    {proto : Scheme m K UK TK W} (A : Adversary proto)
+    (st : A.State × Env proto × TK) (cr : ChallengeResult proto) (b : Bool) (K1 : Option K) :
+    m Bool := do
+  let (aSt, env, tk) := st
+  let Kb := if b then K1 else cr.K0
+  let (b', env') ← (simulateQ (oracleImpl proto tk) (A.post aSt Kb)).run env
```
alexanderlhicks: 🤖 *AI-generated comment*

Small fidelity gap vs. DF'17: the paper's post phase is `b′ ← A^{T(tk)}(K_b)` — oracle access to **T only** — whereas `A.post` shares `oracleSpec` with the first phase and can still issue `stepChallenge`, advancing U's challenge session if it never completed. Traced through, this looks benign (`cr.K0` is already extracted, and `fullPingPong` compares against the `cr.challengeTr` snapshot rather than the live challenge transcript), but it's a deviation the “Model simplifications” docstring doesn't list. Exact match with the paper is a one-liner — freeze the challenge session before the post phase:

```suggestion
  let env := { env with challengeDone := true }
  let (b', env') ← (simulateQ (oracleImpl proto tk) (A.post aSt Kb)).run env
```

(Untested — please confirm it compiles. Alternatively, keep the code and add a bullet to the model-simplifications docstring.)

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
```diff
@@ -0,0 +1,252 @@
...
+  else if !isPingPong cr then
+    return true
+  else
+    let K1 ← some <$> ($ᵗ K : ProbComp K)
+    finalize A st cr b K1
+
+noncomputable def advantage [SampleableType K] [DecidableEq W] [Monad m]
+    [MonadLiftT ProbComp m] [MonadLiftT m SPMF] {proto : Scheme m K UK TK W}
+    (A : Adversary proto) : ℝ :=
+  |(Pr[= true | Exp A]).toReal - 1 / 2|
```
alexanderlhicks: 🤖 *AI-generated comment*

Consider connecting this to the existing advantage API in `SecExp.lean` rather than introducing a new formulation: peers define e.g. `IND_CCA_Advantage runtime adversary := (IND_CCA_Game runtime adversary).boolBiasAdvantage`. For a total (failure-free) game the two conventions differ by exactly a factor of 2, and the bridging lemma already exists (`SPMF.boolBiasAdvantage_eq_two_mul_abs_sub_half`). `|Pr[true] − 1/2|` matches DF'17's Def. 8 directly, so keeping it is defensible — in that case a short comment noting the deliberate paper-convention choice and its relation to `boolBiasAdvantage` (plus routing through `runtime.evalDist` per the comment on `PerfectlyCorrect`) would keep downstream reduction algebra consistent.

VCVio/CryptoFoundations/AKE/UAKE/Party.lean
```diff
@@ -0,0 +1,104 @@
...
+/-- True if the final output message of a party is deterministic given the
+   state. I.e., all non-determinism is in the init and step functions. -/
+def RecoveryDeterministic [Monad m] (P : Party m In W Out) : Prop :=
+  ∀ st : P.State, ∃ out, P.output st = pure out
+
+/-- True if a party outputs a final message only once the protocol is complete. -/
+def OutputsAtCompletion [MonadLiftT m SetM] (P : Party m In W Out) : Prop :=
+  (∀ i r, r ∈ support (P.init i) → ∀ out ∈ support (P.output r.state), out = none) ∧
+    (∀ st w st' w', StepResult.acceptAndSend st' w' false ∈ support (P.step st w) →
+      ∀ out ∈ support (P.output st'), out = none)
```
alexanderlhicks: 🤖 *AI-generated comment*

Both predicates are currently unused — `git grep -n 'RecoveryDeterministic\|OutputsAtCompletion'` shows only these definitions. Notably, `OutputsAtCompletion` is exactly the property that model-simplification bullet 3 in `Defs.lean` leans on (“we do not enforce that U outputs K0 only at completion, which DF'17 (implicitly) assumes”). Two reasonable options:

1. Bundle them into a named well-formedness predicate (e.g. `Scheme.WellFormed`) that future security theorems take as a hypothesis, and reference it from that docstring; or
2. Defer them to the PR that first needs them.

Option 1 seems more in the spirit of a foundational definitions PR — it turns the docstring's informal WLOG argument into a checkable interface.

VCVio/CryptoFoundations/AKE/UAKE/Transcript.lean
```diff
@@ -0,0 +1,74 @@
...
+def merge {A : Type} (t1 t2 : Transcript A) : Transcript A :=
+  ⟨(t1.entries ++ t2.entries).mergeSort fun p q => decide (p.2 ≤ q.2)⟩
+
+def combine {A B : Type} (ta : Transcript A) (tb : Transcript B) :
+    Transcript (A ⊕ B) :=
+  (ta.relabel Sum.inl).merge (tb.relabel Sum.inr)
+
+def prefixWith {A B : Type} (a : A) (t : ℕ) (tb : Transcript B) :
+    Transcript (A ⊕ B) :=
+  ⟨(Sum.inl a, t) :: tb.entries.map fun p => (Sum.inr p.1, p.2)⟩
```
alexanderlhicks: 🤖 *AI-generated comment*

`relabel`, `merge`, `combine`, and `prefixWith` have no consumers in this PR — they are used only by each other (validate: `git grep -nE 'relabel|combine|prefixWith' -- VCVio` returns only these lines). If they're groundwork for a later development (e.g. assembling the global interleaved transcript for proofs about `Matching`), a one-line doc comment saying so would justify keeping them; otherwise consider deferring them to the PR that uses them, keeping this foundational file minimal.

VCVio/CryptoFoundations/AKE/UAKE/Transcript.lean
```diff
@@ -0,0 +1,74 @@
...
+  | ab, (a, b) :: rest => (if ab then [a, b] else [b, a]) ++ interleave (!ab) rest
+
+/-- Def. 3 from DF'17. A pair of transcripts match (T ⊆ T*) if their messages
+   are elementwise identical and their timestamps are "interleaved" as t₁ < t₁*
+   < t₂* < t₂ < ... or t₁* < t₁ < t₂ < t₂* < ..., depending on which party
+   speaks first. -/
+def Matching (oracleLeadsFirst : Bool) (T Tstar : Transcript W) : Prop :=
+  T.entries.map Prod.fst = Tstar.entries.map Prod.fst ∧
+    List.IsChain (· < ·)
+      (interleave oracleLeadsFirst ((T.entries.map Prod.snd).zip (Tstar.entries.map Prod.snd)))
```
alexanderlhicks, line 38: 🤖 *AI-generated comment*

A positive one: this machinery was stress-tested externally (a faithful re-implementation of the `recordOne`/`opImpl` clock bookkeeping plus this `Matching` predicate), and it checks out against DF'17 Def. 3 in every case tried: honest relays for 2–7 rounds under both leader parities match exactly; single-message substitution fails to match; with two interleaved oracle sessions, only the honestly-relayed one matches; and a session opened or extended *after* the challenge completes can never retroactively match a challenge snapshot — its final message's timestamp violates the alternation, by clock monotonicity. That last point means the `fullPingPong`-on-snapshot design in `Defs.lean` can't be gamed from the post phase.

Suggested (non-blocking) follow-up: a machine-checked version of the first fact — e.g. a lemma that the honest-relay adversary against a toy 2-round protocol satisfies `isPingPong` — would lock in the parity flag and `interleave` bookkeeping, the easiest places for a silent regression later.

alexanderlhicks (commented):
## 🤖 AI-generated review

This review was produced by an AI assistant (Claude Code), invoked and supervised by @alexanderlhicks. Each finding was validated before posting (methods noted per comment), but please treat it as reviewer input to verify, not authority.

### Soundness vs. the reference

First, a housekeeping item: **the PR description links the wrong ePrint entry** — [2014/144](https://eprint.iacr.org/2014/144) is Andreeva et al., *How to Securely Release Unverified Plaintext in Authenticated Encryption*. The docstring in `Defs.lean` has the right one: [ePrint 2017/109](https://eprint.iacr.org/2017/109) (Dodis & Fiore, *Unilaterally-Authenticated Key Exchange*, FC 2017). Worth fixing the description for posterity.

I cross-checked the formalization against the paper, and it holds up well:

- `Exp`/`finalize` match the Sec. 3 `Exp^UAKE-Sec` experiment branch-for-branch: `K0 = ⊥ ⟹ K1 = ⊥`; `K0 ≠ ⊥ ∧` not ping-pong `⟹` adversary wins (authenticity); otherwise real-or-random `K1` with the full-ping-pong coin-flip nullifier. The reveal bookkeeping (`revealed` persisting into the post phase, `fullPingPong` checked on post-phase state against the challenge-phase transcript snapshot) correctly captures the paper's “one additional query” formulation.
- `Matching` realizes Def. 3: simulating the oracle's clock/recording logic for honest relays of 2–7 rounds (both leader parities) reproduces the paper's alternation exactly; message substitution and cross-session confusion correctly fail to match; and — a nice robustness property of the snapshot design — sessions opened or extended *after* the challenge completes can never retroactively match, by clock monotonicity. Details in the inline comment on `Matching`.
- `CorrectExp` is faithful to Def. 7, including its (the paper's own) weak conditional form; the docstring is upfront about this.
- The documented model deviations (early reveal, reject-and-retry, unenforced well-formedness) are all adversary-favoring, so security in this model implies security in the paper's model. The “Model simplifications” docstring is genuinely excellent — more definitions PRs should do this.

Two soundness-adjacent polish items are flagged inline (post-phase `stepChallenge` availability; `rounds` semantics being load-bearing but undocumented) — both easy to validate and cheap to address.

### Main suggestions (details inline)

1. **Adopt `ProbCompRuntime`** in place of the `[MonadLiftT ProbComp m]` / `[MonadLiftT m SPMF]` constraints — this is the established pattern across `CryptoFoundations`, and it keeps UAKE composable with the reduction infrastructure already stated against it.
2. **Connect `advantage` to the `SecExp.lean` helpers**, or document the deliberate paper-convention choice.
3. **Trim or wire in currently-unused definitions** (`Transcript.relabel`/`merge`/`combine`/`prefixWith`, `Party.RecoveryDeterministic`, `Party.OutputsAtCompletion`).

### Minor

- Docstring gaps on the public surface: `Scheme` fields (especially `rounds`), `PerfectlyCorrect`, `oracleSpec`, `Adversary`/`ChallengeResult` fields, `challengeSession`, `finalize`, `advantage`, `Session`, `interleave`, `pingPong`, `recordOne`/`recordOpt`, `runHonest`(`Loop`), and the `StepResult.acceptAndSend` `done` flag (“this send is my final message”). A sentence on the deliberate double-`Option` in `Party.output` (outer = “no output yet”, inner = “output ⊥”) would also help readers — it takes a while to reconstruct.
- Peer files are universe-polymorphic in the ambient monad (`m : Type → Type v`, cf. `INDCCA.lean`); consider matching.
- Naming: peers use `*_Game` / `*_Advantage` (e.g. `IND_CCA_Game`); `AKE.UAKE.Exp`/`advantage` are fine namespaced, but worth a consistency thought.
- `challengeSession` returns the `tk` it was handed only to thread it to `finalize`; passing `tk` to `finalize` directly would simplify the tuple.

### Suggested follow-up (not blocking)

A machine-checked sanity lemma that the honest-relay adversary is ping-pong (for a toy protocol), and/or a toy instantiation under `Examples/`, would lock in the trickiest machinery and validate the definition's ergonomics.

Overall: careful, well-documented work — the two-phase stateful `Adversary` and the `HasQuery.toQueryImpl … + opImpl` composition mirror the `INDCCA` idiom nicely, and the game logic is faithful to the paper precisely in the places that are easiest to get wrong.

protoben:
> > [Dodis & Fiore 2017](https://eprint.iacr.org/2014/144).
> 
> Wrong link 😅

Oops! Fixed. Thanks for catching that.

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
protoben: Good point. Added a comment clarifying this.

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
protoben: Changed these to use ProbCompRuntime.

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
protoben: Thanks! Fixed and tested.

dtumad (commented):
I did a fresh pass over head a4e1d66 against Dodis–Fiore 2017, current VCVio APIs, and a current-main merged build. The full VCVio library builds cleanly. The main Exp/finalize split tracks the paper well: bottom challenge keys stay bottom in both branches; a non-bottom, non-ping-pong challenge is an authenticity win; and a ping-pong challenge enters the real-or-random phase, with full ping-pong neutralized by an independent bit. Freezing the challenge before post is also the right behavior.

I also machine-checked small Matching fixtures for honest 2-message and 3-message relays under both T-last parities, plus wrong-parity, message-substitution, and post-challenge-session cases. They behave as intended. The state-machine Party API also seems reasonable alongside the newer PolyFun interaction trees: rejection, retry, and concurrently addressed sessions are genuinely dynamic, so a bridge can be follow-up work rather than a prerequisite here.

One semantic edge is worth tightening before merge. revealT currently sets revealed := true even when t.key = none. DF17 permits the extra reveal only after T has produced its last message; the BR presentation likewise changes the reveal flag only for an accepted session. For a fixed adversary, the current early-reveal rule can only penalize it, so this particular game is slightly weaker than the paper game. The defer-the-query argument is plausible when quantifying over unrestricted adversaries, but it is not a pointwise equality of the advantage defined here. A small adjustment would align the model directly: return none without changing revealed while t.key is none, and mark the session only once a key is present. Keeping the current convention is also possible, but then describing it explicitly as a game variant would avoid overstating the correspondence.

Two documentation refinements would make the scope especially clear:

- Drop-and-retry is more permissive than halting, but it is not literally the alternative giving the adversary the most power: record-and-continue also permits retry and can prevent transcript matching after an injected rejected message. Stating the chosen convention without the dominance claim would be more precise.
- rounds, T-speaks-last, fixed transcript length, and output-at-completion are trusted well-formedness conditions. Not every violation is automatically insecure; for example, an incorrect message count with the same parity can leave Matching unchanged. Framing correspondence with DF17 as conditional on those conventions would be enough here. A named Scheme.WellFormed interface using the existing predicates could be a useful later addition, not necessarily part of this PR.

This looks close after those localized clarifications. Banking the small 2/3-message Matching fixtures in a test or theorem file would also be useful because parity and timestamp orientation are the easiest parts to regress, though I view that as a strong follow-up suggestion rather than a redesign request.

VCVio/CryptoFoundations/AKE/UAKE/Transcript.lean
protoben: Good point. Fixtures added.

VCVio/CryptoFoundations/AKE/UAKE/Transcript.lean
protoben: Agreed. These were helpers I used in a proof, but they don't belong here. I've removed them.

VCVio/CryptoFoundations/AKE/UAKE/Party.lean
protoben: I've added a `Scheme.WellFormed` predicate that I think addresses both this and the related ask in Devon's comment below.

To be clear, I think well-formedness in this case relates to the protocol being a fixed number of rounds (in the honest case) and having parties that output a key only at the end, since these seem to be implicit assumptions made by DF'17. I haven't incorporated `RecoveryDeterministic`. I don't think that's a well-formedness condition, since a perfectly valid UAKE can have parties with randomized output; rather, it's something you would game-hop to in a proof to eliminate randomness from the output stage. I don't think it actually belongs in this PR, so I've removed it.

protoben:
> I did a fresh pass over head [a4e1d66](https://github.com/Verified-zkEVM/VCVio/commit/a4e1d66f0480947211b4f513f3d07e83f68fb46a) ...

Thanks! That was a very helpful series of comments!

I believe I addressed them all, including adding a `Scheme.WellFormed` predicate and fixtures for `Matching`. One note:

> One semantic edge is worth tightening before merge. revealT currently sets revealed := true even when t.key = none. DF17 permits the extra reveal only after T has produced its last message; the BR presentation likewise changes the reveal flag only for an accepted session.

The DF'17 and BR versions differ slightly here (as you correctly point out) in that DF'17 allows reveal queries on rejected sessions (`t.key = some none`), whereas BR allows reveals only for accepted sessions. I've gone with the DF'17 version in my fix.

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
protoben: Good point. Devon made a similar suggestion below. I've addressed this. See my comment here for a little more info: https://github.com/Verified-zkEVM/VCVio/pull/476#issuecomment-5186332574

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
protoben: Thanks! I've addressed this by adding
1. a comment about how the current 1/2-biased advantage is intentionally non-idiomatic in order to closely match the advantage defined in DF'17 and what the ramifications of this are;
2. an additional `UAKE.boolBiasAdvantage` that uses the existing advantage API; and
3. a lemma and a `NeverFail` instance that bridges the gap in the common `ProbCompRuntime.probComp` case.

protoben:
Thanks for the reviews! I've resolved all outstanding comments and merged in the current main (as of 8/13). Feel free to take another look.

dtumad:
Did a final review pass — everything from the earlier review rounds is addressed (the `revealT` no-op fix, `Scheme.WellFormed`, the doc refinements, the banked fixtures), and the `boolBiasAdvantage` addition looks right. Content-wise this is ready to merge.

One mechanical blocker surfaced after your 8/13 main merge: #510 migrated the repo to Lean's module system, and a `module` file cannot import a legacy-prologue file — so `lake build VCVio` now fails at `VCVio.lean` with `cannot import non-\`module\` VCVio.CryptoFoundations.AKE.UAKE.Defs from \`module\``. (CI never caught it because the full workflow hasn't run on the branch since that push.) The fix is just the standard prologue for the four new files: `module`, `public import`, and `@[expose] public section` after the module docstring. I validated the patch below against current main — full library build, style lint, and isolation checks all pass with it applied.

The fork doesn't allow maintainer edits, so applying it is yours to do — this exact patch (also fine to just re-type the three-line change per file):

```diff
diff --git a/VCVio/CryptoFoundations/AKE/UAKE/Defs.lean b/VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
index 13df622f..9a4caa6b 100644
--- a/VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
+++ b/VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
@@ -3,9 +3,11 @@ Copyright (c) 2026 Galois Inc. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Ben Hamlin
 -/
-import VCVio.CryptoFoundations.AKE.UAKE.Party
-import VCVio.CryptoFoundations.AKE.UAKE.Transcript
-import VCVio.OracleComp.ProbCompLift
+
+module
+public import VCVio.CryptoFoundations.AKE.UAKE.Party
+public import VCVio.CryptoFoundations.AKE.UAKE.Transcript
+public import VCVio.OracleComp.ProbCompLift
 
 /-!
 # UAKE Core Definitions
@@ -43,6 +45,8 @@ Model simplifications
   fewer than 2 rounds.
 -/
 
+@[expose] public section
+
 open OracleSpec OracleComp
 
 namespace AKE.UAKE
diff --git a/VCVio/CryptoFoundations/AKE/UAKE/Party.lean b/VCVio/CryptoFoundations/AKE/UAKE/Party.lean
index fdf937df..3e97a4ec 100644
--- a/VCVio/CryptoFoundations/AKE/UAKE/Party.lean
+++ b/VCVio/CryptoFoundations/AKE/UAKE/Party.lean
@@ -3,9 +3,11 @@ Copyright (c) 2026 Galois Inc. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Ben Hamlin
 -/
-import VCVio.CryptoFoundations.SecExp
-import VCVio.OracleComp.SimSemantics.Append
-import VCVio.OracleComp.SimSemantics.SimulateQ
+
+module
+public import VCVio.CryptoFoundations.SecExp
+public import VCVio.OracleComp.SimSemantics.Append
+public import VCVio.OracleComp.SimSemantics.SimulateQ
 
 /-!
 # UAKE Scheme Party Definitions
@@ -18,6 +20,8 @@ is the key). It is up to the protocol realization to ensure that the output
 function produces output only at the end.
 -/
 
+@[expose] public section
+
 open OracleSpec OracleComp
 
 namespace AKE.UAKE
diff --git a/VCVio/CryptoFoundations/AKE/UAKE/Transcript.lean b/VCVio/CryptoFoundations/AKE/UAKE/Transcript.lean
index cf8df7a8..da397a55 100644
--- a/VCVio/CryptoFoundations/AKE/UAKE/Transcript.lean
+++ b/VCVio/CryptoFoundations/AKE/UAKE/Transcript.lean
@@ -3,7 +3,9 @@ Copyright (c) 2026 Galois Inc. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Ben Hamlin
 -/
-import Mathlib.Data.List.Chain
+
+module
+public import Mathlib.Data.List.Chain
 
 /-!
 # Transcript Definition for UAKE from DF'17
@@ -12,6 +14,8 @@ The transcripts in DF'17 bundle messages with timestamps from a global clock
 incremented whenever a party sends a message.
 -/
 
+@[expose] public section
+
 namespace AKE.UAKE
 
 variable {W : Type}
diff --git a/VCVio/CryptoFoundations/AKE/UAKE/TranscriptFixtures.lean b/VCVio/CryptoFoundations/AKE/UAKE/TranscriptFixtures.lean
index 91bef688..0c07c445 100644
--- a/VCVio/CryptoFoundations/AKE/UAKE/TranscriptFixtures.lean
+++ b/VCVio/CryptoFoundations/AKE/UAKE/TranscriptFixtures.lean
@@ -3,7 +3,9 @@ Copyright (c) 2026 Galois Inc. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Ben Hamlin
 -/
-import VCVio.CryptoFoundations.AKE.UAKE.Defs
+
+module
+public import VCVio.CryptoFoundations.AKE.UAKE.Defs
 
 /-!
 # Transcript Fixtures
@@ -20,6 +22,8 @@ tick.
 These fixtures do *not* exercise `opImpl`, `recordOne`, or `recordOpt`.
 -/
 
+@[expose] public section
+
 open OracleSpec OracleComp
 
 namespace AKE.UAKE
```

Once pushed, CI will run and this can merge.

quangvdao (changes requested):
Reviewed at `5aadeddeaf9ad8618fbeab984e35ea695f5823b6`.

Content-wise, the earlier semantic findings are closed: the experiment tracks DF'17's branches, early reveal is a no-op until a key exists, the challenge session is frozen before post, the well-formedness boundary is named, the advantage conventions are bridged, and the transcript/ping-pong fixtures cover the parity and timestamp edge cases. I found no new formalization blocker in those definitions.

There is one mechanical P1 blocker: `lake build VCVio` fails because all four new Lean files still use the legacy prologue after the repository migrated to the module system:

- `VCVio/CryptoFoundations/AKE/UAKE/Defs.lean`
- `VCVio/CryptoFoundations/AKE/UAKE/Party.lean`
- `VCVio/CryptoFoundations/AKE/UAKE/Transcript.lean`
- `VCVio/CryptoFoundations/AKE/UAKE/TranscriptFixtures.lean`

Each needs `module`, `public import`, and `@[expose] public section` in the standard positions. The exact validated patch is already posted in [this comment](https://github.com/Verified-zkEVM/VCVio/pull/476#issuecomment-5352086071). The live `Build Project` check is failing at this head for the same reason. Please apply that patch and rerun CI; after it is green, this is ready to merge.

protoben:
Oops. fixed this locally (very similar to your patch), but neglected to push it.

I've applied your diff instead and merged in the newest main. Should be fixed, but let me know if there are still any issues.

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
```diff
@@ -0,0 +1,337 @@
...
+def PerfectlyCorrect [DecidableEq K] [Monad m] (proto : Scheme m K UK TK W)
+    (runtime : ProbCompRuntime m) : Prop :=
+  Pr[= true | runtime.evalDist (CorrectExp proto)] = 1
+
+/-- True if
+  1. both parties output iff the protocol run is complete, and
+  2. an honest run of the protocol transfers exactly `rounds` messages.
+
+  This does *not* enforce the WLOG T-speaks-last convention from DF'17. -/
+def Scheme.WellFormed [Monad m] [MonadLiftT m SetM] (proto : Scheme m K UK TK W) : Prop :=
```
quangvdao: **P1 — `WellFormed` accepts vacuous support semantics.** Both this predicate and `Party.OutputsOnlyAtCompletion` require only `MonadLiftT m SetM`. I compiled a counterexample monad whose lift maps every computation to empty support; an always-outputting malformed party then satisfies both predicates because every support premise is false. This also leaves the predicate unrelated to the `ProbCompRuntime` later used to interpret correctness and security. Please require at least `LawfulMonadLiftT m SetM`, and preferably make well-formedness relative to the same runtime or add an explicit support/runtime coherence bridge. A negative canary for output-on-init would pin the invariant.

VCVio/CryptoFoundations/AKE/UAKE/TranscriptFixtures.lean
```diff
@@ -0,0 +1,263 @@
+/-
+Copyright (c) 2026 Galois Inc. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Ben Hamlin
+-/
+
+module
```
quangvdao: **P2 — this module currently fails the repository warning budget.** On the exact PR head, `lake build VCVio` reports `The current module only contains private declarations`; Build Project rejects every non-`sorry` warning under `VCVio/`. Since this is test-only material and is publicly imported from `VCVio.lean`, please move it to `VCVioTest/`, remove the production umbrella import, and register it there. Named public canary theorems would also avoid the private-only module shape.

VCVio/CryptoFoundations/AKE/UAKE/TranscriptFixtures.lean
```diff
@@ -0,0 +1,263 @@
...
+Machine-checked fixtures for `Matching` (DF'17 Def. 3), `pingPong` (Def. 4), and
+the `isPingPong` / `fullPingPong` wrappers used by the security experiment.
+
+For clarity, the transcripts are written out literally. The positive tests
+contain the timestamps an honest relay produces under the clock discipline of
+`recordOne` / `recordOpt`: each recorded message consumes one tick, and the
+adversary relaying a message to the other session records it there at a later
+tick.
+
+These fixtures do *not* exercise `opImpl`, `recordOne`, or `recordOpt`.
```
quangvdao: **P2 — producer coverage stops before the security game.** This note correctly says the fixtures do not exercise the recording helpers; they also do not directly test `runHonest`, `WellFormed`, `challengeSession`, `finalize`, `Exp`, or the advantage branches. I validated those paths with a scratch two-round protocol, but those canaries are not committed. Please bank a small toy protocol/adversary suite covering bottom `K₀`, the authenticity win path, ping-pong real/random behavior, full-ping-pong neutralization, frozen challenge state, invalid sessions, early reveal, post-completion no-op behavior, and exact clock/orientation.

quangvdao (changes requested):
Ben, thank you for the careful work here—and for thoughtfully addressing the earlier rounds of feedback. I re-reviewed the current head `b6d2c8` against DF17 and the repository's current policies.

The core security-game logic is in good shape: the experiment/finalization branches are faithful to DF17; the challenge state is frozen before `post`; early reveal is a no-op until completion; matching uses the right noncommutative orientation and timestamp parity; and the absolute-value advantage is consistent with Definition 8. I found no core branch, clock, transcript-orientation, or reveal bug, and no new trust-boundary issue.

I am still requesting changes for three merge blockers, detailed inline:

1. `WellFormed` can be proved vacuously with a lawful-looking but unconstrained `MonadLiftT m SetM` instance, so it does not yet establish its advertised invariant.
2. The production-imported fixture module emits a deterministic warning that the repository's clean-build warning budget rejects.
3. Committed tests stop at transcript predicates and do not pin the actual experiment/runtime behavior.

I validated the exact head with the targeted and umbrella builds, style lint, module-list check, agent-doc and isolation checks, `git diff --check`, and the axiom sweep. The umbrella build succeeds apart from the fixture warning above; the hosted substantive workflows remain approval-gated.

Nonblocking follow-ups: narrow the `Party` imports, generalize the monad universe, and add public API docstrings. The full PPT/negligibility/`SecurityGame` bridge can reasonably be separately scoped, but the PR should be explicit that it currently provides the fixed-parameter experiment and per-adversary advantage.

Once the three blockers are addressed, this looks close to merge-ready. Thanks again, Ben—this is a substantial and useful foundation.

dtumad:
Deep semantic audit across the open PR set (semantics only). This is the closest-to-merge PR in the lattice/crypto family; the 08-24 review's three items stand, and I have one new finding on the notion itself.

**MAJOR (new): correctness and security are jointly satisfied by the always-⊥ scheme — the notion pair has no liveness condition.** Two verified facts at head `b6d2c80`:

- `CorrectExp` (`Defs.lean:80–83`) returns `decide (uOut.join = none ∨ tOut.join = none ∨ uOut.join = tOut.join)`, and `.join` collapses "has not output yet" and "output ⊥" into the same `none` — so a protocol that never completes, or whose honest parties always output ⊥, is `PerfectlyCorrect`.
- `Exp` (`Defs.lean:287–289`) routes `cr.K0.isNone` to `finalize` with `K1 := none`, so `Kb = none` regardless of the bit and the adversary's post-phase view is independent of `b` — advantage exactly 0.

So the trivial scheme is perfectly correct *and* perfectly secure under this pair. That mirrors DF'17's own conditional Def. 7 (and the docstrings are upfront about the conditional form), but a reader will take `PerfectlyCorrect ∧ advantage ≈ 0` as meaningful, and today it is not. A named liveness predicate (honest runs produce a non-`none` key with probability 1) plus one line in the docstrings connecting the three notions would close this cheaply. For contrast, the notion *is* non-vacuous against the constant-key scheme — an adversary that completes the challenge session solo with no matching T session wins via the `!isPingPong` branch — so the authenticity content is real; it's only silent-failure schemes that slip through.

Standing items from the 08-24 review (unaddressed — it landed 31 minutes after the last push, so likely just not seen yet):

1. `Scheme.WellFormed` and `Party.OutputsOnlyAtCompletion` are provable vacuously under an unlawful `MonadLiftT m SetM` (empty-support lift falsifies every premise). They need `LawfulMonadLiftT` — or better, relativisation to the same `ProbCompRuntime` that interprets correctness and security, since `WellFormed` is load-bearing for the `rounds`-parity convention `isPingPong` depends on.
2. `TranscriptFixtures.lean` publicly imported from `VCVio.lean` → move to `VCVioTest/`.
3. Tests stop at transcript predicates; nothing pins `Exp`/`finalize`/`challengeSession`/advantage branches.

Rebase note: the head predates the measure-first change — `runtime.evalDist` no longer exists on main (`ProbCompRuntime.evalSPMF`); its five uses here (`PerfectlyCorrect`, `advantage`, `boolBiasAdvantage`, the `NeverFail` instance) are statement rewrites, not import fixes.

Two scope items to state explicitly in the module docstring rather than leave implicit: (i) there is no Corrupt / long-term-key-exposure oracle, so the notion captures neither forward secrecy nor KCI (faithful to DF'17 §3, but say so); (ii) the PR description links ePrint 2014/144, which is a different paper — the correct reference (ePrint 2017/109, as in the `Defs.lean` docstring) should be in the body too.

VCVio/CryptoFoundations/AKE/UAKE/TranscriptFixtures.lean
protoben: I added additional fixtures for the game. (Actually, I had Claude add them, but they seem correct to me.) You may want to check that I covered everything you were interested in, but I think I got it all.

VCVio/CryptoFoundations/AKE/UAKE/TranscriptFixtures.lean
protoben: Moved these and added public canaries.

VCVio/CryptoFoundations/AKE/UAKE/Defs.lean
protoben: I added a `MonadLiftT m SetM` requirement. Currently working on the runtime issue.

dtumad:
I prepared an alignment of this PR with the current native measure APIs at
[`14fc4735`](https://github.com/Verified-zkEVM/VCVio/commit/14fc4735cc201e787a4932307c17ec35a4d44c5d),
on [`landing/uake-476-20260917`](https://github.com/Verified-zkEVM/VCVio/tree/landing/uake-476-20260917).
It preserves your commits and attribution, with this original PR as the landing vehicle.

The repair ports the game probabilities to native measures, adds a lawful runtime-coherence
certificate, and keeps the paper's conditional correctness separate from an explicit honest
completion condition. Positive two-/three-message models and rejection/zero-observer counterexamples
exercise the admission boundary. It makes no security claim about the toy protocols.

`./scripts/validate.sh --lint --test --axioms` passes on Lean 4.34.0, with no new sorry,
nonstandard axiom, global lint, or exposure debt.

The fork currently disables maintainer edits and does not grant direct push access. Could you
apply the repair branch to `uake`, or enable maintainer edits so we can continue on this PR?

To apply the reviewed commits while preserving history:

```sh
git fetch https://github.com/Verified-zkEVM/VCVio.git landing/uake-476-20260917
git switch uake
git merge --ff-only FETCH_HEAD
```

protoben:
Thanks! I had this work queued for myself, but hadn't gotten to it yet.

I'd like to allow maintainer edits, but the option isn't showing up for me for some reason. (Possibly because the fork is under the GaloisInc github group, rather than my personal account?) Anyway, I've pulled in those changes using your instructions above and pushed the results to the PR branch.

PR #478, opened 2026-07-12
feat(GPV): generic GPV hash-and-sign framework with EUF-CMA collision bounds

alik-eth:
Extracts the generic GPV development out of #466 into its own self-contained PR, per @quangvdao's review there. #466 will be rebased on top of this branch and retitled to carry only the idealized-Falcon instantiation.

## What this adds

A generic hash-and-sign framework in `VCVio/CryptoFoundations/GPVHashAndSign/` following [GPV08], parameterized by a preimage-sampleable function (PSF), with the EUF-CMA reduction of [FGdG+25, Thm 1] shape:

- `GPVHashAndSign.euf_cma_split_bound` — the headline: any EUF-CMA forger's advantage is bounded by a PSF-collision adversary's advantage, plus an exact-match term `(qSign + qHash) · programmedPreimageAdvantage`, plus the salt-collision birthday bound `(qSign + qHash)² / (2 · |Salt|)`.
- `GPVHashAndSign.euf_cma_collision_bound` — the collision-only corollary under a preimage-min-entropy hypothesis.
- `GPVHashAndSign.forgery_yields_collision_or_exact_match` / `forgery_yields_collision` — the core game hops (pre-sampled embed index, never-overwrite cache; no reservoir overwrite-restore).
- `Examples/GPVNonVacuity.lean` — a hypothesis-consistency (inhabitance) witness: a toy bijective PSF instance satisfying all side conditions simultaneously. Logical consistency only; no quantitative security content.

## Module layout

Split along proof phases per review of #465/#466: `Basic → CollisionTelescope → Factorization → GameRuns → TapeFactorization → FlagHandlers → CombinedHandler → GameIdentification → VerificationBridge → TrapCount → EmbedIndex → Security`, re-exported by the umbrella module (no consumer or name changes).

The only touch outside the new framework is an 8-line arity adaptation of the two *sorried* legacy Falcon placeholder statements (`collisionBound` now takes `qHash`); that file is rewritten wholesale in #466.

## Verification

- `lake build ToMathlib VCVio FFI LatticeCrypto HashSig Examples VCVioWidgets` green on `leanprover/lean4:v4.31.0`, zero non-sorry warnings.
- Kernel axiom check: all four headline theorems and the witness report `[propext, Classical.choice, Quot.sound]`.

## Scope caveats (disclosed in the code)

- The forge-a-queried-point convention (`hForge`) is the standard ROM WLOG, stated as a hypothesis; the append-query compiler discharging it is planned follow-up work (#466 thread).
- Query counts are pinned by the `signHashQueryBound` hypothesis; reductions are not cost-tracked.


alik-eth:
The disclosed hForge caveat is now discharged: pushed `GPVHashAndSign/AppendQuery.lean` (additive module) formalizing the append-forgery-query compiler and the all-adversaries corollary.

- `appendForgeQuery adv` — the compiled adversary that re-queries its own forgery point as a final ambient RO query.
- `signHashQueryBound_appendForgeQuery` — sign count unchanged, hash count `qHash + 1`.
- `forgesQueriedPoint_appendForgeQuery` — the compiled adversary satisfies `ForgesQueriedPoint` unconditionally.
- `advantage_appendForgeQuery` — the compiler is advantage-**preserving** (equality, not a bound): the appended query is re-issued by the verification read at the same point, so the lazy-caching random oracle absorbs it without affecting the output distribution, the signing log, or the freshness mask.
- **`euf_cma_split_bound_of_queryBound`** — the headline: the split GPV PFDH bound for *any* adversary, with only the query-bound hypothesis — no `ForgesQueriedPoint` — at the price of `qHash ↦ qHash + 1` in the exact-match multiplier and the salt-collision bound.

All twelve new declarations kernel-check at `[propext, Classical.choice, Quot.sound]`; full CI library set green, lint clean. The "append one hash query" WLOG that was previously prose in the module docstring is now a theorem.

alik-eth:
Pushed one more generic module: `GPVHashAndSign/SamplerTransport.lean` — the adaptive trapdoor-sampler-swap accumulation. For any two PSFs sharing `eval`/`isShort` whose trapdoor samplers are within per-call total-variation `ε` on honest keys, any adversary's advantage transports between the two schemes at cost `qSign · ε` (`advantage_le_advantage_add_of_trapdoorSample_tvDist`, plus the two-sided form). Proven sorry-free by charging the signing steps through the generic `tvDist_simulateQ_run_le_queryBoundP_mul` telescope; uniform/RO/verification steps are proven implementation-identical. This is the engine for discharging Falcon's transport hypothesis in #466. All declarations `[propext, Classical.choice, Quot.sound]`.

VCVio/CryptoFoundations/GPVHashAndSign/GameIdentification.lean
```diff
@@ -0,0 +1,720 @@
...
+      refine (ENNReal.div_lt_top ?_ ?_).ne
+      · simp
+      · simp only [ne_eq, mul_eq_zero, OfNat.ofNat_ne_zero, Nat.cast_eq_zero, false_or]
+        exact Fintype.card_ne_zero)
+    (probEvent_saltSeq_le_collisionBound Salt qSign qHash c hcache))
+
+/-! ## Step 1 and the wiring to the headline bounds
+
+**Step 1 (sign-then-hash ≡ real).** `gpv_tvDist_real_programmed_le_collisionBound` *consumes* the
+direct front-tape coupling `gpv_tvDist_tape_runs_le_collisionBound`: after the front-tape
```
dtumad, line 696: This does not match the live proof. `gpv_tvDist_real_programmed_le_collisionBound` (`:513-536`) is `le_trans (gpv_tvDist_orig_run_le_probEvent_flag …) (gpv_orig_flag_le_collisionBound …)` — the original-run route, as the comment at `:530-533` says. `gpv_tvDist_tape_runs_le_collisionBound` is not on this path.

VCVio/CryptoFoundations/GPVHashAndSign/AppendQuery.lean
```diff
@@ -0,0 +1,400 @@
...
+/-! ## The all-adversaries split bound -/
+
+/-- **Split GPV PFDH bound without the `ForgesQueriedPoint` hypothesis.** For *any* adversary
+obeying the query bound `(qSign, qHash)`, the EUF-CMA advantage satisfies the split GPV bound
+at hash budget `qHash + 1`: the append-forgery-query compiler `appendForgeQuery` discharges the
+forger-queries-its-forgery-point convention at the cost of one extra hash query
+(`forgesQueriedPoint_appendForgeQuery`, `signHashQueryBound_appendForgeQuery`) while preserving
+the advantage exactly (`advantage_appendForgeQuery`).  This is the all-adversaries form of
+`euf_cma_split_bound`; the witnesses are the reductions built from the compiled adversary. -/
+theorem euf_cma_split_bound_of_queryBound [DecidableEq Domain]
```
dtumad, line 326: These two are the endpoints I would point consumers at, since they are what keeps `ForgesQueriedPoint` from becoming an unstated convention in each scheme instantiation. Worth saying so in the module docstring.

dtumad (commented):
I went through this looking for a load-bearing premise that is false or vacuous, and did not find one. Notes below are all non-blocking.

The replacement statement is the right one, and it is stronger than "the old proof needed finishing". The previous collision-only endpoint quietly omitted the case where the forgery matches the programmed preimage exactly; that case is not a collision under `psf.eval`, so the old bound did not cover it. This PR exposes it as `programmedPreimageAdvantage` in the split theorem and then prices it by the trapdoor-sample guessing probability, instead of setting it to zero and moving on. Discovering that the old statement was not generally true is the useful outcome here.

What I checked in the spine: the inline-handler identical-until-bad Step 1 hop, the adaptive query/cardinality telescope, the combined cache/table Step 2 semantics, the distinct-output collision branch against the exact-match branch, the reservoir/embed-index reductions, the explicit failure-mass conditions, and the append-forgery-query compiler. I also instantiated both `_of_queryBound` endpoints against a temporary consumer to check they are usable as stated, rebuilt the GPV modules and `Examples/GPVNonVacuity.lean`, and probed axioms on the Step 1 hop, both headline bounds, append-query preservation, min-entropy, and sampler transport — standard three throughout.

The append-query compiler is the part I would most want advertised. It discharges `ForgesQueriedPoint` for an arbitrary adversary at the cost of one hash query, which is what stops that hypothesis from turning into a silent well-formedness convention at each scheme instantiation. I would make `euf_cma_split_bound_of_queryBound` / `euf_cma_collision_bound_of_queryBound` the documented public endpoints and let the raw-hypothesis versions read as internal.

One doc correction. The Step-1 section prose says `gpv_tvDist_real_programmed_le_collisionBound` consumes the front-tape coupling and that "the front-tape coupling is invoked on the Step-1 path". The proof does not: it goes through `gpv_tvDist_orig_run_le_probEvent_flag` and `gpv_orig_flag_le_collisionBound`, the original-run/inline-salt-flag route, as its own inline comment says. The tape factorization is still worth keeping as reusable infrastructure, but a reader auditing Step 1 from the section header will follow the wrong chain.

The joint toy witness is correctly scoped as consistency-only. Whenever a natural consumer shows up, one quantitative min-entropy instantiation would be worth adding — not something I would hold this PR for.

quangvdao (approved):
Reviewed at `b74efcb96ebc2625740f86f39baf1da6d1e4aee2`.

No blocking correctness finding. I checked the load-bearing split: the real→programmed hop uses the inline-salt original-run bad-event route; the programmed game splits a successful forgery into a distinct-preimage collision and an exact programmed-preimage replay; the latter is priced by the sampler guessing probability rather than erased; and `appendForgeQuery` gives the unrestricted `_of_queryBound` endpoints at one additional hash query. The failure, regularity, min-entropy, and sampler-transport assumptions remain explicit, and the consistency witness is correctly scoped as non-quantitative. No new trust debt appears, and exact-head CI is green across the full build and axiom checks.

Two P3 documentation/API notes remain non-blocking:

1. The section beginning “Step 1 and the wiring to the headline bounds” in `GameIdentification.lean` still says the front-tape coupling is invoked on the live Step-1 path. The actual theorem at this head calls `gpv_tvDist_orig_run_le_probEvent_flag` followed by `gpv_orig_flag_le_collisionBound`, as its own theorem docstring correctly explains. Please fix that stale section prose in a follow-up.
2. `euf_cma_split_bound_of_queryBound` and `euf_cma_collision_bound_of_queryBound` are the consumer-facing endpoints because they eliminate `ForgesQueriedPoint`; advertise them in the umbrella/module documentation so downstream users do not stop at the raw-hypothesis variants.

Approved for merge; neither note changes the theorem statements or proof chain.

alik-eth:
@quangvdao — this PR does not build on current `main` (Lean 4.33.0), and neither do #466, #507 or #508. The `7/7 pass` badges on all four are 4.32.2 results, as you noted on #514. Merging `main` in is clean on every one of them — no textual conflicts — but the proofs then break.

I have migrated most of it and hit one blocker that I think is yours to call rather than mine. Partial work is on `gpv188-generic-r433` (`3dc6349b`), branched from a clean merge of `main` into this PR's head, so nothing here is disturbed.

## Root cause

`OracleSpec` addition is `Sum.elim`, so `((a + b) + c).Range (Sum.inl (Sum.inr t))` reduces to `b.Range t` only through unfolding that 4.33 no longer performs at the transparency tactics use. Any proof that unfolds a composed handler ends up with a goal that is **ill-typed at implicit transparency**, after which tactics fail to match patterns that are plainly visible in the printed goal — `rw` reports "did not find the pattern" against a hypothesis that displays exactly it.

The symptoms look unrelated but are all this: `unsolved goals`, `tauto failed`, `rewrite failed`, `application type mismatch`. `main` itself builds fine, so the core layer is healthy; it is specifically proofs that unfold through the sum.

## Fixed: 14 across three modules

| shape | n | fix |
| --- | --- | --- |
| composed handler will not unfold | 1 | `change` to the explicit form, then `simp only [QueryImpl.simulateQ_add_liftM_query_right]` — the idiom `gpvRealImpl_run_sign` already uses |
| `id` survives under `do` binders | 7 | append `rfl`; `id_eq` no longer clears it |
| `tauto` cannot refute a tuple's `false = true` | 4 | `obtain` then `absurd (congrArg (·.2.2) h) (by simp)` — `simp` cannot destructure the pair, because the pair's type carries the unreduced `Range` |
| `mem_support_bind_iff` applied to a map | 1 | `mem_support_map_peel` first |

`GameRuns`, `FlagHandlers` and `CombinedHandler` are clean.

One ordering constraint is worth knowing, since it cost me the most time: in `FlagHandlers`' random-oracle branch the map must be flattened **before** rewriting with `gpvRealImpl_run_read`. That lemma's sides are definitionally equal but not syntactically so — LHS type `Range (Sum.inl (Sum.inr mc))`, RHS `Range mc` — so rewriting first leaves a goal that is ill-typed at implicit transparency and everything downstream fails. Reordering two lines fixed what no amount of simp-set tuning would.

## The blocker, and why I stopped

`GameIdentification.lean:179`. Same root cause, but in a **type position** rather than a tactic-reachable one:

```
fun msg ↦ gpvRealImpl psf hr M Salt pk sk (Sum.inr msg)
  has type      (msg : (ofFn …).Domain) → StateT … (((unifSpec + …) + ofFn …).Range (Sum.inr msg))
  but expected  QueryImpl (ofFn fun x ↦ Salt × Domain) (StateT …)
```

No tactic can rewrite the expected type of a lambda being passed to `appendInputLog`. Fixing it means phrasing the handler so the composed `Range` does not leak into a `QueryImpl` position — annotating at the `QueryImpl` type, or adjusting `gpvRealImpl`'s signature. That is a design change to your architecture with other callers, so I would rather you make the call than have me guess at it.

## Two hypotheses I tested and refuted

So nobody repeats them:

- **`Sum.elim` being semireducible is not the cause.** Tested directly with `attribute [local implicit_reducible] Sum.elim`; the attribute took and the errors were unchanged.
- **Type-ascribing the right-hand side does not help.** Elaboration discards the ascription; error unchanged.

## Scope

Each module reveals its own breakages only once the previous one is fixed, since the build halts at the first failure — so the counts above are what surfaced, not a final total. #466 shares the GPV files fixed here and should come out substantially cheaper; #507 and #508 fail instead at `GhostBodies/GhostLayer.lean:326` on a `subst`, which I have not yet diagnosed.

dtumad:
Deep semantic audit across the open PR set (models, non-vacuity, faithfulness to sources; no style content). **Verdict: no blocking semantic issues found — this is the PR the rest of the lattice family should land on top of.**

What I attacked and could not break, for the record:

- **Degenerate samplers are genuinely excluded.** `PreimageSampleableFunction.Regularity` is an *exact* joint-distribution equality whose first marginal is uniform on `Range`, so a constant (or otherwise collapsed) trapdoor sampler cannot satisfy it for `|Range| > 1`; combined with `CorrectAt` the abstraction has no trivial models.
- **The exact-match branch is priced, not hidden.** `programmedPreimageAdvantage` with `euf_cma_split_bound`, and the explicit statement that `εpp = 0` is unsatisfiable, is exactly the honest treatment — this is a real discovery relative to the collision-only endpoint currently sorried on `main`.
- **`ForgesQueriedPoint` is not a silent restriction**: `appendForgeQuery` discharges it unconditionally at one extra hash query with an *exact* advantage equality via cache-hit absorption.
- The collision reduction (`CollisionTelescope`) never touches `sk`, so it is a legitimate collision-game adversary.

Remaining before merge (all known, none blocking):

1. Rebase to the 4.33.1 / measure-first main. One thing to re-verify by hand on rebase rather than assume textual: the load-bearing `𝒟[…] = 𝒟[…]` equalities in `Regularity`/`hReg` — on current main `𝒟[·]` denotes a `Measure` and `𝒮[·]` is the SPMF; on these finite carriers they agree, but the hypotheses should be restated in whichever form the downstream accumulation lemmas actually consume.
2. The two P3 doc notes from the 08-23 review (stale front-tape-coupling prose in `GameIdentification.lean`; advertise the `_of_queryBound` endpoints) are still unaddressed at head `b74efcb`.
3. Follow-up paper check I'd still like recorded in the module docstring: the salt-collision constant `(qS+qH)²/(2|Salt|)` vs GPV08 Prop 6.2's tighter `Q_s·(C_s+Q_H)/2^k` — fine to keep the looser constant, but cite which one is being proved.

Landing order for the family, so everyone can rebase deliberately: **#478 → #514 → #476 → #466 → #507 → #508 → #471 (split per its review)**. #466 and #471 both currently carry snapshots of this PR's files; after this merges they should rebase down to their true deltas.

alik-eth:
All three remaining items are addressed. Head is `5a85bd36`.

**1. Rebase to the 4.33.1 / measure-first main** — done at `384fe731` (merged `main` through #582; 8/8 CI checks green). The `Regularity` / `hReg` equalities are restated as `𝒮[·]` (`evalSPMF`), not `𝒟[·]`, because the SPMF form is what the accumulation actually consumes: `AdaptiveFactorizesSignRunF` takes `realRun progRun : SPMF α` literally, and `signRunF_tvDist_le_collisionBound` is an `SPMF.tvDist` bound. The `Measure` reading is available by `rfl` through `evalDist_eq_evalSPMF_toMeasure` for anyone who needs it downstream. The GPV files also had to be added to `scripts/pmf_boundary_baseline.tsv` (they do not exist on `main`, so their ceiling was 0); no existing ceiling was raised.

**2. The two P3 doc notes** — done at `5a85bd36`. The Step-1 section header in `GameIdentification.lean` now describes the route the proof takes (`gpv_tvDist_orig_run_le_probEvent_flag` then `gpv_orig_flag_le_collisionBound`, salts drawn inline). Fixing it turned up the actual cause: `gpv_tvDist_tape_runs_le_collisionBound` has never existed in this branch's history. It was the planned coupling headline of the tape route, and 15 docstrings across `GameRuns`, `Factorization`, `TapeFactorization`, and `GameIdentification` cited it as if proved; three more names in the same family were also phantoms (`gpv_tvDist_tape_run_le_probEvent_flag`, `signRunF_flag_le_saltSeq`, `signRunF_tvDist_le_flag`). Every one is rewritten to describe what is established: the tape route is ingredients only (the per-step agreements and the salt-tape birthday infrastructure), its coupling headline is not proved, and the Step-1 bound comes from the inline-salt route. The `_of_queryBound` pair is now documented as the public endpoints in `AppendQuery.lean`, the `Security.lean` module docstring, both raw-headline docstrings, and a new *Public endpoints* section of the umbrella; the `ForgesQueriedPoint`-hypothesis forms are described as the primitives the compiler chains through.

**3. Which salt-collision constant is proved** — the closed birthday form `(qSign + qHash)² / (2 · |Salt|)` (`collisionBound`) is what the headlines state. The telescope actually establishes the exact union sum `∑_{j < qSign} (j + qHash) / |Salt|` (`sum_range_div_card_le_collisionBound`), and the closed form is looser than that sum by the `qHash² / 2` term. On attribution: I checked ePrint 2007/432 §6.2, and GPV08's Proposition 6.2 proof only sketches the same-message salt reuse term `Q_sign² / 2^k`; the `Q_s · (C_s + Q_H) / 2^k` form is the [FGdG+25] Theorem 1 term, as the umbrella docstring already recorded. The `collisionBound` docstring had mis-attributed the closed form to GPV08 Prop 6.2, the umbrella called it "slightly looser" (it is looser by `qHash² / 2`, which dominates when `qHash ≫ qSign`), and the proof-decomposition overview still said `q_S² / (2 · |Salt|)`. All three are corrected, and the `euf_cma_collision_bound` docstring now names the constant it proves and points at the `collisionBound` docstring for the comparison.

CI on `5a85bd36` is running; the change is docstring-only.

alik-eth:
@dtumad I merged current main into this branch (3a544061, plus 46d0d013) and have a question on the retiring probability API before going further.

**What the merge did**
- Ported to v4.34 and the measure-valued runtime. `runtime` is `ProbCompRuntime.rom`, and `runtime_evalDist_singleton` identifies its point masses with the old game observation, so `unforgeableAdvantage` is read in exactly one place.
- Following #707, the headline bounds now name their reductions: `reduction … adv domainSample` and `programmedPreimageReduction … adv domainSample qSign qHash`, with the domain sampler explicit. The `_of_queryBound` forms use the reductions of `appendForgeQuery … adv`. There are no `∃ red` conclusions left.
- Split `GPVHashAndSign/Security.lean` into `TrapProjection`, `Reservoir` and `Security` to stay under the 1500-line limit.
- Fixed every `unusedArguments` finding, mostly instance binders your refactors made unnecessary.

Build, style lint, the boundary checks and `axiomsweep --check` pass. The headlines and both reductions are no longer sorry-tainted.

**The question**
The Lint job fails only on `usesRetiredProbability`. 138 declarations here and 34 more on #466 state or prove things through `probOutput`/`evalSPMF`. `scripts/lint.py` would accept them as ledger additions, but AGENTS.md says the ledger should only shrink, so I haven't added them.

Converting them isn't self-contained. The exact-match branch rests on `OracleComp.DeferredSampling` (`evalSPMF_bind_comm`, `evalSPMF_step_commute_tape`, `evalSPMF_bind_const_neverFails`, `Factorizes`), and those lemmas are themselves grandfathered in the ledger. So are the relational epsilon lemmas, the `withStateOracle` SPMF semantics and the PMF total-variation lemmas the coupling uses. From the conversion roadmap, the analogous reader direct-coupling family is still waiting on those prerequisites.

How do you want in-review work like this handled?
1. Enrol these declarations in the ledger for now, and migrate them once deferred sampling and the relational coupling layer have measure owners.
2. Convert first, starting with `DeferredSampling`. That overlaps your roadmap, so I'd want to coordinate on its shape.
3. Something else you have in mind.

The same answer applies to #466. I expect the Fiat-Shamir stack (#507/#508/#639) to be in the same position.

alik-eth:
Confirmed for the Fiat-Shamir side: after merging main, #507 has 100 declarations flagged by `usesRetiredProbability`, #508 adds 14, and #639 adds none. The coupling engine there has the same dependency on the grandfathered deferred-sampling layer. None of these branches enrols anything in `scripts/nolints.json`; everything else passes on all five.

PR #479, opened 2026-07-12
fix(MLDSA): make Primitives.Laws inhabitable; bank the FIPS-shaped concrete witness

alik-eth:
Extracts the ML-DSA `Primitives.Laws` repair and its concrete witness out of #467, per @quangvdao's review there, so it can land independently of the EUF-CMA development. #467 will be rebased on top once this merges.

## The problem

On `main`, `Primitives.Laws` is **uninhabitable** — three fields are jointly (and individually) unsatisfiable — so every Laws-conditional MLDSA statement (`nma_security`, completeness, HVZK) is vacuously true. `#print axioms` cannot see this: the theorems are axiom-clean, they just assert nothing at any realizable instance.

- `expandS_honest_sampling` asserts the deterministic image of `2^256` seeds is *uniform* on `RqVec l × RqVec k` — false by cardinality (`2^256 < |Rq|`).
- `w1Encode_injective` demands injectivity on the full `Vector High p.k` carrier — false for the truncating packer.
- `expandMask_bound` uses `γ₁ − 1` where FIPS 204's `ExpandMask` coefficients lie in `[-γ₁ + 1, γ₁]` — spec drift.

## The fix

- `expandS_honest_sampling` is dropped. The MLWE key-swap hop instead carries an explicit total-variation slack hypothesis `honestSamplingSlack idealGap`, threaded through `nma_keyswap_hop` and `nma_security` as `+ idealGap`. This is honest: trivially satisfiable at `idealGap = 1`, quantitative under a ROM reading of the XOF.
- `w1Encode_injective` weakens to `Set.InjOn` on vectors of `highBits` representatives — exactly what commitment binding uses (`w₁` always arises as such a vector).
- `expandMask_bound` corrected to `γ₁`.

## The witness (the non-vacuity certificate)

`Concrete/Laws.lean` + `Concrete/NonVacuity.lean` prove `MLDSA.mldsa_laws_inhabited`: the FIPS-shaped `seedRevealingPrims` (= `concretePrimitives` with the `ρ`-component of `expandSeed` exposed, a logical-consistency witness only) satisfies **all** Laws fields for any approved parameter set. `Primitives.Laws` is now inhabitable, so the conditional security statements have content.

## Verification

- `lake build` of the full CI library set on `leanprover/lean4:v4.31.0`: green, zero non-sorry warnings.
- `#print axioms MLDSA.nma_security`: `[propext, Classical.choice, Quot.sound]`.
- `#print axioms MLDSA.mldsa_laws_inhabited` / `seedRevealingPrims_laws`: the same plus the one pre-existing, disclosed `native_decide` axiom for the NTT matrix inversion identity (`invNTTMatrix_nttMatrix_entry`).

The diff outside `Primitives.lean` + `Concrete/` is the key-swap adaptation in `SecurityNMA.lean` and supporting norm lemmas in `Ring/`.


dtumad (changes requested):
The original defect is genuine: the exact expandS_honest_sampling field cannot be inhabited by deterministic ExpandSeed/ExpandS over the stated full-ring uniform distributions. This PR should not merge in its current form, however.

Blocking mathematical issue: mldsaMLWE samples both the secret and error uniformly over the entire RqVec carriers. In a finite additive group, adding an independent uniform error makes the real output uniform, so this MLWE advantage is identically zero and does not model short-secret/short-error MLWE. The new honestSamplingSlack compares deterministic eta-bounded ExpandS outputs with independent full-ring-uniform vectors. A random-oracle treatment of ExpandS can justify a short-output replacement distribution, not uniformity over the whole ring. The universal witness at idealGap = 1 only proves satisfiability; it makes the resulting security bound non-informative.

Blocking integration issue: this branch predates the Extern split. On current main, the SHAKE implementation lives under Extern/MLDSA, while proof libraries may not import Extern. The merge currently leaves the new Laws file importing the now-absent LatticeCrypto.MLDSA.Concrete.Instance and attempts to reintroduce old concrete umbrella imports. SHAKE-dependent laws and the nonvacuity witness need to live under Extern; purely algebraic/range lemmas may remain in LatticeCrypto.

Recommended remediation: retain the sound primitive corrections here, notably the gamma1 ExpandMask bound and Set.InjOn w1Encode law, but replace the full-ring slack/security stopgap with the short-box construction already developed on #467: sampleShortVec, mldsaMLWEShort, the exact short-key swap, and an explicit ExpandS replacement assumption. Then restack #467 on the remediated branch and rebuild against current main.

dtumad (changes requested):
I re-reviewed the revised head 7615dac0 against current main and built LatticeCrypto plus Extern on the clean merged tree. The prior two blockers are substantively fixed: the key swap now uses short-box secret/error sampling rather than full-ring uniform error, and SHAKE-dependent law proofs/non-vacuity live under Extern while algebraic range facts stay in LatticeCrypto. The gamma1 ExpandMask bound and Set.InjOn commitment-encoding law also match the ranges actually consumed.

There is one remaining mathematical blocker in the SelfTargetMSIS leg. SelfTargetMSIS.Problem.isValid receives only challenge, target, hash output, and response. The experiment looks up hashInput in the RO cache, but never passes that preimage to isValid. Consequently mldsaSTMSISShort.isValid recomputes a commitment w-prime from pk, c-tilde, z, and h and verifies against that same recomputed value; it never requires the commitment component of the queried pair (msg, w) to equal w-prime. The proof even closes the relevant equality as X = X. This establishes RO consistency for some queried input, but not the self-target binding to the commitment that was hashed. The resulting problem is much easier than the literature relation and can admit the zero-response/zero-hint style strategy after querying an unrelated commitment, subject only to the verifier norm gates.

Please make the hashed preimage available to validity, for example:

  isValid : Challenge → Target → HashInput → HashOutput → Response → Bool

and have the experiment call isValid with hashInput. For ML-DSA, require hashInput.2 = recoveredCommitment in addition to the existing verifier checks, then update the read-back proof. The current #467 branch contains a candidate version of precisely this repair and useful algebraic characterization lemmas, but it should still be reviewed as code rather than assumed correct merely because it exists there. A focused characterization such as isValid at ExpandA(pk.rho) iff the norm gates and hashInput.2 = recoveredCommitment would make this regression hard to reintroduce.

Because the headline NMA inequalities are currently bounded by an incorrectly weakened, effectively trivial SelfTargetMSIS problem, this head is not safe to merge despite green CI. Architecturally, the cleanest order would be to extract the generic SelfTargetMSIS API/binding repair first, then restack this PR (or narrow this PR back to the advertised Laws/Extern work and move the NMA development to #467). I am leaving that split/restack choice to the author because it changes semantic organization rather than being a mechanical edit.

alik-eth:
@dtumad The generic SelfTargetMSIS binding repair you asked to extract first is now #503, off current main:

- `SelfTargetMSIS.Problem.isValid` gains the `HashInput` argument and `experiment` threads the cache-consistent preimage into it;
- `mldsaSTMSIS` additionally requires `hashInput.2 = w'`, closing the trivial `X = X` solution — the extraction proof discharges the binding for free since the forger's returned `(msg, w')` *is* the queried preimage;
- `mldsaSTMSIS_isValid_eq_true_iff` states acceptance as verifier-acceptance ∧ the binding conjunct, so the self-target requirement stays visible and hard to drop.

It's deliberately narrow — just the binding — and builds green on 4.32 with the two affected declarations kernel-clean. The short-MLWE model, the standard-STMSIS normal-form reduction, and the CMA-to-NMA engine remain the separate later pieces per your proposed order. Once #503 lands, this branch can rebase onto the fixed API.

alik-eth:
On the split-vs-restack choice you left to the author: I'll take the cleaner option you outlined — **narrow #479 back to the advertised Laws/Extern work** (the `γ₁` ExpandMask bound, the `Set.InjOn` `w1Encode` law, the short-box key-swap you already fixed here, and the SHAKE-law/Extern reorganization), and **move the NMA/SelfTargetMSIS development to #467**, restacked on the #503 binding repair. That keeps this PR's diff matching its title and lets the STMSIS soundness land through the small #503 → #467 path rather than being re-litigated here.

alik-eth:
Narrowed to the Laws/Extern scope as discussed — force-pushed onto your head, so your Extern reorganization and the primitive corrections (γ₁ ExpandMask bound, Set.InjOn w1Encode, dropping the false `expandS_honest_sampling` field) are preserved intact.

The change: `SecurityNMA.lean` is reset to `main` with only the two theorems coupled to the dropped field removed — `nma_keyswap_hop` and its sole consumer `nma_security`. Nothing outside `SecurityNMA.lean` depended on the NMA chain in code (only three docstrings, made generic). The whole short-model NMA development (`sampleShortVec` … `nma_security_short`/`_fips`, `mldsaSTMSISShort`, the MLWE short problems, the extraction) is gone from here and consolidated on #467, restacked on the #503 binding repair.

Diff vs main is now Laws/Extern/Concrete/Ring/witness files only; full CI set green on 4.32 and `mldsa_laws_inhabited` kernel-clean (`[propext, Classical.choice, Quot.sound]` + the pre-existing disclosed NTT `native_decide`). Once #503 lands, this needs only a routine rebase — it currently carries main's pre-binding `mldsaSTMSIS`, which is orthogonal to the deletions here.

dtumad (approved):
Re-review after the Laws/Extern narrowing and merge with current main: the original defects are genuinely repaired. The `ExpandMask` law now matches FIPS 204’s `[-γ₁ + 1, γ₁]` range, `w1Encode` injectivity is restricted to the actual `HighBits` image consumed by the proof, and the impossible deterministic/full-ring `expandS_honest_sampling` equality is gone. The concrete encoding, rounding, sampler, convolution-norm, and inhabitance proofs are consistent with those corrected statements; the `seedRevealingPrims` witness is appropriately documented as a logical consistency witness, not a faithful or secure implementation.

I resolved the two merge conflicts mechanically: retained the short-secret key-swap development already on main, removed only the obsolete full-ring key-swap/headline theorem coupled to the deleted law, and retained `validKeyPairShort`. I also corrected the remaining comments that described the old full-ring `mldsaMLWE` branch as if it reproduced seed-derived key generation, and clarified the inherited `native_decide` trust surface.

Validation on the exact merged tree:
- `lake build LatticeCrypto Extern` passes.
- `lake build mldsa_test` passes with the pinned native backends.
- `mldsa_test`: 128 passed, 0 failed, including ML-DSA-44/65/87, ACVP vectors, sampler bounds, byte-for-byte key/signature comparison, and cross-implementation verification.

Approved; safe to merge once the newly triggered CI checks are green.

PR #484, opened 2026-07-13
docs(agents): bank elaboration, vacuity-certificate, and restacking gotchas

alik-eth:
Five field-tested additions to `docs/agents/gotchas.md`, distilled from recent proof campaigns and written as timeless guidance:

- **Type System #8** — structure values closing over `Fintype.ofFinite`/`SampleableType.ofFintype` are whnf-hostile: the mechanism, the `where`-syntax symptom, and the fixes (anonymous constructors; abstract-value + pinning-equality statement architecture; explicit `Classical.propDecidable`).
- **Proof Patterns #14** — hypothesis satisfiability is a proof obligation: vacuous conditional theorems are invisible to `#print axioms`; the two common failure classes (relation-pinning pairs vs `gen_sound`; deterministic small-seed-image cardinality), and the rule of shipping a kernel-checked inhabitance certificate in the same commit.
- **Build/Tooling #21** — warm-starting worktrees from a built donor's `.lake/build` (placed as a cluster with the existing cache entry).
- **Build/Tooling #22** — `lake env lean`/LSP don't check olean staleness: phantom whnf timeouts and unknown-identifier errors that mimic real elaboration pathologies (including a section-dependence illusion via shared instance fvars); rule: no-op `lake build` before diagnosing, trust only symptoms that reproduce after.
- **Build/Tooling #29** — restacking with `git rebase --onto <new-base> <old-base-tip>` after folding commits into a base, and the tree byte-identity gate before pushing.

Docs-only; entries renumbered into a clean sequence (verified nothing references gotchas by number).


dtumad (approved):
The new gotchas are useful and concrete. I updated the waiting branch for the current repository state: Lean/Mathlib v4.32.0, HashSig in the allowed top-level umbrellas, and less categorical wording for the whnf failure mode. scripts/check-agent-docs.py and git diff --check pass. Approving the resulting one-file diff; merge should remain gated on fresh CI.

dtumad (approved):
Re-reviewed the entire final diff, not only the version refresh. I verified the ofFintype mechanism against the current implementation, the relation-pinning claim against GenerableRelation.gen_sound, the root/import lists against the repository and update script, the rebase/range-diff syntax against Git, and the cache-copy shape in a scratch directory. I corrected overbroad claims, made the satisfiability requirement apply to nontrivial bundles, made the tree-identity restack check conditional on an equivalent folded base, guarded the build-cache copy against an existing destination, and qualified stale-olean symptoms. scripts/check-agent-docs.py and git diff --check pass. The resulting advice is useful and accurate; approving head 76c0566d, with merge still gated on fresh CI.

alik-eth:
Thanks — and thanks for updating the entries for the current repo state (4.32, HashSig umbrella, the softened whnf wording). Good to go on your say-so.

PR #487, opened 2026-07-14
prove(polytime): close nontriviality certificate

eliasjudin:
Aristotle-generated closure of the three isolated nontriviality-certificate gaps, with the accepted proof artifacts and local validation recorded in the branch.

The sole commit is explicitly co-authored with Aristotle (Harmonic).

The contribution has now been incorporated into the current dependency stack and the VCVio facade:

- [PolyFun #178](https://github.com/Verified-zkEVM/PolyFun/pull/178) carries the state normalization and finite covering proofs in `ToCslib/Computability/SingleTape/Counting.lean` and the fixed-bit append machine in `ToCslib/Computability/SingleTape/Snoc.lean`. The source headers retain Elias Judin's credit.
- [PolyFun #179](https://github.com/Verified-zkEVM/PolyFun/pull/179) provides the pure-program nontriviality endpoint `exists_not_isPPolyBy_pure` using the current initialization/head-observation bridge. That bridge supersedes the old iteration argument; the old proof is not claimed to survive unchanged.
- [VCVio #576](https://github.com/Verified-zkEVM/VCVio/pull/576) provides the VCVio-facing `exists_not_isNonuniformPPTBy_pure` endpoint on the current pinned PolyFun. Its repair passed the complete build, lint, tests, and axiom checks, including a concrete fair-coin machine certificate.

The PolyFun PR descriptions now record the original #487 / [#500](https://github.com/Verified-zkEVM/VCVio/pull/500) provenance and original co-author trailer:

```text
Co-authored-by: Aristotle (Harmonic) <aristotle-harmonic@harmonic.fun>
```

This PR is retained as the original proof and review record. Closure follows the [author's consent](https://github.com/Verified-zkEVM/VCVio/pull/487#issuecomment-5523923740) after the remaining VCVio endpoint landed; the original branch history is preserved.

PR #481, opened 2026-07-12
feat(polytime): non-vacuous PPT/IsPolyTime over PolyFun (+ Interaction/UC migration)

dtumad:
Draft — lands a **non-vacuous, TM-grounded `OracleComp.IsPolyTime`** (probabilistic polynomial time) stated as a thin layer over PolyFun's `PointedMachine`. This is a corrected re-do of #460, whose per-step cost witnesses were dischargeable for free (via a per-`n` lookup table), collapsing the notion onto `PolyQueries`; here the per-step Turing-machine witnesses carry a single polynomial bounding **both running time and description size (state count) uniformly across `n`**, together with pinned canonical boundary encodings — which is what makes the class non-vacuous.

## Depends on
- **Verified-zkEVM/PolyFun#29** (`PointedMachine.Implements`). PolyFun is pinned at that branch head `5c3a070`; re-pin to the merge commit once it lands.
- Adds a new dependency **`cslib @ v4.31.0`** (single-tape Turing machines, for the cost grounding).

## What's here
- **Coinductive machine layer** over PolyFun: `OracleMachine := PointedMachine`, `Implements` (definitionally PolyFun-native, with `runK_eq`/`of_runK_eq` recovering the ∀-handler reading), `runK`/`runD`/`toComp`, `seqComp` (`VCVio/OracleComp/Coinductive/`).
- **cslib-grounded cost model** (`ToMathlib/Computability/{Encoding,CslibPolyTime,PolyTimeTM,BitEncoding}`) and the join `IsPolyTime` + `BoundaryData` + `MachineAdversary` + `PolyTimeWitness`.
- **Security-game wrapper** `SecurityGame.secureAgainstPolyTime` (`Asymptotics/PolyTime`).
- **Non-vacuity evidence**: certified `isPolyTime_coin` (a single coin flip is PPT, hypothesis-free, axiom-clean) and a **staged diagonalization skeleton** for `exists_not_isPolyTime_pure` — the counting/cardinality/growth spine (`2^(2^n)` functions vs. sub-doubly-exponentially many small-machine behaviours) is proven and axiom-clean in `ToMathlib/Computability/MachineCounting.lean`; both headline non-triviality theorems are assembled with no literal `sorry` in body.
- **Interaction/UC migration** to PolyFun's reshaped `ProcessOver` (state type promoted to an explicit parameter).

## Known gaps (3 named, documented `sorry`s — the isolated hard core)
[...]

PR #500, opened 2026-07-26
feat(polytime): non-vacuous PPT/IsPolyTime over the DynComputation carrier, sorry-free

dtumad:
## Motivation

Re-extraction of the Turing-machine-grounded polynomial-time layer from the closed reference
draft #481, onto current main: Lean v4.32, current Mathlib/cslib, and the merged
`DynComputation`-based machine layer (#498/#499) in place of the retired `PointedMachine`
API. Includes the closure of all three non-triviality `sorry`s contributed in #487.

The model is unchanged from #481 (and from the design agreed in the #460 review): non-uniform
P/poly relative to pinned canonical boundary encodings, with a single polynomial bounding both
per-step running time and machine description size uniformly across the security parameter —
the description-size bound is what closes the truth-table/unbounded-advice escape. The model
and its statement-site discipline are now documented in `docs/agents/polytime-model.md`.

## What moved under the port

**`ToMathlib/Computability/` + `ToMathlib/Data/BitVec.lean`** — the TM toolkit on current
Mathlib/cslib. Encodings are raw injective `List Bool` families (`BitEncFam` / `StrEncFam`)
from the start — no intermediate alphabet layer. cslib's restructure
(`SingleTapeTuring.Basic` → `Turing/SingleTape/Deterministic`, namespace
`Cslib.Turing.SingleTapeTM`) is a pure rename — zero proof changes. Per the repo decision
(recorded in the new design doc), nothing here is aimed at upstream Mathlib/cslib PRs; it
stays in `ToMathlib/`.

**`MachineCounting.lean` and the non-triviality closure (from #487).** The counting cruxes
(`exists_tmTable_of_card_le`, the realizable-covering surjection) and the B-factor argument
port with only signature-level adjustments (`ReflTransGen.lift` is now a `≤`-statement; the
trajectory argument reads through the new run lemma below). Credit: Elias Judin, with the
proofs generated by Aristotle (Harmonic); co-author trailers on the porting commits and
co-authorship in the two files' headers.

**`Coinductive/PolyTime.lean`** — `MachineAdversary` / `PolyTimeWitness` /
`OracleComp.IsPolyTime` over the `DynComputation` carrier. The TM-facing step maps (`expose`,
`updateFlat`, `output`, `stepD`) are total functions **derived** from the machine's one-step
`view`, spelled with `Sum` combinators so they transport definitionally between machines
sharing the same dynamics. Three fields of the old bundle are gone as theorems, not
hypotheses: readout stability (returns are absorbing by construction), steadiness
(`PolyTimeWitness.resolvesIn`, handler-free via `ResolvesIn`), and the syntactic query bound
(`PolyTimeWitness.queryBound`, from the bounded half of `ImplementsWithin` via
`implementsWithin_iff_implements_and_bound` — `IsTotalQueryBound` is definitionally
`IsTotalRollBound`).

**`Coinductive/PolyTimeClosure.lean`** — closure under input precomposition and output maps.
Much thinner: replacing `init` is a real `OracleMachine.setInit` whose runs are unchanged
(`unroll_setInit`), output post-composition is upstream `mapResult` (`unroll_mapResult`), and
the precomposition implements-proofs reduce to input reindexing.

**`Coinductive/CoinFold.lean`** — the reusable bounded-coin-fold combinator. The machine is
built by a new generic `OracleMachine.ofStep` (one-step transition + initialization) and
implements the fold program by one direct `unroll` induction.

**`Asymptotics/PolyTime.lean`** — `SecurityGame.secureAgainstPolyTime` and the
per-query-loss former, consuming the derived `PolyTimeWitness.queryBound`.

## Semantic-audit pass (post-extraction)

A full audit of the layer (definitions vs. the #460 review record, cslib upstream direction,
axiom footprint) produced the following fixes, all in this PR:

- **`detTotalTime` charges the final readout** — the run's answer is
  `output (stepD^[steps] s)`, so the readout at the budget state is a real evaluation; the
  accounting and both bounding theorems gained the matching term. This closes the last open
  #460 accounting bullet (the off-by-one).
- **`PolyTimeWitness.queryBound` is a derived theorem, not a field** (see above); every
  witness construction site shed the redundant obligation.
- **`PolyTimeComputable.size` is pinned to the `Bool` alphabet** — a bare state count only
  measures description size over a fixed alphabet; the restriction is now in the type rather
  than in prose.
- **The unused `PackedEncoding`/`boolify` layer is dropped** (`Encoding.lean`, 337 lines of
  dead code that two docstrings mispresented as the model's encoding source).
- `exists_tmTable_of_card_le` generalized to any `SingleTapeTM Bool`; stale docstrings
  refreshed throughout (retired names, sorry-era prose, dangling module references);
  `IsPolyTime` now states its non-uniformity (P/poly) explicitly.

## Status of the #460 checklist

- advice/description-size bound: **done** (`descBound`; non-triviality certified below);
- termination and state-size bounds over all answer paths: **done** (`ResolvesIn`
  quantifies over every typed answer path; `StrEncFam.len_le` over all states);
- input/message-size assumptions: **done** (`BitEncFam.widBound` is intrinsic —
  `detTotalTime_le` and `exists_polynomial_detTotalTime_le` are hypothesis-free);
- final-readout accounting: **done** (this PR's audit pass).

Still deliberately out of scope (documented in `docs/agents/polytime-model.md`): the
end-to-end compiled single-machine theorem (#460 item 7's "second tier"), `bind` closure,
and a uniform variant. `secureAgainstPolyTime` accordingly remains unused by any concrete
security statement this cycle.

## Validation

- `lake build` green; `scripts/check-extern-isolation.sh`, `scripts/check-interop-isolation.sh`,
  `scripts/check-agent-docs.py`
- no literal `sorry` anywhere in the added files; headline theorems
  (`exists_not_isPolyTime_pure`, `exists_not_implements_pure_of_steps_eq_zero`,
  `secureAgainstPolyTime_of_advantage_le_mul_totalQueries`, `PolyTimeWitness.queryBound`,
  `exists_polynomial_detTotalTime_le`) all check at exactly
  `[propext, Classical.choice, Quot.sound]`
- the `ToMathlib` layer builds independently of the machine layer

## On merge

- Close #487 (its content is fully ported here, with attribution carried).


ToMathlib/Computability/BitEncoding.lean
```diff
@@ -0,0 +1,455 @@
...
+
+* `Computability.StrEncFam` — a security-parameter-indexed family of injective raw
+  `List Bool` encodings with a polynomial length bound. This is the *variable-width*
+  notion, the representation freedom left to a machine's internal state.
+* `Computability.BitEncFam` — the *fixed-width* refinement: at each parameter every
+  value encodes to exactly `wid n` bits, with `wid` polynomially bounded. This is the
+  canonical *boundary* representation for inputs, outputs, and oracle interfaces. The
+  polynomial width bound is the formal content of the Katz–Lindell `1^n` convention:
+  all game values at parameter `n` have `poly(n)`-length representations, so
+  "polynomial in `n`" and "polynomial in the input length" agree.
```
dtumad, line 35: A polynomial upper bound on widths gives only one direction. `1^n` works by forcing input length `≥ n`; `widBound` does not force the boundary to carry `n` at all. `Unit` at every `n` has width `0`, so an `n²`-time machine is poly-in-`n` but not poly-in-input-length.

VCVio/OracleComp/Coinductive/PolyTime.lean
```diff
@@ -0,0 +1,608 @@
...
+  same reason. Statement-site discipline: `bd` is always an explicit pinned parameter
+  of a security definition, never existentially quantified and never adversary-chosen
+  (the sole documented exemption is a multi-phase adversary's *own* cross-phase state,
+  which its phases share — every bit cached there was produced by a witnessed machine
+  from canonical inputs and answers). The registry of canonical constructors
+  (`BitEncFam.const/bitVec/pair/option`) is deliberately tiny and structural; theorems
+  mean "secure against P/poly relative to the standard representation".
+* **The `1^n` convention, formalized.** Katz–Lindell define PPT as polynomial in the
+  *input length* and reconcile it with "polynomial in `n`" by handing algorithms the
+  security parameter in unary. Here the boundary widths are polynomially bounded by
```
dtumad, line 50: Same point. The bullet further down — "Bounds are functions of `n` alone" — is the accurate framing for an indexed non-uniform family; the width bound's job is ruling out superpolynomial boundary values, not reconciling the two cost measures.

ToMathlib/Computability/CslibPolyTime.lean
```diff
@@ -0,0 +1,214 @@
...
+  outputsFunInTime a := (h.outputsFunInTime a).of_le (h.bounds _)
+  poly := h.poly
+  bounds _ := le_rfl
+
+theorem PolyTimeComputable.monotone_normalize_timeBound {f : List Symbol → List Symbol}
+    (h : PolyTimeComputable f) : Monotone h.normalize.timeBound :=
+  fun _ _ hmn => Polynomial.eval_le_eval hmn
+
+/-- The description size of a machine witness over the two-symbol tape alphabet: its
+number of states. Over the fixed `Bool` alphabet the transition table has exactly three
```
dtumad, line 70: `Θ(d log d)`, not `Θ(d)` — each of the three rows per state names a next state, which costs `log d` bits. Polynomially equivalent, which is all the counting bound needs.

dtumad (commented):
Two conceptual corrections in the documentation, both about what the width bound buys.

**The `1^n` reconciliation does not follow from a width upper bound.** `BitEncoding.lean` says the polynomial width bound "is the formal content of the Katz–Lindell `1^n` convention: all game values at parameter `n` have `poly(n)`-length representations, so 'polynomial in `n`' and 'polynomial in the input length' agree", and `Coinductive/PolyTime.lean` repeats it — "the boundary widths are polynomially bounded by definition (`BitEncFam.widBound`), so the two readings agree — the width bound *is* the `1^n` convention", and again that widths being `poly(n)` "is what makes this equivalent to input-length-based bounds".

The `1^n` convention works by supplying a *lower* bound: handing the algorithm `n` symbols forces the input length to be at least `n`, so anything polynomial in `n` is polynomial in the input length. A polynomial *upper* bound on width gives the converse direction only, and the direction that is actually needed fails. `Unit` at every `n` is the counterexample: its canonical width is `0`, so a machine running in `n²` steps is polynomial in `n` and not polynomial in its input length. `widBound` cannot recover unary padding, because it does not force the boundary to carry `n` at all.

The model is fine — it just is not this. `PolyTime.lean` already says the honest thing a few bullets down: "Bounds are functions of `n` alone, matching `PolyQueries`". For an indexed non-uniform family that is a legitimate cost measure, with the security parameter carried by family selection rather than by the input. I would state that directly and drop the equivalence claim; the width bound's real job is preventing superpolynomial boundary values, which is worth saying on its own.

**State count is a polynomial proxy, not a constant-factor one.** `CslibPolyTime.lean` says that over `Bool` the transition table has three rows per state "so the state count measures the machine's description up to a constant factor". Each row has to name a next state, which costs `log d` bits for a `d`-state machine, so the table is `Θ(d log d)` bits and the two measures are polynomially — not constant-factor — related. Polynomial equivalence is all the counting bound needs, so this is a one-word fix. The paragraph immediately after is already careful about exactly this kind of undercounting for growing alphabets, which makes the constant-factor claim the odd one out.

Neither of these touches a theorem statement — `size` is still the right quantity to bound and the counting argument goes through unchanged.

dtumad:
Subsumed by #576 (/ other follow ups)

PR #502, opened 2026-07-28
docs+test(RoundByRound): disclose the per-message reading of Def 3.12 and certify the extraction clause

alexanderlhicks:
Follow-up to #475, from validating it against the source. Two independent changes, both small.

## 1. `ExtractionCondition` is not vacuously satisfied — certificate

`ExtractionCondition` is a conditional, and for `oneRound` its antecedent `error < Pr[escape]` is **unsatisfiable** at `1 / 2`: escape has probability exactly `1 / 2`, so the strict trigger never fires. `oneRound_extractionCondition_half` is therefore vacuously true — axiom-clean while asserting nothing about extraction. That is the failure class in `docs/agents/gotchas.md` §14, and as §14 notes, `#print axioms` cannot detect it. No instance reached the relation-holds branch of `extractionCondition_iff_isBounded` either.

`oneRoundValid` supplies the missing witness. It differs from `oneRound` only in its relation, so `extract`'s constant output satisfies it. Following §14's fix — a witness for the whole bundle rather than each hypothesis separately:

- `oneRoundValid_hypotheses_satisfiable` exhibits a context that is doomed **and** whose escape probability exceeds the error, so the clause's antecedents are jointly inhabitable;
- `oneRoundValid_extractionCondition_zero` discharges the clause at `error = 0` *because extraction succeeds*, with the trigger firing rather than silent;
- `oneRoundValid_isBounded_zero` reaches the relation-holds branch of the bridge;
- `oneRoundValid_extensionalConditions` bundles all three extensional clauses.

Read against the existing `oneRound_not_extractionCondition_zero` — same error, failing extractor, condition fails — this brackets the definition from both sides and pins down that `ExtractionCondition` measures the extractor rather than the trigger.

## 2. Def 3.12 quantifier placement — docstring

`ExtractionCondition` quantifies the prover message **outside** the escape hypothesis, so each message carries its own obligation. Definition 3.12 places it inside:

> If `(x,τ) ∈ PartTr(i−1)`, `(x,τ) ∈ D`, and **for all `m ∈ M_i`** we have `Pr_c[(x,τ,m,c) ∉ D] > ε_i`, then `Ext(x,τ,m)` outputs a witness…

Read literally that is `(∀ m, P m) → (∀ m, Q m)`, which this clause **strictly strengthens**.

The per-message reading is the right one, and the source supports it twice: the same clause names the specific transcript `(x, τ, m)` an *RBR extractable partial transcript*, which carries no information under the literal reading; and the proof of Theorem 1.2 concludes that `(x, τ, EndNode(τ))` is extractable from *that one message's* escape probability. The literal `∀ m` appears carried over from Definition 3.11's soundness clause, where it belongs to the conclusion.

So this is not a defect and it errs in the safe direction — but it is a fourth deviation beyond the three the module header discloses, and it makes "gives the extensional clauses of Definition 3.12" slightly overstated. The docstring now records it, and repeats the extensionality caveat where a reader lands via go-to-definition.

## Validation

- `lake build` clean (2713 jobs) on merged `main` @ `9d2f9b00`.
- `#print axioms` on all eight added declarations: `[propext, Classical.choice, Quot.sound]`. No `sorryAx`, no custom axiom.
- `lake exe lint-style ToMathlib VCVio Extern LatticeCrypto Examples VCVioWidgets Interop` exits 0.
- `theorem` rather than `lemma` throughout, matching `VCVioTest/` (26 `theorem`, 0 `lemma`); the library's own preference runs the other way, so say if you'd rather these match it.

Two additions were considered and dropped as unjustifiable: a concrete evaluation of `experiment_advantage` (verifies plumbing behind an already-proven `simp` lemma — an unconditional equation, so no vacuity to certify) and a bundled `ExtensionalConditions` for `oneRound` itself (bundles the vacuously-true conjunct; the `oneRoundValid` bundle is strictly better evidence for the same claim).

Context in https://github.com/Verified-zkEVM/VCVio/pull/475#issuecomment-5088847704.


dtumad (approved):
Reviewed exact head 19393495 against current main and BGTTZ Definition 3.12 / Theorem 1.2. The literal quantifier placement and the per-message reading are disclosed accurately. The new oneRoundValid certificate is genuinely non-vacuous: it uses the same doomed context and escape distribution, the trigger 0 < 1/2 fires, the extracted false witness satisfies the modified relation, and the existing failing oneRound certificate supplies a negative control. The merged module build, style and isolation checks, and an axiom audit all pass; current CI is green.

PR #503, opened 2026-07-28
fix(LatticeCrypto): bind SelfTargetMSIS solutions to their hashed preimage

alik-eth:
Extracts the generic `SelfTargetMSIS` binding repair as a small, self-contained PR — step 1 of the decomposition @dtumad proposed on #467/#479. It is independently necessary (it closes a soundness gap in the tailored ML-DSA self-target problem) and it unblocks a sound review of #479.

## The soundness gap

`SelfTargetMSIS.Problem.isValid` received only `Challenge → Target → HashOutput → Response`, never the RO **preimage** the experiment looks up. So `mldsaSTMSIS.isValid` recomputed a commitment `w'` from `(pk, c̃, (z, h))` and verified against that same recomputed value, without ever requiring the commitment component of the *hashed* pair `(msg, w)` to equal `w'`. The relevant equality closed as `X = X`. That makes the tailored problem strictly easier than the literature relation — a zero-response / zero-hint solution wins after querying an unrelated commitment, subject only to the verifier norm gates.

## The repair

- **Generic** (`ShortIntegerSolution.lean`): `isValid : Challenge → Target → HashInput → HashOutput → Response → Bool`, and `experiment` now passes the cache-consistent `hashInput` (the same preimage it already looks up) into `isValid`. The structure's type parameters are unchanged, so abstract references to `SelfTargetMSIS.Problem …` elsewhere are unaffected.
- **ML-DSA binding** (`SecurityNMA.lean`): `mldsaSTMSIS.isValid` now additionally requires `hashInput.2 = w'` — the recovered commitment must equal the commitment component of the hashed preimage. On the accepting path the forger's own returned `(msg, w')` *is* the queried preimage, so the extraction proof discharges the binding for free (it reduces to `w' = w'`); the read-back proof (`stmsis_tail_le`) is updated accordingly.
- **Characterization** (`mldsaSTMSIS_isValid_eq_true_iff`): acceptance is exactly *verifier acceptance ∧ the binding conjunct*, stated explicitly so the self-target requirement stays visible — an instantiation that silently dropped the preimage would fail this lemma, making the regression hard to reintroduce.

## Scope (deliberately narrow)

This PR fixes **only** the SelfTargetMSIS binding. The other items from the #467 review are separate, larger pieces and are **not** in this PR:
- the short-secret MLWE model / exact short-key swap (the full-ring-uniform MLWE issue) — a focused model PR;
- the explicit reduction from this tailored verifier relation to the standard-STMSIS normal form (`H([I_m | A]·y, μ)` with the challenge as `y`'s final coefficient block) — the endpoint bridge;
- the generic CMA-to-NMA engine.

The characterization lemma here is the lightweight binding statement, not the full algebraic normal-form bridge (that lives with the standard-STMSIS reduction).

## Verification

Off current `main`. `lake build LatticeCrypto` (and the full CI library set) green on Lean v4.32.0; `mldsaSTMSIS_isValid_eq_true_iff` and the affected `nmaAdvantage_keygen1_le_stmsis` extraction kernel-check `[propext, Classical.choice, Quot.sound]`. No statement outside the three touched declarations changes.

### DIFF
--- a/LatticeCrypto/HardnessAssumptions/ShortIntegerSolution.lean
+++ b/LatticeCrypto/HardnessAssumptions/ShortIntegerSolution.lean
@@ -28,9 +28,13 @@ over a finite coefficient ring.
 ## SelfTargetMSIS
 
 The SelfTargetMSIS problem (introduced in the Dilithium security proof): given a random matrix
-`A` and access to a random oracle `H`, find `(c, z)` such that `Az = c*t + w` where `w` has
-high bits matching a specific target, and `z` is short. The "self-target" refers to the fact
-that the adversary must produce a hash preimage `c = H(...)` as part of the solution.
+`A` and access to a random oracle `H`, find a hash preimage `(μ, w)` and a short response `z`
+such that `c = H(μ, w)` and `Az = c*t + w'` where the recomputed commitment `w'` matches `w`.
+The "self-target" is this binding of the solution to its own hash preimage: `isValid` receives
+the preimage alongside the RO output `c`, and an instantiation must require the commitment it
+recomputes from the response to equal the commitment component `w` of the hashed preimage. The
+experiment enforces the RO side of the binding by looking the preimage up in the oracle's
+query cache.
 
 ## References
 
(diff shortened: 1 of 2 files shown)


dtumad (approved):
Reviewed the exact head against the generic experiment, the ML-DSA extraction, and the standard self-target shape used in the Dilithium literature. This is the focused soundness repair the stack needs: the experiment still enforces RO consistency through the cache, but now also makes that exact preimage available to validity; the ML-DSA instance uses it to require the hashed commitment to equal the commitment recovered from the response. The characterization theorem exposes that conjunct directly, and the NMA extractor preserves it by querying and returning the same (message, commitment) pair.

The scope is appropriately narrow: this fixes the previously trivial tailored relation without claiming that the tailored UseHint/HighBits relation has already been reduced to standard [I | A]-form SelfTargetMSIS. I also built LatticeCrypto from this exact head on Lean 4.32. This looks safe to land as the dependency root for #504/#506.

PR #504, opened 2026-07-28
feat(MLDSA): short-secret MLWE model and the exact NMA key-swap hop

alik-eth:
Step 2 of the #467 decomposition @dtumad proposed: the corrected short-secret MLWE model and the exact NMA key-swap hop, as a focused, purely additive PR.

**Stacked on #503** (the SelfTargetMSIS binding repair) — until #503 lands, its diff shows here too; this PR's own content is the ~489 additive lines in `SecurityNMA.lean` + `Scheme.lean`. The two are independent in content (MLWE leg vs SelfTargetMSIS leg); the stacking is just to keep the decomposition ordered.

## What this adds

The MLWE leg of the ML-DSA NMA reduction, modelled over the **η-bounded box** rather than the full ring — fixing the information-theoretically-zero-advantage issue of the earlier full-ring-uniform `mldsaMLWE`:

- **`sampleShortVec`** (uniform on the η-box), **`keygenShort`/`keygenShort1`** (idealized key generation sampling ρ, K, and short (s₁, s₂) independently).
- **`validKeyPairShort`** (the ∃-material key relation) + **`identificationSchemeShort`**, with the satisfiability certificate **`keygenShort_generable`** (`⟨hrShort, rfl⟩`) — so the relation/pinning hypotheses these introduce are inhabited, not vacuous.
- **`mldsaMLWEShort`** (seed-based short-secret MLWE) and **`mldsaMatrixMLWE`** (the standard uniform-matrix form), with **`expandAIdealization`** (the named XOF-as-random-matrix assumption, computational reading disclosed) and the proven seed→matrix bridge **`advantage_mldsaMLWEShort_le_matrix`** via **`matrixLift`**.
- **`nma_keyswap_hop_short`** — the headline: the NMA-game gap between the real and uniform-`t` short key generators is bounded by the `mldsaMLWEShort` distinguishing advantage of `distinguisherBShort`, as an **exact identity** — no `idealGap`/`hSlack` statistical-slack term (the earlier `honestSamplingSlack` stopgap is not used).

## Scope

Purely additive — no existing declaration is modified or removed. The old `mldsaMLWE`/`nma_security` chain on the base is untouched here; it is retired when the NMA headline moves over (a later step). The SelfTargetMSIS/extraction/headline development (`mldsaSTMSISShort`, `nma_security_short/fips`, ...) is **not** in this PR — it is steps 3–4.

## Verification

Off current main (v4.32.0). `lake build ToMathlib VCVio LatticeCrypto Extern HashSig Examples VCVioWidgets` green. `nma_keyswap_hop_short`, `advantage_mldsaMLWEShort_le_matrix`, and `keygenShort_generable` all kernel-check `[propext, Classical.choice, Quot.sound]`. Both touched files CI-lint clean.

### DIFF
--- a/LatticeCrypto/HardnessAssumptions/ShortIntegerSolution.lean
+++ b/LatticeCrypto/HardnessAssumptions/ShortIntegerSolution.lean
@@ -28,9 +28,13 @@ over a finite coefficient ring.
 ## SelfTargetMSIS
 
 The SelfTargetMSIS problem (introduced in the Dilithium security proof): given a random matrix
-`A` and access to a random oracle `H`, find `(c, z)` such that `Az = c*t + w` where `w` has
-high bits matching a specific target, and `z` is short. The "self-target" refers to the fact
-that the adversary must produce a hash preimage `c = H(...)` as part of the solution.
+`A` and access to a random oracle `H`, find a hash preimage `(μ, w)` and a short response `z`
+such that `c = H(μ, w)` and `Az = c*t + w'` where the recomputed commitment `w'` matches `w`.
+The "self-target" is this binding of the solution to its own hash preimage: `isValid` receives
+the preimage alongside the RO output `c`, and an instantiation must require the commitment it
+recomputes from the response to equal the commitment component `w` of the hashed preimage. The
+experiment enforces the RO side of the binding by looking the preimage up in the oracle's
+query cache.
 
 ## References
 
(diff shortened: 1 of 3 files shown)


dtumad (approved):
Reviewed the now-clean 489-line slice after #503 landed. The corrected model matches the relevant ML-DSA/Module-LWE semantics: s1 and s2 are sampled from the eta-bounded centered box, rather than using the information-theoretically trivial full-ring-uniform error; keygenShort and keygenShort1 differ only by replacing A·s1+s2 with uniform t; and the NMA game discards the secret key, so the proof correctly strips the independent K draw (and the now-unused short vectors on the uniform branch). The resulting key-swap is an exact game identity.

The seed-to-matrix bridge is also honestly scoped. `expandAIdealization` states the XOF-as-random-matrix replacement explicitly, and its docstring correctly notes that the unrestricted deterministic-function form is not a small-epsilon fact pending a cost-bounded model. `matrixLift` ignores the challenged matrix itself but receives its A·s+e target, which is exactly what the comparison requires.

I built LatticeCrypto from the exact post-merge head on Lean 4.32. One optional wording refinement: `sampleShortVec k b` is uniform over the subtype defined by the centered infinity-norm bound for every b; describing this as each coefficient uniform on the integer interval [-b,b] is literally exact only while b stays within the canonical centered residue range. That includes every intended ML-DSA eta, so it does not affect this model or any theorem. This looks safe to land once the rerun finishes.

PR #506, opened 2026-07-28
feat(MLDSA): tailored SelfTargetMSIS problem and the NMA extraction bound

alik-eth:
The tailored ML-DSA SelfTargetMSIS leg of the #467 decomposition — the STMSIS side to match the MLWE side in #504. Purely additive.

**Stacked on #504** (short-MLWE model) → #503 (SelfTargetMSIS binding). Its own content is the ~313 additive lines in `SecurityNMA.lean`; the rest of the diff is the two branches below it (shrinks as they land).

## What this adds

The tailored self-target problem and the NMA→STMSIS extraction, built on the bound `isValid` API from #503 and the short scheme from #504:

- **`mldsaSTMSISShort`** — the tailored problem; its `isValid` carries the self-target binding conjunct `hashInput.2 = w'` (from #503's threaded preimage) and runs the `identificationSchemeShort` verifier; `sampleParams` draws the uniform-`t` key via `keygenShort1`.
- **`nmaAdvantage_keygenShort1_le_stmsis`** — the headline: the uniform-`t` NMA advantage is bounded by the tailored-STMSIS advantage of `extractorCShort`.
- **The algebraic characterization** — `computeWApprox_eq_mul_sub_smul` (the `UseHint`/`computeWApprox` bridge to `Â·z − c·(t₁·2^d)` over R_q, under the transform laws), `identificationSchemeShort_verify_eq_true_iff`, `stmsisAlgebraicSolution` (norm gates + matrix equation + binding), and `mldsaSTMSISShort_isValid_iff` / `_isValid_expandA_iff`. At the sampled matrix, acceptance is exactly the norm gates plus the binding — so the self-target requirement is load-bearing, not decorative.

## Scope

This is the **tailored** verifier relation. The explicit reduction to the standard SelfTargetMSIS normal form (`H([I_m | A]·y, μ)` with the challenge as `y`'s final coefficient block) is a separate follow-up; the algebraic characterization added here is the foundation for it. The composed headline (`nma_security_short`/`nma_security_fips`) and the CMA-to-NMA engine are **not** in this PR.

## Verification

Off current main (v4.32.0). `lake build ToMathlib VCVio LatticeCrypto Extern HashSig Examples VCVioWidgets` green. `nmaAdvantage_keygenShort1_le_stmsis`, `mldsaSTMSISShort_isValid_expandA_iff`, and `computeWApprox_eq_mul_sub_smul` all kernel-check `[propext, Classical.choice, Quot.sound]`. Touched file CI-lint clean; purely additive (no base declaration changed).


dtumad (commented):
The core read-back/extraction argument looks sound to me. In particular, the `keygenShort1` parameter distribution matches the short NMA game, both tails share the same forger/RO execution, and verifier acceptance supplies the commitment equality needed by the repaired self-target binding. I also like that the PR description explicitly scopes this as a tailored verifier relation rather than claiming the standard normal form already follows.

Two small refinements would make that boundary especially clear in the checked API/docs:

- `identificationSchemeShort_verify_eq_true_iff` currently says an accepted forgery carries the “full literature SelfTargetMSIS relation” and points to a module-docstring qualification that does not appear to be present. The relation proved here is the useful `UseHint`/`HighBits` tailored relation; the EasyCrypt/CRYPTO 2023 problem instead packages the linear relation as `[I_m | A] · y` with the challenge in the final coefficient block. It may be clearer to call this the tailored algebraic verifier relation here and say directly that the standard-normal-form bridge remains follow-up work. The nearby “see the module docstring” references could be adjusted at the same time.

- Both `identificationSchemeShort_verify_eq_true_iff` and `mldsaSTMSISShort_isValid_iff` accept the full `Primitives.Laws prims nttOps`, although their proofs consume only `h_laws.transform`. Since `computeWApprox_eq_mul_sub_smul` already takes the precise `NTTRingLaws nttOps` premise, using that same minimal premise in these two characterizations would keep this semantic certificate independent of unrelated primitive-law assumptions (and of the separate Laws work in #479).

These look localized rather than changes to the extraction theorem itself. For comparison, the standard formulation I used is the Formosa EasyCrypt development (`SelfTargetMSIS`/`SimplifiedScheme`): https://github.com/formosa-crypto/dilithium/tree/main/proofs/security and the CRYPTO 2023 paper: https://eprint.iacr.org/2023/246

alik-eth:
@dtumad Both refinements from your review are applied (pushed a while back, so flagging that this is ready when you next have a window).

**Minimal premise.** `identificationSchemeShort_verify_eq_true_iff` and `mldsaSTMSISShort_isValid_iff` now take `(h_transform : NTTRingLaws nttOps)` instead of the full `Primitives.Laws prims nttOps` — matching the premise `computeWApprox_eq_mul_sub_smul` already used, since that was the only part either proof consumed. These two semantic certificates are now independent of the unrelated primitive-law assumptions (and of the Laws work in #479). `mldsaSTMSISShort_isValid_expandA_iff` needed no laws and is unchanged.

**Tailored vs. standard framing.** You were right that the "full literature SelfTargetMSIS relation" wording overclaimed and that the module-docstring qualification it pointed at wasn't there. The dangling pointers are gone, replaced by a self-contained *Tailored vs. standard SelfTargetMSIS* section in the module docstring which states plainly that what is proved here is the tailored `UseHint`/`HighBits` verifier relation, and that reducing it to the standard normal form — `[I_m | A] · y` with the challenge in the final coefficient block of the short preimage — is follow-up work. The per-declaration docstrings now say the same thing locally rather than deferring.

Thanks for the Formosa/CRYPTO 2023 pointers — the standard form there is what the follow-up targets.

**Verification.** `lake build LatticeCrypto` green; `identificationSchemeShort_verify_eq_true_iff`, `mldsaSTMSISShort_isValid_iff`, and the headline `nmaAdvantage_keygenShort1_le_stmsis` all kernel-check `[propext, Classical.choice, Quot.sound]`; touched file CI-lint clean. No statement changed beyond the hypothesis weakening.

dtumad (commented):
Follow-up to my 2026-07-28 review, re-checked at `f87e7f99`. Both points are addressed: `identificationSchemeShort_verify_eq_true_iff` and `mldsaSTMSISShort_isValid_iff` now take `NTTRingLaws nttOps` rather than the full `Primitives.Laws`, so the semantic certificate no longer drags in unrelated primitive-law assumptions, and the module documentation consistently calls the problem tailored with the normal-form bridge named as follow-up work.

On the substance, this is the repair that mattered. The earlier relation did not bind the hash preimage, so a candidate could satisfy it with a degenerate witness — `z = 0` and an empty hint — that has nothing to do with the commitment being attacked. Requiring the stored hash input's second component to equal the recovered `w'` is what closes that, and pairing it with an actual random-oracle cache-membership check in the NMA experiment means the extractor cannot manufacture a preimage it never queried. The norm and hint gates line up with the verifier's own gates, the short-secret key marginal is the same distribution on both sides of the reduction, and the toy/dummy branches do not route around the fixed-point binding.

One documentation-accuracy nit while you are in this file. It is not from this PR — the text is already on `main` at `SecurityNMA.lean:809` and `:830`, outside your hunks — but it sits right next to the binding you just repaired, and the surrounding argument is otherwise careful about exactly this kind of claim. `extractorC`'s docstring says the aborting branch "is mapped to a dummy preimage with a zeroed response, which the STMSIS RO-consistency check rejects", and the inline comment says it "fails RO consistency / `isValid`". That is stronger than what is proved, and stronger than what the reduction needs: nothing rules out the forger having already queried `H(msg, default)`, and the target is an upper bound on the NMA side, so any extra success on the STMSIS side only loosens the inequality in the direction we want. Something like "the STMSIS experiment reads back `H(msg, default)`; the NMA tail does not depend on it, and extra right-hand successes only help the bound" would be both weaker and true. Purely wording — no hole in the reduction, and no reason to hold this PR for it.

dtumad (approved):
Approving at f87e7f99 — the head my 2026-08-17 review already checked in substance. Both refinements are in (minimal `NTTRingLaws` premise; tailored-relation framing with the normal-form bridge scoped as follow-up), and a full local build of the current-main merge passes together with the other queued PRs. The `extractorC` docstring wording nit from that review is pre-existing text on main outside these hunks; it will get a separate docs-only follow-up rather than expanding this diff.

PR #507, opened 2026-07-28
feat(FiatShamir): Fiat-Shamir-with-aborts CMA-to-NMA reduction engine

alik-eth:
The generic Fiat-Shamir-with-aborts **CMA-to-NMA reduction engine** — step 4a of the #467 decomposition, reviewable on its own. Pure VCVio machinery: no ML-DSA (or any LatticeCrypto/Extern) content, no imports of them (verified).

Current `WithAbort/Security.lean` on main is a 151-line stub: the loss functions `cmaToNmaLoss`/`perKeyLoss` plus two `sorry`-backed placeholder theorems (`euf_cma_bound`, `euf_cma_bound_perfectHVZK`, marked "placeholder, not the final theorem"). This PR replaces that stub with the real reduction and its supporting development.

## What this adds

The statistical CMA-to-NMA reduction for Fiat-Shamir with aborts (after Theorem 3, CRYPTO 2023): **`euf_cma_to_nma`**, which bounds an EUF-CMA adversary's advantage by the managed-RO NMA advantage of the simulated adversary plus `cmaToNmaLoss`, under pointwise good-key hypotheses (`0 ≤ ε`, `0 ≤ p_abort < 1`, `0 ≤ δ`, and a per-attempt HVZK simulator). The development is organised into phase modules under `WithAbort/Security/` (`Loss`, `BodyHops`, `GhostReadCharge`, `HiddenReadFold`, `CouplingEngine`, `ReadRecording`, `TapeFactorization`, `HopLemmas`, `NMAReduction`, `Assembly`) and the ghost-layer machinery under `WithAbort/GhostBodies/` (`Bodies`, `GhostLayer`, `Projections`, `BodyBounds`, `NMAHandler`), each re-exported by its umbrella.

## Removed: two sorry placeholders

The stub's `euf_cma_bound` / `euf_cma_bound_perfectHVZK` are **deleted** — they are `sorry`-backed placeholders superseded by the proven `euf_cma_to_nma`. Confirmed to have **zero consumers** repo-wide (including tests); the only `euf_cma_bound` references in the tree are to the unrelated `FiatShamir.euf_cma_bound` Sigma variant in a different namespace. `cmaToNmaLoss`/`perKeyLoss` are preserved (in `Security/Loss.lean`, re-exported by the umbrella), so their consumers are unaffected. Net effect: **−2 `sorry`**.

## Scope

Generic engine only — the ML-DSA instantiation (`nma_security_short`/`_fips` composing this engine with the short-MLWE model #504 and the tailored SelfTargetMSIS leg #506) is a separate follow-up.

## Verification

Off current main (v4.32.0). `lake build` of the full CI library set + the test libraries green; the 17 engine files carry no `sorry`. `FiatShamirWithAbort.euf_cma_to_nma` kernel-checks `[propext, Classical.choice, Quot.sound]`. No engine file imports any LatticeCrypto/Extern/MLDSA content; CI-lint clean on all 17.


dtumad (commented):
I took a correctness-focused pass over the headline and the difficult
read-recording/NMA bridges. The quantitative shape matches the Formosa
EasyCrypt reduction: the two random-oracle collision charges give
`2·qS·(qH+1)·ε/(1-p)`, and the signing/signing term is
`qS·ε·(qS+1)/(2·(1-p)^2)`. The additional
`qS·ζ_zk/(1-p)` is a sound consequence of this API's per-attempt TV-distance
HVZK assumption. The pointwise good-key guessing hypothesis here is stronger
than the averaged condition used in the paper, which is conservative rather
than a gap. I also checked the Option-B managed-RO bridge and the direct
value-substitution argument for rejected commitments; I have not found a
mathematical counterexample in those paths.

There are a few scope/organization points that I think would make this much
easier to audit and maintain before merging:

* The final pair-charge path currently goes
  `readRecord_expected_pairs_le` → `readRecord_expected_pairs_tape_le` →
  `readRecord_expected_pairs_nontape_le`. The first step rewrites the inline
  run to the tape run, while the second immediately uses the same fold equality
  in reverse and applies the already-proved direct inline induction. Those
  transports cancel; the headline can use
  `readRecord_expected_pairs_nontape_le` directly. Would it make sense to use
  that as the live proof and move the tape factorization to clearly separate
  reusable/follow-up infrastructure? There is a similar amount of eager/lazy
  coupling material explicitly described as not being on the live path. A
  smaller live slice would make the semantic review substantially clearer
  without discarding that work.

* I compared this with the Cursor/Path work merged in #492. This branch already
  contains that work, so there is no stale-representation problem, and
  `IsQueryBoundP` is already literally a PolyFun `IsRollBound`. Cursors do not
  by themselves solve the hard probability step here: the commitment draws are
  introduced inside the signing handler, rather than being occurrences in the
  original adversary tree. They may still help remove some bespoke structural
  traversal. In particular, `deferredDrawReadImpl` is input-only
  instrumentation of adversarial read queries; it seems worth considering
  whether `withTraceAppendBefore`, `withPath`/`withQueryLog`, or a generic state
  projection/Displayed decoration can express that part. If the current
  handler is preferable, a short explanation of where those generic APIs stop
  being sufficient would also resolve the architectural question.

* A number of comments still record intermediate campaign state rather than
  describing the finished declarations. For example, `NMAReduction.lean`
  successively says the projection is “still-open”, “not yet built”, and “not
  landed” before later recording its resolution; `TapeFactorization.lean`
  repeatedly calls proved results the “sole open core”; and several comments
  refer to `#228`, “multi-week” work, or “banked axiom-clean” stages. Recasting
  these as intrinsic proof explanations would avoid making a completed,
  kernel-checked theorem look unfinished. The `probEvent_ghostRead_bad_le`
  docstring also discusses `p_abort < 0`, although the theorem now assumes
  `0 ≤ p_abort`.

These are large enough that author-directed cleanup seems more appropriate
than a reviewer rewriting the development. The main mathematical chain I
checked looks coherent; my hesitation about merging the current 11k-line form
is chiefly that the live proof and the retained alternative infrastructure are
interleaved and documented as if several already-proved obligations were
still open.

For comparison:

* Formosa EasyCrypt development:
  https://github.com/formosa-crypto/dilithium/blob/main/proofs/security/FSa_CMAtoKOA.eca
* CRYPTO 2023 paper: https://eprint.iacr.org/2023/246

alik-eth:
@dtumad All three organizational points from your review are addressed (pushed a while back, so flagging that this is ready when you next have a window).

**Live path vs. retained infrastructure.** You were right that the transports cancel. The headline now consumes `readRecord_expected_pairs_nontape_le` directly — the already-proved direct inline induction — and `readRecord_expected_pairs_tape_le` / `readRecord_expected_pairs_le` are documented as separate reusable infrastructure that is not on the live path. Same for the eager/lazy coupling material. Nothing was discarded; the live slice is just smaller and labelled as such.

**Cursors / #492.** I kept the current handler and added an intrinsic note at `deferredDrawReadImpl` explaining where the generic APIs stop being sufficient: `withTraceAppendBefore` / `withPath` / `withQueryLog` and state-projection decorations instrument *occurrences in the computation being simulated*, but the commitment draws this bound charges are introduced inside the signing handler by `ghostSignDrawBody` when it answers a signing query — they are not occurrences in the adversary tree at all. So input-only instrumentation of adversarial read queries cannot express the deferred signing-body draw, which is the part the probability argument needs. Agreed that cursors may still subsume some of the structural traversal; that seemed better as a separate incremental change than folded in here.

**Campaign-state comments.** Recast as intrinsic descriptions of the finished declarations across the `Security/` and `GhostBodies/` modules — issue references, "banked", "multi-week", "still-open", "sole open core", and the "not yet built → later resolved" progressions in `NMAReduction.lean` and `TapeFactorization.lean` are gone. You were right that they made completed, kernel-checked theorems read as unfinished. The `probEvent_ghostRead_bad_le` docstring (and its two siblings) now state the real hypothesis `0 ≤ p_abort` rather than discussing `p_abort < 0`.

**Verification.** `lake build VCVio` green; no `sorry` anywhere in the `WithAbort/` tree; `euf_cma_to_nma` kernel-checks `[propext, Classical.choice, Quot.sound]`; all touched files CI-lint clean. The only non-comment change is the one-line repoint of the pair charge to `_nontape_le`.

VCVio/CryptoFoundations/FiatShamir/WithAbort/Security/Loss.lean
```diff
@@ -0,0 +1,87 @@
...
+the all-attempts-reprogram hybrid and in the accepted-only-reprogram hybrid, hence the
+factor 2; the `qH + 1` accounts for the final verification query). The second term pays
+for collisions among the signing oracle's own commitments. The third term glues the
+per-attempt simulator across the restart loop, whose expected length is at most
+`1/(1-p)` (see `tvDist_firstSome_le_geometric`); a simulator for the accepted-transcript
+distribution itself (the paper's acHVZK notion) would shave this `1/(1-p)` factor. -/
+noncomputable def cmaToNmaLoss (qS qH : ℕ) (ε p ζ_zk δ : ℝ) (_hp : p < 1) : ℝ :=
+  2 * qS * (qH + 1) * ε / (1 - p) +
+  qS * ε * (qS + 1) / (2 * (1 - p) ^ 2) +
+  qS * ζ_zk / (1 - p) +
```
dtumad, line 71: This is the change I mean below: `main`'s `cmaToNmaLoss` (`Security.lean:67`) has `qS * ζ_zk`. The strengthening is right, but `LatticeCrypto/MLDSA/Security.lean` is not in this diff and still prints the old term at `:34`, `:281`, `:306`, `:325` — including the "This is exactly `FiatShamirWithAbort.cmaToNmaLoss qS (qH + qS) ...`" claim that closes the ML-DSA wrapper's docstring.

dtumad (commented):
Follow-up to my 2026-07-28 review, re-checked at `652d2598`.

The organizational point I raised is resolved: the live coincidence path now calls the direct inline bound rather than routing through the tape factorization and back, so the transports no longer cancel, and the tape material reads as reusable infrastructure instead of an unverified critical hop. That makes the semantic slice a lot smaller to audit.

I re-derived the loss independently and get the same thing the definition computes:

`2·qS·(qH+1)·ε/(1-p) + qS·ε·(qS+1)/(2·(1-p)²) + qS·ζ_zk/(1-p) + δ`

with the two random-oracle read charges giving the doubled first term, pair coincidences the quadratic signing term, retry-amplified HVZK the third, and bad-key mass `δ`. The replacement also fixes the real problem with the old placeholder, which was not that it was unproved but that it was unsound: signed `ε`, abort probability and slack all passed through `ENNReal.ofReal`, so a negative loss could vanish rather than falsify the statement. The new premises pin nonnegativity and `p_abort < 1`, the denominators are well defined at the boundary, and the managed/plain NMA bridge is an exact equality rather than an inequality in the convenient direction.

One thing to fix before this lands, in a file this PR does not touch.

This PR strengthens the HVZK term from `qS·ζ_zk` to `qS·ζ_zk/(1-p)`. I agree that is the right per-attempt accounting — the ζ charge is paid per rejection-sampling attempt, so it inherits the same `1/(1-p)` retry amplification the other statistical terms have. But `LatticeCrypto/MLDSA/Security.lean` is outside the diff and still documents the old value in four places (`:34`, `:281`, `:306`, `:325`), and `:281` is inside the docstring of the ML-DSA `cmaToNmaLoss` wrapper, which closes with "This is exactly `FiatShamirWithAbort.cmaToNmaLoss qS (qH + qS) ...`". That sentence is true on `main` and false the moment this merges. The definition stays correct — it just forwards — so this is purely the printed formula, but it is the formula a reader checks the reduction against.

Worth doing in the same pass: drop the comments that narrate discarded proof attempts or describe the direct path as a future replacement, now that it is the live one.

Otherwise the game sequence (real → programmed → transformed → simulated → managed-NMA) is explicit and the hops check out; I did not find a counterexample in the abort/failure accounting.

quangvdao (changes requested):
Reviewed at `652d2598b2e63199ad07ff771e07674aadaa5197`.

The reduction chain itself checks out: the real→programmed→transformed→simulated→managed-NMA hops are explicit; the direct inline pair-charge is now the live path; nonnegativity and `p_abort < 1` prevent the old `ENNReal.ofReal` vacuity; and the assembled loss is

`2·qS·(qH+1)·ε/(1-p) + qS·ε·(qS+1)/(2·(1-p)^2) + qS·ζ_zk/(1-p) + δ`.

One blocking consistency finding remains:

- **P2 — update the public ML-DSA wrapper documentation with the strengthened HVZK term.** This PR changes the generic loss from `qS·ζ_zk` to `qS·ζ_zk/(1-p)`, but `LatticeCrypto/MLDSA/Security.lean` still prints the old expression at lines 34, 281, 306, and 325. Line 281 is in the `MLDSA.cmaToNmaLoss` wrapper docstring immediately before the claim that it is exactly the generic loss. The definition forwards correctly, but merging this head makes the documented theorem formula false. Please update those four formulas in this PR, then rerun CI.

I found no added `sorry`, `admit`, axiom, unsafe/native trust, or linter suppression, and exact-head hosted CI is otherwise green. Once the wrapper documentation agrees with the new generic statement, I see no remaining blocker.

quangvdao (commented):
Thank you, @alik-eth, for the substantial work here. I re-reviewed the full engine at exact head `652d2598b2e63199ad07ff771e07674aadaa5197`, including the real → programmed → transformed → simulated → managed-NMA chain, the Option-B cache bridge, the query accounting, and the CRYPTO 2023/Formosa source correspondence.

The mathematical reduction chain checks out. I independently recover

`2·qS·(qH+1)·ε/(1-p) + qS·ε·(qS+1)/(2·(1-p)²) + qS·ζ_zk/(1-p) + δ`.

The additional `1/(1-p)` on the HVZK term is a sound consequence of this API using a per-attempt optional-transcript simulator rather than the paper's accepted-transcript simulator. I did not find a P0/P1 error in the probability or cache semantics.

I do see three P2 items to fix before merge:

1. **The public ML-DSA wrapper still prints the old HVZK loss.** `LatticeCrypto/MLDSA/Security.lean` still says `qS·ζ_zk` at lines 34, 281, 306, and 325, although the forwarding definition now computes `qS·ζ_zk/(1-p)`. In particular, the claim immediately after line 281 that the displayed formula is exactly the generic loss becomes false. Please update all four occurrences. This is the already-open changes-requested item.

2. **The split introduces forbidden subdirectory umbrellas.** `VCVio/CryptoFoundations/FiatShamir/WithAbort/GhostBodies.lean` and `.../WithAbort/Security.lean` are thin re-export umbrellas. The policy-governing `docs/agents/gotchas.md` at the review base, lines 199–214, explicitly permits these only at repository roots. The root module already imports the leaf modules. Please delete `GhostBodies.lean`; delete `Security.lean` or make it own actual declarations; and update callers (`Loss.lean`, `LatticeCrypto/MLDSA/Security.lean`, etc.) to import the precise owning modules.

3. **The foundational headline needs a producer-level falsifiable canary.** This slice adds `euf_cma_to_nma`, the managed/plain equality bridge, and the NMA query-bound transfer, but no `VCVioTest`/example directly pins their difficult branches. The later #508 unit-slack/no-query witness establishes inhabitation, not mutation-sensitive coverage. Please add a small toy test with distinguishable messages/commitments/challenges that would fail under: omitted forgery-point erasure, swapped base/ghost projections, `.inl`/`.inr` query-selection mistakes, or drift in the strengthened loss term.

Two P3 documentation cleanups are also worth doing in the same pass:

- `NMAReduction.lean:156–182` describes an earlier/corrected projection rather than the finished declaration intrinsically.
- `NMAReduction.lean:919–921` says a query-bound transfer "should accompany" the NMA lemma even though `simulatedNmaAdv_nmaHashQueryBound` now exists in `Assembly.lean`.

Finally, this head is 14 commits behind current `main` (`3ecd5523`, Lean 4.33). The synthetic merge is textually clean, but the exact final head still needs the full v4.33 build/import/lint/test suite before re-review. Historical exact-head CI is green, `git diff --check` is clean, and this slice adds no `sorry`, `admit`, axiom, unsafe/native trust, or linter suppression.

Thanks again for pushing through a genuinely difficult formal reduction. The hard mathematical chain is in much better shape than the old placeholder; the remaining requests are about making that result accurate, policy-compliant, and regression-resistant at the public boundary.

dtumad:
Deep semantic audit across the open PR set (semantics only). **Verdict: the mathematics is sound — I ran the DFPS23/Barbosa-checklist attacks against the actual definitions and none lands. The remaining blockers are the contract-level items already on record from the 08-17/08-23 threads, unaddressed at head `652d259` (frozen since 08-09).**

For the record, the abort-handling pitfalls from the FSwA literature are all handled correctly:

- **RO programming on aborted transcripts** is the design's spine: `realSignBody` runs the loop under the caching oracle so aborted attempts leave their challenge in the cache; `progSignBody` overwrites at `(msg, w)` on every attempt (`cacheQuery` is a genuine `Function.update` overwrite); `transSignBody` then moves to accepted-only programming. The `2·qS·(qH+1)·ε/(1−p)` term is exactly the price of those two hops.
- **Simulation closeness includes aborts**: `IdenSchemeWithAbort.HVZK` is a TV bound over `Option (Commit × Chal × Resp)` with `none` = abort — a per-attempt *optional-transcript* simulator, stronger than the paper's accepted-transcript (acHVZK) notion, and the extra `1/(1−p)` on the ζ term via `tvDist_firstSome_le_geometric` is its sound consequence.
- **Query accounting is non-vacuous**: `simulatedNmaAdv_nmaHashQueryBound` proves the simulated signing handler issues zero live hash queries, so the NMA adversary genuinely stays at `qH`.
- **The Option-B bridge is an exact equality** resting on EUF-CMA freshness (programmed entries all have the form `(signedMsg, w)`, so erasing a fresh forgery point is a no-op) — the right argument, and stated in the cache-invariant lemma.

Blocking before merge (all previously raised, listed here so the set is in one place):

1. **The four stale `qS·ζ_zk` formula lines in `LatticeCrypto/MLDSA/Security.lean` (34, 278, 301, 320 on main) become false on merge**: this PR's proved `cmaToNmaLoss` carries `qS·ζ_zk/(1−p)` (`Security/Loss.lean:71`), and this PR does not touch the ML-DSA file. Since the *definition* is replaced under the ML-DSA statements, their docstring formulas and the ":281 this is exactly…" claim must be updated in the same change.
2. The thin umbrella files `WithAbort/GhostBodies.lean` and the import-only `WithAbort/Security.lean` re-export layer — fold or justify per repo layout conventions.
3. A producer-level mutation canary (the review's point about pinning the hybrid bodies against silent drift).
4. Rebase: `runtime.evalDist` no longer exists on main (`ProbCompRuntime.evalSPMF` now) — ~29 occurrences here are statement-level rewrites, not import fixes.

One paper cross-check to record in the docstring rather than re-derive: the pointwise `hGuess` on good keys is stronger than ePrint 2023/246's averaged condition (conservative, so sound); worth one sentence saying the averaged form would be needed only for tightness, not soundness.

Coordination: #508 consumes this branch; land this first once the four items above are done.

alik-eth:
All four blocking items are now on the branch; head is `8f6a547e`.

**1. The four `qS·ζ_zk` formula lines** in `LatticeCrypto/MLDSA/Security.lean` (34, 278, 301, 320) — fixed at `60000337`. They read `qS·ζ_zk/(1−p)` now, matching `cmaToNmaLoss` in `Security/Loss.lean`, and the "this is exactly `FiatShamirWithAbort.cmaToNmaLoss …`" claim under the display is true again because the display and the definition agree.

**2. The umbrella modules** — folded at `8f6a547e`, per `docs/agents/gotchas.md` §19 (no re-export umbrellas below the repository roots). `WithAbort/Security.lean` is a content module on `main` that `LatticeCrypto/MLDSA/Security.lean` already imports, so rather than delete that module path I folded `Security/Assembly.lean` into it: the Option-B bridge and the headline `euf_cma_to_nma` now live at `VCVio.CryptoFoundations.FiatShamir.WithAbort.Security`, and `Assembly.lean` is gone. The `GhostBodies.lean` re-export is deleted, with its overview moved to `GhostBodies/Bodies.lean`, the first module of that chain. Each `Security/*` module imports the `GhostBodies` submodule it first uses (`Loss` → `Bodies`, `BodyHops` → `Projections`, `GhostReadCharge` → `BodyBounds`, `NMAReduction` → `NMAHandler`), and the root `VCVio.lean` is regenerated to match.

**3. The producer-level mutation canary** — not yet. It is the one remaining item and will come as its own push.

**4. Rebase** — done at `60000337`: merged `main` through #582, with `runtime.evalDist` → `evalSPMF` and `𝒟[·]` → `𝒮[·]` as statement-level rewrites throughout the engine, as predicted.

Two disclosures from CI on `60000337`. *Check Library File Imports* failed on the PMF/SPMF boundary ratchet: six proof-internal uses in the new engine files (`SPMF.ext` twice in `CouplingEngine`; `evalSPMF_bind_congr (m := SPMF)`, `SPMF.evalSPMF_def`, and one `SPMF (α' × σ')` ascription in `NMAReduction`; `tsum_probOutput_of_liftM_PMF` in the assembled headline) had a ceiling of 0 because those files do not exist on `main`. `8f6a547e` records them in `scripts/pmf_boundary_baseline.tsv`; no existing ceiling was raised. The advisory report also lists `+4` in `SignatureAlg`, `SimulateQ`, and `Sigma/Stateful/Hops`, which this branch does not add: `main` tightened those files after the branch's merge base, and that delta disappears on the next merge.

The paper cross-check sentence on `hGuess` is in the `euf_cma_to_nma` docstring at `8f6a547e`: the pointwise form on good keys is stronger than ePrint 2023/246's key-averaged condition, so it is conservative and sound, and the averaged form would matter only for tightness.

alik-eth:
The remaining item, the producer-level mutation canary, is in at `aaff593d`, together with the two P3 doc notes from the 08-23 review.

**Canary** — `VCVioTest/CryptoFoundations/FiatShamirWithAbortCanary.lean`, on a toy identification scheme over `Bool` (uniform commitment, abort exactly on challenge `false`, response echoes the challenge, verification checks response = challenge; exact simulator). Every pin is a closed probability proved by normalisation, one per mutation class named in the review:

- *loss drift*: `cmaToNmaLoss` at three rational points chosen so that dropping the `1/(1-p)` on the ζ term changes the value (`53/16`, `105/16`, `7/8`);
- *swapped base/ghost projections*: `ghostSignBody` over two attempts, with the joint distribution of the output and both cache layers read at both commitment points pinned at three atoms (`1/8`, `1/16`, `5/16`) — any layer swap zeroes them;
- *`.inl`/`.inr` bookkeeping*: `simulatedNmaAdv` on a probe adversary (one uniform query, the same hash point twice, a signature on `false`, a forgery on `true` whose message bit records whether the two hash answers agreed), observed through the returned managed cache at the uniform point and the four `(message, commitment)` points — three atoms (`1/4`, `1/8`, `1/4`) pin that the uniform answer is never cached, repeated hashes agree, and the signing transcript is programmed at `(false, w)`;
- *omitted forgery-point erasure*: the third probe atom is the direct witness — when the simulator's commitment coincides with the probe's earlier live read at `(true, false)`, Option B erases that cached read, and the atom has probability `1/4` exactly because of it; plus the replay adversary's managed-RO NMA success pinned at `1/4` (accepted signature `1/2` × fresh-challenge acceptance `1/2`; a replayed programmed challenge would give `1/2`), proved through the bridge lemma, and the same `1/4` on the plain `eufNmaExp` side.

The proofs stage the normalisation deliberately: the nested managed/runtime simulation is collapsed with the engine's own `simulateQ_unifSim_run` and the `roSim` lemmas before any probability is expanded, otherwise `simp` sums over cache-valued tuples. Nothing here is a security statement; the file says so.

**P3 doc notes** — `NMAReduction.lean` now describes `proj2` intrinsically (which component must carry the overlay and why, with the swapped assignment named as the one that is not a state function), and the NMA bridge docstring points at `simulatedNmaAdv_nmaHashQueryBound` in `FiatShamir.WithAbort.Security` instead of saying a transfer "should accompany" it.

That closes all four blocking items and both P3 notes on this PR.

alik-eth:
@quangvdao Re-merged on current main, head c268398c. Changes since your last review:

- Ported to Lean v4.34 and the measure-valued runtimes, adopting main's renames (`UnforgeableAdversary`, `unforgeableAdvantage`, `EufNmaAdversary`, `managedRoNmaAdvantage`, `inROM`). Three lemmas are restated at the measure-valued advantages and renamed to match, for example `unforgeableAdvantage_eq_hybridExpAtKey_real`. The replay canary pins still evaluate to 1/4.
- For the 1500-line limit, `TapeFactorization` is split into FirstMoment, TapeFactorization, SignStepCharge and ReadRecordBound, and `CouplingEngine` into CouplingEngine, BodyResampling and DeferredCoupling.
- The two placeholder theorems this PR replaces leave the axiom baseline.

Build, style lint and the axiom sweep pass. The Lint job fails only on `usesRetiredProbability` (100 declarations here); the policy question on that is open with dtumad on #478.

PR #508, opened 2026-07-28
feat(MLDSA): compose the EUF-NMA/CMA headlines; model ExpandA as a programmed random oracle

alik-eth:
The composed ML-DSA EUF-NMA/CMA security headlines, assembled on the landed legs (#503 binding, #504 short-MLWE, #506 SelfTargetMSIS) and the CMA-to-NMA engine of #507.

**Stacks on #507.** Until #507 lands this diff shows its content too; the PR's own delta is `LatticeCrypto/MLDSA/SecurityNMA.lean`, the new `LatticeCrypto/MLDSA/SecurityExpandARO.lean`, `LatticeCryptoTest/MLDSA/NonVacuity.lean`, and the two library roots.

## What this adds

- **`nma_security_short`** — the EUF-NMA bound in the idealized short-key model: the exact key-swap hop, the reprogramming hops, and the tailored-STMSIS extraction, `advantage ≤ ofReal (MLWE + εbridge) + STMSIS`.
- **`nma_security_fips`** — the same at the FIPS seed-derived key generation `keygen0`, at cost `εPRG` through the named `expandSReplacement` assumption (a genuine PRG assumption: the distinguisher never sees the seed).
- **`euf_cma_security_of_nma_short`** — the CMA-to-NMA composition over `identificationSchemeShort` via the #507 engine's `euf_cma_to_nma` and the Option-B bridge.
- **`euf_cma_security_of_nma_fips`** and **`euf_cma_security_of_nma_fips_hvzkReal`** — the same composition over the seed-tagged `identificationScheme` with `nma_security_fips` as the NMA leg; the corollary discharges the HVZK hypothesis with the proved simulator `hvzkSimulatorReal` at the proved bound `hvzkBoundReal` (`idsWithAbort_hvzk_real`, under `Primitives.Laws`, whose approved-parameter witness is `mldsa_laws_inhabited` in `Extern/MLDSA/NonVacuity.lean`).
- **`euf_cma_security_short_of_uniform_advantage_bounds`** — the security-parameter-indexed closure of the loss regime. This is a numerical statement, not computational security: its advantage-bound families dominate every adversary with no cost model restricting them.
- **`nma_security_rom`** — EUF-NMA security with `ExpandA` modelled as a **programmed random oracle**, in the new `LatticeCrypto/MLDSA/SecurityExpandARO.lean`. `ExpandA` becomes a third oracle alongside uniform sampling and the commitment hash, shared by key generation, verification, and the forger, and simulated lazily from a cache. Both reductions program that cache at a self-chosen seed with the matrix their challenger supplies, so the forger's view is the honest game's. The result reaches the literature's uniform-matrix Module-LWE `mldsaMatrixMLWE` with **no idealization slack and no hypothesis beyond the forger**: `advantage ≤ ofReal (MLWE) + STMSIS`. The real branch is an exact identity (`romNmaAdvantage_eq_game0`), the uniform branch is the SelfTargetMSIS bound (`game1_le_stmsis_rom`). The oracle algorithms are pinned to the scheme's: answering `ExpandA` from a fixed table recovers the scheme's own algorithms at that table, and at `prims.expandA` they are literally `keygenShort` and the signature's `verify`.
- **`nma_security_short_matrix`** — the older uniform-matrix corollary through `expandAIdealization`, kept as scaffolding and now documented as non-quantitative. That assumption has no instance below the trivial budget: the distinguisher receives both `ρ` and the candidate matrix, and recomputing `ExpandA(ρ)` is the same efficient evaluation key generation and verification perform, so a cost model does not rescue it either. Its docstring, the module status, and this list say so; `nma_security_rom` is the statement to cite.
- **The consistency witness** (`LatticeCryptoTest/MLDSA/NonVacuity.lean`): `mldsa_short_cma_hyps_inhabited` — every hypothesis of the short CMA headline is jointly inhabited, at unit slacks and a no-query adversary. It certifies logical consistency only.

## What this does not close

- **The short-model HVZK hypothesis has no witness below `ζ_zk = 1`.** In the short model the public key does not determine the withheld `t₀` across material-valid pairs, so no single simulator is exact-on-accept, and `hvzkSimulatorReal` does not transport from the seed model. At `ζ_zk = 1` the loss is at least `1`, so `euf_cma_security_of_nma_short` is a conditional statement until a quantitative short-model HVZK bound accounting for the hint mismatch is proved. That bound is a named follow-up; the seed-model headline above is the one with a proved witness.
- **Commitment-carrying, not the FIPS-204 signature format.** Both CMA headlines are stated for the standard Fiat-Shamir-with-aborts signature that carries the commitment. Reaching the recovered-commitment format costs `qS` extra hash queries (the `qH + qS` in `MLDSA.cmaToNmaLoss`); that bridge is follow-up work.
- **The tailored SelfTargetMSIS problem.** The STMSIS leg lands on `mldsaSTMSISShort` (`mldsaSTMSISMatrix` in the random-oracle model), not the standard `[I_m | A] · y` normal form; that reduction is separate follow-up work.
- **Query counts and reduction cost in the random-oracle model.** `nma_security_rom` places no bound on the forger's `ExpandA` queries and does not account for either reduction's cost; both wait on the cost-model infrastructure. Composing that leg under an EUF-CMA forger additionally needs the Fiat-Shamir-with-aborts engine to forward a third oracle, so the CMA headlines keep their abstract MLWE bridge hypothesis.
- **The legacy `euf_cma_security` sorry in `LatticeCrypto/MLDSA/Security.lean` survives this PR** (currently at line 337). This PR does not touch that file beyond the loss-formula docstrings, so the axiom baseline does not shrink when it lands; the redirect for the old theorem is tracked separately.

## Verification

On the current #507 head (Lean 4.33.1, measure-first main through #582). `lake build LatticeCrypto LatticeCryptoTest` green; `SecurityNMA.lean`, `SecurityExpandARO.lean`, and the witness elaborate clean under the mathlib standard linter set. The PMF/SPMF boundary check passes at main's ceiling for `SecurityNMA.lean`, and the new file has zero `PMF`/`SPMF` occurrences. `lake exe axiomsweep --check` passes with no new taint. `euf_cma_security_of_nma_short`, `euf_cma_security_of_nma_fips`, `euf_cma_security_of_nma_fips_hvzkReal`, `nma_security_rom`, `romNmaAdvantage_eq_game0`, and `game1_le_stmsis_rom` each depend on `[propext, Classical.choice, Quot.sound]` only.


dtumad (commented):
I reviewed the PR-specific commit (`ef1439a..16ac818d`) separately from the four stacked dependencies. The finite NMA composition itself looks coherent: the real/short PRG hop instantiates the advertised distinguisher, and the CMA theorem goes through the managed-RO/plain-NMA equality before applying the short NMA bound. The asymptotic argument also appears mathematically valid under its stated family-wise assumptions, and its note that this is not yet a PPT/cost-model theorem is useful.

There are a couple of scope distinctions that I think would be worth making sharper before treating this as the ML-DSA headline:

1. `euf_cma_security_of_nma_short` is currently a theorem about the generic `FiatShamirWithAbort` signature type `Option (Commitment × Response)`, where the commitment is carried in the signature. The published EasyCrypt Dilithium theorem first uses commitment recovery to reach that generic form; this adds one hash query per signing query, so its final loss uses `qH + qS` (and the resulting `qH + qS + 1` term). The repository already documents exactly this distinction in `MLDSA.cmaToNmaLoss` in `Security.lean`. It might be clearest either to describe this theorem consistently as the commitment-carrying/standard-FS short-model composition, or to add the commitment-recovery bridge before calling it the end-to-end ML-DSA CMA headline. The same distinction also means the present CMA theorem is not yet the FIPS-keygen theorem; `nma_security_fips` is presently NMA-only.

2. `nma_security_short_matrix` discharges the seed-to-uniform-matrix MLWE bridge, which is useful, but its SelfTargetMSIS target is still the tailored `mldsaSTMSISShort` relation. Since the PR itself correctly says that the tailored-to-standard normal-form reduction remains follow-up work, wording such as “matrix-MLWE-facing” would be more precise than “literature-facing” for the whole theorem until that second bridge exists.

For comparison, the published mechanization makes both boundaries explicit: [the final Dilithium theorem](https://github.com/formosa-crypto/dilithium/blob/main/proofs/security/Dilithium.ec) has the `qH + qS` loss, while [the simplified-scheme proof](https://github.com/formosa-crypto/dilithium/blob/main/proofs/security/SimplifiedScheme.ec) contains the generic FS-with-aborts and commitment-recovery layers.

A smaller organizational suggestion: `NonVacuity.lean` is very candid that it proves only joint logical consistency at unit slacks and a no-query adversary. That certificate is fine as stated, but it may fit better as a test/example or a follow-up module rather than a new top-level `LatticeCrypto.lean` import; separating it and the asymptotic wrapper from the finite capstone would also make the semantic core easier to review and merge. The new file also just needs the standard blank line between its copyright header and imports.

Finally, the PR is currently conflicting because of its stacked history, so I would wait for #506/#507 to settle and then review the rebased, small diff. None of the points above calls for a maintainer-side semantic rewrite; they seem better left as author-directed framing/bridging choices.

LatticeCrypto/MLDSA/SecurityNMA.lean
```diff
@@ -992,4 +2070,44 @@ The live short-secret reduction and the extraction bound are fully proven:
...
+
+variable (p : Params) (prims : Primitives p) [nttOps : NTTRingOps]
+  [DecidableEq prims.High]
+  {M : Type} [DecidableEq M] [DecidableEq (Commitment p prims)]
+  [Inhabited (Commitment p prims)] [Inhabited (Response p prims)]
+  [SampleableType (RqVec p.l)] [SampleableType (RqVec p.k)]
+  [SampleableType (TqMatrix p.k p.l)]
+  [SampleableType (CommitHashBytes p)]
+
+/-- **The matrix-MLWE-facing NMA headline.** `nma_security_short` with the abstract-problem
```
dtumad, line 2083: This is the endpoint I would hold. Its `expandAIdealization εA` hypothesis (already on `main` at `:446`) cannot hold at small `εA`: `D rho A := decide (A = prims.expandA rho)` accepts the real branch with probability exactly `1` — compiled, see the main comment. The assumption's docstring proposes reading it "against bounded distinguishers" pending #460, but recomputing `ExpandA(ρ)` is itself efficient, so that does not rescue it.

dtumad (commented):
Follow-up to my 2026-07-28 review, re-checked at `5a6ff0c6`. Both scope points are taken: `:1680` now says "commitment-carrying, not yet the end-to-end FIPS CMA headline", and `:2083` says "matrix-MLWE-facing". Those are the right boundaries and they are stated where a reader will hit them.

The composition itself holds up. The short-secret key distribution, the exact key-swap hop, the tailored STMSIS extraction, the FIPS keygen-side NMA bridge, and the numerical asymptotic closure fit together, and the notes about the asymptotic theorem quantifying numerical families rather than establishing a PPT claim are worth keeping.

One endpoint I would not ship on its current plan.

To be clear about what belongs to this PR: `expandAIdealization` is already on `main` (`SecurityNMA.lean:446`). What this PR adds is `nma_security_short_matrix`, the headline that consumes it — which is what makes it worth settling now rather than later.

The assumption's own docstring already says a distinguisher may recompute `ExpandA(ρ)` and compare, which is right. The part that does not work is the escape hatch: "pending the cost-model infrastructure (#460) it should be read computationally, against bounded distinguishers, where it is the assumption that SHAKE-based expansion yields a pseudorandom matrix." Bounding the distinguisher does not rescue it, because recomputation is exactly what a bounded distinguisher can afford — `ρ` is public and `expandA` is the same efficient function the honest verifier runs. So #460 will not change the situation.

The structural point is that this is not shaped like a pseudorandomness assumption. A PRG/XOF distinguisher is never handed the seed; here `D` gets both `ρ` and the matrix, which makes the comparison decidable:

```lean
def D (prims : Primitives p) : Bytes 32 → TqMatrix p.k p.l → ProbComp Bool :=
  fun rho A => pure (decide (A = prims.expandA rho))

theorem D_real_branch :
    Pr[= true | do
        let rho ← $ᵗ (Bytes 32)
        D prims rho (prims.expandA rho)] = 1 := by
  simp [D]

theorem expandAIdealization_forces_large (εA : ℝ) (h : expandAIdealization p prims εA) :
    1 - (Pr[= true | do
          let rho ← $ᵗ (Bytes 32)
          let A ← $ᵗ (TqMatrix p.k p.l)
          D prims rho A]).toReal ≤ εA := by
  have hD := h (D prims)
  rw [D_real_branch, ENNReal.toReal_one] at hD
  exact le_trans (le_abs_self _) hD
```

Both are axiom-clean. The real branch accepts with probability exactly one, so `εA` is forced up to `1` minus the chance a uniform matrix happens to equal `ExpandA(ρ)` — which for the actual matrix space is where you would want the *advantage* to be, not the slack.

`advantage_mldsaMLWEShort_le_matrix` and `nma_security_short_matrix` stay valid as implications; the issue is that their hypothesis cannot hold at small `εA`, so the matrix headline should not be read as a quantitative literature-facing statement. Two ways out that both seem fine to me: represent `ExpandA` by a shared/programmed random oracle, which is what "the standard random-oracle reading" would actually amount to here, and let the reduction program the table; or keep the seed-based MLWE residual as the honest endpoint and label the matrix bridge as a deliberately non-quantitative scaffold. I would not want to lose the tailored reduction scaffolding either way.

Separately, on sorry accounting: this PR is additive with respect to the legacy hole. `LatticeCrypto/MLDSA/Security.lean:366` still has its `sorry` at this head, and that file is not in the diff, so the axiom baseline will not shrink when this lands. The deletion/redirect for that old theorem is in #467. Worth being explicit about that in the description so this is not read as closing the ML-DSA headline sorry.

LatticeCrypto/MLDSA/SecurityNMA.lean
```diff
@@ -992,4 +2070,44 @@ The live short-secret reduction and the extraction bound are fully proven:
...
+bridge discharged: the MLWE leg lands on the standard uniform-matrix short-secret problem
+`mldsaMatrixMLWE`, at the cost of one application of the `expandAIdealization` assumption
+(`advantage_mldsaMLWEShort_le_matrix`, supplying the bridge slack `εA`). The SelfTargetMSIS leg
+still lands on the *tailored* `mldsaSTMSISShort`, not the standard SelfTargetMSIS normal form —
+that second bridge is follow-up work, so only the MLWE side is stated against a standard
+literature problem here. No caller-supplied inequality remains: every hypothesis is a
+satisfiable pinned equality (`keygenShort_generable`), a proven reduction, or the named XOF
+idealization. -/
+theorem nma_security_short_matrix (maxAttempts : ℕ) (εA : ℝ)
```
dtumad, line 2092: Still a valid implication — the concern is only that its hypothesis is unsatisfiable at the `εA` a reader will assume, so this should not be presented as the quantitative matrix-facing result until `ExpandA` is modelled as a shared/programmed oracle (or the seed-based residual is kept as the honest endpoint).

quangvdao (changes requested):
Thank you, @alik-eth, for assembling the capstone and for tightening the scope language after the earlier review. I re-reviewed exact head `5a6ff0c668a5acfe00ed196cf8cd26cff458ee86`, including the short-secret key swap, tailored SelfTargetMSIS extraction, FIPS-keygen NMA hop, CMA composition, asymptotic closure, and current stack/CI state.

The seed-based/tailored composition is coherent, and the file now correctly says that the CMA result is commitment-carrying rather than the end-to-end FIPS CMA theorem, and that the STMSIS endpoint remains tailored. I nevertheless need to request changes because the new matrix-facing endpoint consumes a fundamentally non-quantitative assumption.

## P1 — `expandAIdealization` is efficiently distinguishable

At `SecurityNMA.lean:512–533`, the quantified distinguisher receives both public `ρ` and the candidate matrix:

```lean
D : Bytes 32 → TqMatrix p.k p.l → ProbComp Bool
```

It can choose

```lean
fun rho A => pure (decide (A = prims.expandA rho))
```

The real branch accepts with probability exactly `1`; the independent-matrix branch accepts only when uniform `A` happens to equal `ExpandA(ρ)`. Thus any `hA : expandAIdealization ... εA` forces

`1 - Pr[A = ExpandA(ρ)] ≤ εA`,

which is essentially unit slack for the actual matrix space. This distinguisher is efficient—it runs the same public expansion as verification—so adding the planned PPT/cost model does not rescue the assumption. I independently kernel-checked this canary at the exact PR head.

The implications `advantage_mldsaMLWEShort_le_matrix` and `nma_security_short_matrix` remain logically valid, but their small-slack hypothesis is unavailable; the committed non-vacuity test proves only the trivial `εA = 1` case.

Smallest safe repair: withhold `nma_security_short_matrix` and expose an honestly labelled seed-facing corollary of `nma_security_short` using `mldsaMLWEShort`, the identity reduction, and zero bridge slack. A real matrix-facing theorem requires representing `ExpandA` by a shared/programmed random-oracle table used consistently by keygen, verification, and the reduction; it is not a local cost-bound patch.

## Other pre-merge items

- **P2 — the asymptotic theorem is numerical, not computational security.** `euf_cma_security_asymptotic_short` candidly has no cost predicate, while `hMlweBound` and `hStmsisBound` dominate every unrestricted adversary. The theorem is valid under those strong assumptions, but I suggest renaming it to something like `..._of_uniform_advantage_bounds` and describing it as numerical closure. A genuine computational headline needs a PPT predicate and reduction-preservation proofs.

- **P2 — restack and refresh the PR body.** This head is conflicting/dirty against current `main` (`3ecd5523`, Lean 4.33). The direct textual conflict is small (`have := h_transform` versus current-main `haveI := h_transform`), but the branch should be rebuilt after #507 and the already-merged #503/#504/#506 histories removed. The true delta over #507 is three files, `+1368/-3`. The current body is stale: it says "purely additive", references the old non-vacuity path, calls the matrix theorem "literature-facing", and does not disclose that the legacy `LatticeCrypto/MLDSA/Security.lean:366` `sorry` remains.

- **P3 — exact-hop wording overstates the exported result.** `nma_keyswap_hop_short` says the gap "is" the MLWE advantage, but its public statement is only `≤`. Either export the equality established by the branch identities and derive the inequality, or soften the wording.

- **P3 — the abort-extractor explanation is false.** `extractorC` says its dummy `none` output necessarily fails RO consistency / `isValid`. A forger might already have queried `(msg, default)`, and generic validity need not fail. This is harmless because the NMA abort branch is already deterministically false and extra STMSIS acceptance only helps the upper bound; the documentation/comment should say that instead.

The #507 blockers carry into this branch and are tracked on #507 rather than repeated here: the stale public loss formula, forbidden thin umbrellas, and missing producer-level mutation canary.

Validation: historical exact-head GitHub checks are green; `git diff --check` is clean; a fresh full targeted build at this exact head completed all 2,786 jobs under Lean 4.32.2; and the recomputation counterexample kernel-checked. No new `sorry`/`admit` was added, though the inherited legacy placeholder remains. A final Lean 4.33/current-main run is mandatory after restacking.

Thanks again for the substantial work. The short-model key swap and tailored extraction are valuable, and the generic CMA engine is a serious formalization achievement. I would be happy to re-review once the matrix endpoint is either honestly deferred or rebuilt around a sound shared-oracle model.

dtumad:
Deep semantic audit across the open PR set (semantics only). Extends the 08-17/08-23 threads; head `5a6ff0c` unchanged since 08-09. The modelling here is more faithful than the Formosa simplified scheme (genuine `Power2Round` split with `t₀` withheld, real `UseHint ∘ computeWApprox` verifier relation, `nma_keyswap_hop_short` as an exact monad identity, `keygenShort_generable` as a real satisfiability certificate) — worth preserving. But two hypotheses have no useful instances, which makes the composed headline vacuous as an end-to-end claim until repaired or re-scoped:

**BLOCKING 1 (confirmed from the earlier thread): `NMA.expandAIdealization` is efficiently refutable.** The distinguisher receives both `ρ` and the candidate matrix, so `D := fun rho A => pure (decide (A = prims.expandA rho))` accepts the real branch with probability 1 and the ideal branch with probability `1/|TqMatrix|` — the hypothesis holds only at `εA ≈ 1`. The head's docstring now discloses this and defers to a cost model, but a cost model does **not** rescue it: recomputing `ExpandA(ρ)` is the same efficient function verification itself runs, so the bounded-distinguisher reading is false too. `nma_security_short_matrix` therefore has no meaningful instance in any model. The repair is the one named in review: model `ExpandA` as a shared programmed random oracle used consistently by keygen, verify, and the reduction — or withhold the matrix-form theorem from this PR and land the seed-form results, which are fine.

**BLOCKING 2 (new): the composed CMA headline's `hhvzk` has no witness below `ζ_zk = 1`.** `euf_cma_security_of_nma_short` takes `hhvzk : (identificationSchemeShort p prims).HVZK sim ζ_zk`, and its own docstring concedes that in the short model no single simulator is exact-on-accept across material-valid pairs and that satisfiability is "any simulator at `ζ_zk = 1`". What the docstring does not say: main proves `idsWithAbort_hvzk_real` only for `identificationScheme` (non-Short), so today the *only* available witness for this hypothesis is `ζ_zk = 1` — at which `cmaToNmaLoss` contains `qS·1/(1−p) ≥ 1` and the headline is trivially true. This must be surfaced in the PR body and the theorem docstring as an open obligation with a named follow-up (a quantitative short-model HVZK bound accounting for the hint mismatch), not left as a satisfiability aside — otherwise the composed statement reads as an end-to-end result it is not.

**Non-blocking but required renames/disclosures:**

- `euf_cma_security_asymptotic_short` is numerical, not computational — `hMlweBound`/`hStmsisBound` dominate every unrestricted adversary, and `asymptotic_loss_regime_satisfiable` candidly instantiates numerical sequences only. Rename per the earlier suggestion (`…_of_uniform_advantage_bounds`) so nobody cites it as an asymptotic security theorem.
- The PR body still says "purely additive" and does not disclose that `LatticeCrypto/MLDSA/Security.lean`'s `euf_cma_security` sorry survives this PR. State what the PR closes and what it leaves.
- Positive note to keep as-is: `expandSReplacement`'s docstring (distinguisher never sees the seed — a genuine PRG assumption) is the model for how these assumptions should be stated; the contrast with `expandAIdealization` is exactly the point.

Coordination: consumes #507 — rebase after it lands (this PR's true delta over #507 is three files). The `1/(1−p)` ζ-term change in #507's `cmaToNmaLoss` affects the four formula lines in this PR's own ML-DSA prose as well.

alik-eth:
Rebased onto the current #507 head; both blockers are now closed at `c5908770`. The PR's own delta is `LatticeCrypto/MLDSA/SecurityNMA.lean`, the new `LatticeCrypto/MLDSA/SecurityExpandARO.lean`, `LatticeCryptoTest/MLDSA/NonVacuity.lean`, and the two library roots.

**BLOCKING 1 (`expandAIdealization`)** — agreed on the refutation, and agreed that a cost model does not rescue it: recomputing `ExpandA(ρ)` is the same efficient evaluation that key generation and verification perform. We took the first of your two ways out rather than withholding the matrix form.

`LatticeCrypto/MLDSA/SecurityExpandARO.lean` models `ExpandA` as a random oracle `Bytes 32 →ₒ TqMatrix p.k p.l`, a third oracle alongside uniform sampling and the commitment hash. It is genuinely shared: `keygenShortRO` obtains `Â` by querying it at the seed it sampled, `verifyRO` re-queries it at `pk.ρ`, and the forger queries it at will, all against one lazily simulated cache. Both reductions **program** that cache: they receive a uniform matrix from their challenger, choose the seed themselves, and pre-seed the cache with `ρ ↦ Â`, so the forger's view is exactly the honest game's.

- `romNmaAdvantage_eq_game0` — the random-oracle EUF-NMA game **is** the real branch of the uniform-matrix Module-LWE game `mldsaMatrixMLWE` against `distinguisherBRom`. An exact identity, not a bound: the honest generator's first query at a fresh seed draws a uniform matrix and caches it, which is the programmed cache at the challenger's matrix.
- `game1_le_stmsis_rom` — the uniform branch is bounded by the SelfTargetMSIS advantage of `extractorCRom` against `mldsaSTMSISMatrix` (uniform matrix, uniform `t`, ML-DSA verifier relation).
- `nma_security_rom` — `Adv^{EUF-NMA}(A) ≤ Adv^{MLWE}(B) + Adv^{SelfTargetMSIS}(C)` over the uniform-matrix problem, with no idealization slack and no hypothesis beyond the forger. Axiom-clean on the standard three.

To keep the model honest about being the same scheme, the oracle algorithms are pinned to the scheme's own: answering `ExpandA` from a fixed table `f` recovers `keygenShortWith f` and `verifyAtMatrix (f pk.ρ)`, and at `f := prims.expandA` those are literally `keygenShort` and the signature's `verify` (`simulateQ_tableImpl_expandA_keygenShortRO`, `simulateQ_tableImpl_expandA_verifyRO`). The only difference between the random-oracle game and the scheme's game is that `prims.expandA` is replaced by a lazily sampled random function.

`nma_security_short_matrix` and `expandAIdealization` stay as scaffolding, with the emptiness written into the assumption's docstring, the corollary's docstring, the module status section, and the PR body, pointing at `nma_security_rom` as the statement to cite. Not modelled and stated as such: the forger's `ExpandA` query count and both reductions' cost, which wait on the cost model; and the SelfTargetMSIS leg still lands on the tailored verifier relation, not the `[I_m | A] · y` normal form.

**BLOCKING 2 (`hhvzk`)** — two things. The obligation is now stated as such in the theorem docstring, the module status, and the PR body: no witness below `ζ_zk = 1`, why (`t₀` is not determined by the public key across material-valid pairs, so `hvzkSimulatorReal` does not transport from the seed model), and what would discharge it (a quantitative short-model HVZK bound accounting for the hint mismatch), named as a follow-up. And there is now a composed headline that does not have the problem: `euf_cma_security_of_nma_fips` runs the same three-step composition over the seed-tagged `identificationScheme` with `nma_security_fips` as the NMA leg, and `euf_cma_security_of_nma_fips_hvzkReal` discharges its HVZK hypothesis with the proved simulator from `main` (`idsWithAbort_hvzk_real`, at `hvzkBoundReal`, under `Primitives.Laws`; `mldsa_laws_inhabited` in `Extern/MLDSA/NonVacuity.lean` is the approved-parameter witness). The price relative to the short model is the `εPRG` term of the `ExpandS` hop; the only simulator-side hypothesis left is its abort rate on good keys. Both are axiom-clean on the standard three.

**Renames and disclosures** — `euf_cma_security_asymptotic_short` is now `euf_cma_security_short_of_uniform_advantage_bounds`, with its docstring and section header saying it is a numerical closure, not computational security. The PR body no longer says "purely additive" and states that the legacy `euf_cma_security` sorry in `LatticeCrypto/MLDSA/Security.lean` (now at line 337) survives this PR.

Migration notes for the record: the merge onto #507 needed only main's side on three hunks; #508's own additions had four `SPMF.evalDist_def` rewrites (now `evalSPMF_id`, which also keeps the file at main's PMF ceiling) and the `SampleableType (RqVec …)` binders that #582 made generic. The new file has zero `PMF`/`SPMF` occurrences; `lake exe axiomsweep --check` passes with no new taint.

alik-eth:
@quangvdao Re-merged on the updated #507 and current main, head dddcd578. Changes since your last review:

- The NMA and CMA headlines no longer conclude `∃ mlweReduction stmsisReduction`, which main's #707 treats as vacuous. The MLWE bridge is a supplied reduction `bridge` with advantage bound `hMlweBridge`, and each bound names `bridge` and `extractorCShort` at the forging strategy: `adv.main`, or in the CMA forms the strategy of `simulatedEufNmaAdv`. The abstract `stmsis`/`hStmsis` pair is gone because the equation already fixed it. `nma_security_short_matrix` names `matrixLift ∘ distinguisherBShort`, and `nma_security_rom` names `distinguisherBRom`/`extractorCRom`.
- The headlines move to `LatticeCrypto/MLDSA/SecurityHeadlines.lean`, keeping `SecurityNMA.lean` under 1500 lines. `idsWithAbort_hvzk_real` is now main's `idsWithAbort_hvzk`.
- Dead `DecidableEq (Commitment …)` and `SampleableType (PublicKey …)` binders are removed, which drops the `instCommEq` and `instPk` hypotheses.

Build (including LatticeCryptoTest), style lint and the axiom sweep pass locally. The Lint job fails only on `usesRetiredProbability` (14 declarations here, plus #507's); the policy question on that is open with dtumad on #478.

PR #510, opened 2026-08-05
Migrate to Lean modules and update dependencies to v4.32.2

dtumad:
## Summary

- migrate the active Lean libraries and tests to the Lean module system
- use `public import`, implementation-only imports, `public meta import`, `public section`, and `public meta section` according to each dependency and declaration surface
- preserve proof-facing definitional compatibility with `@[expose] public section` during the migration, while keeping executable/runtime implementation bodies opaque
- keep six migration-only proof helpers private; explicit `by exact` proof blocks retain the exposed computational definitions without exporting those proof names
- make umbrella modules module-aware and add CI checks that generated umbrellas remain complete
- update Lean, Mathlib, and PolyFun to the stable `v4.32.2` release line and refresh the Lake manifest

## Dependency scope

- Lean toolchain: `v4.32.2`
- Mathlib: `v4.32.2`
- PolyFun: `v4.32.2`
- Loom2 remains pinned to the current compatible `lean-4.32` revision
- native C sources and native submodule gitlinks are unchanged
- dormant Interop dependencies remain unchanged

The Falcon test executable helpers are public but opaque. This allows its public `main` declaration to survive native code generation without exposing implementation bodies.

## Follow-up boundary

- broader public-import narrowing is deferred until it can be tested as an incremental compatibility change
- computational helper visibility and finer-grained exposure are deferred for a downstream-focused audit, with ArkLib and secure-messaging as the primary canaries
- no one-time module-cutover script remains in the repository

## Validation

- direct Lean compilation of all six proof-privacy modules
- `lake build`
- `lake env lean VCVioTest/Smoke.lean`
- generated-umbrella checks for all active libraries and tests
- Extern-isolation and Interop-isolation checks
- clean cached production build of `ToMathlib VCVio LatticeCrypto Extern HashSig Examples VCVioWidgets`
- repository warning-budget check: zero non-`sorry` Lean warnings
- `lake build VCVioTest LatticeCryptoTest HashSigTest`
- VCVio smoke, round-by-round, and Merkle batch modules
- ML-KEM native tests: 175 passed, 0 failed
- ML-DSA native tests: 128 passed, 0 failed
- Falcon native tests: 201 passed, 0 failed
- both SLH-DSA KAT executables
- generated-umbrella, style, Extern-isolation, and Interop-isolation checks
- shell syntax and `git diff --check`
- fresh downstream Lake project importing `VCVio` on Lean `v4.32.2`


dtumad:
cc @alexanderlhicks @BoltonBailey

BoltonBailey:
Thanks!

One thing I've noticed tends to get thrashed in these upgrades is private definitions, which lose the keyword because private now has different meaning, IDK if we want to make any attempt to save these or mark a TODO about considering their status.

dtumad:
> Thanks!
> 
> One thing I've noticed tends to get thrashed in these upgrades is private definitions, which lose the keyword because private now has different meaning, IDK if we want to make any attempt to save these or mark a TODO about considering their status.

#511 restores a bit more privacy for some testing defs, but I think it makes sense to do this iteratively through future PRs. The main uncertainty I guess is that non-reducible definitions in `VCVio` make `PolyFun` machinery insufficient, and I think iterative steps with `ArkLib` migration are the most natural testing ground for that.

dtumad:
One open question might be if it is worth just getting rid of `OracleComp spec A` for `PFunctor.FreeM ↑spec A` (or at least the namespace if not the `abbrev`), as then everything in that space reduces to module scoping of `PolyFun` itself.

quangvdao:
### Falcon build timing

I profiled `LatticeCryptoTest/Falcon/Main.lean` at the previous PR head. Direct compilation took 63.62 seconds. With `-Dtrace.profiler=true`, elaboration took 1.68 seconds, while `Lean.compileDecls` took 104.09 seconds under the profiler. The first compiler `simp` pass used 103.07 seconds. The cause was the 700 line `main` declaration.

Commit `0cc445b9` splits the runner into four opaque declarations. `main` calls them in the original order. This preserves every test, the output order, and failure behavior.

Local results:

* Direct compilation took 5.56 seconds, down from 63.62 seconds.
* Lake compiled the Falcon test module in 3.4 to 4.5 seconds.
* A full incremental `lake build falcon_test`, including native compilation and linking after the declaration change, took 10.31 seconds.
* `.lake/build/bin/falcon_test` reported 201 passed and 0 failed.
* `git diff --check` passed.

This removes the Falcon compile bottleneck. Adding this module to CI should now cost about 5 seconds of Lean compilation on this machine instead of about one minute.

alexanderlhicks:
> One open question might be if it is worth just getting rid of `OracleComp spec A` for `PFunctor.FreeM ↑spec A` (or at least the namespace if not the `abbrev`), as then everything in that space reduces to module scoping of `PolyFun` itself.

I can see this making sense, it would be nice to preserve some cryptographer-friendly naming/syntax but beyond that the cleaner/simpler it is go all the way down to PolyFun is bets imo. That said, I haven't reviewed all the recent work in PolyFun so I'm really just speaking at a high level and might be ignoring some practical nuances.

alexanderlhicks:
> I think iterative steps with ArkLib migration are the most natural testing ground for that.

Yep. I'm merging my split of the abf26 PRs before bumping ArkLib to 4.32.2 (so that I can then pin the proximity prize to that release) and we can do that then.

quangvdao:
### Downstream consumer CI fix

The slow job was not caused by the no submodule stubs. PR 510 changed the scratch executable to import the full `VCVio` umbrella while this clean job intentionally disabled repository and Mathlib caches. That caused a second native build with about 6,000 dependency targets. The previous run spent about 48 minutes in that step before the new push canceled it.

Commit `a03dc825` restores a clear responsibility split:

* The downstream job builds a minimal module style executable with no VCVio import. Lake still links the VCVio dependency `extern_lib` targets, so it still covers the missing submodule failure from issue 495.
* `VCVioTest/Smoke.lean` already imports the full VCVio umbrella after the normal project build. It now contains `#check OracleComp`, so the public module export assertion remains covered without a second native build.

Local clean reproduction:

* All three native submodules were uninitialized.
* `lake update` completed in 68.92 seconds with the unnecessary Mathlib cache download disabled.
* `lake build` built 12 targets in 3.05 seconds. It created empty stubs for `leanhashing`, `leanmlkem`, `leanmldsa`, and `leanfalcon`.
* The scratch executable printed `consumer ok`.
* `lake build VCVio` passed, and `lake env lean VCVioTest/Smoke.lean` resolved `OracleComp`.

GitHub cache remains disabled in the downstream job, so stale native artifacts cannot hide a missing submodule regression.

GitHub confirmation: the [first downstream run after the fix](https://github.com/Verified-zkEVM/VCVio/actions/runs/31286312694/job/93175724194) passed in 67 seconds total.

quangvdao:
Looks good for me to merge

alexanderlhicks (approved):

PR #513, opened 2026-08-10
feat(scripts): kernel-level axiom sweep with committed regression baseline

alexanderlhicks:
## What

Adds `lake exe axiomsweep`: kernel-level axiom / `sorry` accounting for the whole library, with a committed regression baseline.

The tool loads the built `.olean` environment (the same data the kernel checked) and computes, for every declaration under the swept root modules, the set of axioms it transitively depends on — i.e. the `#print axioms` information, library-wide, in one pass. Because it reads elaborated data rather than source text, it sees exactly what the kernel accepted: private declarations, macro-generated declarations, and instances included; no regex heuristics.

## Why

VCV-io already treats trust boundaries as first-class (the Interop TCB-isolation gate, the no-submodules consumer job), but had no kernel-level accounting of what its theorems actually rest on. The sweep's first run surfaced a real, previously-invisible dependency: **20 declarations in the concrete ML-DSA/ML-KEM path rest on `native_decide` trust axioms** (`Lean.ofReduceBool`), inherited through two *private* NTT matrix lemmas that no source scan or manual review would flag. Downstream reviews (ArkLib) have also repeatedly traced `sorryAx` inheritance chains by hand; this makes both mechanical.

Scope note: `Interop` (the declared TCB) is deliberately excluded from the default sweep — its boundary is enforced by the import-isolation gate; this tool accounts for the axioms everything *outside* it rests on. Sweep it explicitly with `--root Interop` if desired.

## How

- **`scripts/AxiomSweep.lean`** (+ `lean_exe axiomsweep` in the lakefile): memoised traversal of the compiled environment; only imports core `Lean`, runtime-imports the built oleans, so it needs a completed `lake build` and adds no dependencies. Runs in seconds.
- **`scripts/axiom_baseline.json`**: the currently-known `sorryAx`-tainted declarations (and any non-standard-axiom dependents). `--check` fails **iff a declaration is tainted that the baseline does not list**; closing gaps stays green and prints a nudge to shrink the baseline via `--update-baseline`.
- **CI**: one report-only step after the build (`::warning` on regression). Flip to enforcing by deleting the `|| echo` once the baseline has soaked for a few PRs.
- Docs updated in the same PR.

## Modes

```
lake exe axiomsweep                     # summary
lake exe axiomsweep --out report.json   # full per-declaration report (name, module, kind, line, axioms)
lake exe axiomsweep --check             # regression gate against scripts/axiom_baseline.json
lake exe axiomsweep --update-baseline   # refresh the baseline; commit the diff
lake exe axiomsweep --root Foo          # override the swept root modules (repeatable)
```

## Baseline at this commit

- **13,985 declarations across 438 modules** in the seven non-test libraries; sweep runs in seconds on a built tree.
- **51 sorryAx-tainted declarations** (honest WIP, now pinned by name).
- **20 declarations resting on `native_decide` trust axioms** — all inherited from `_private.…MLDSA/MLKEM.Concrete.invNTTMatrix_nttMatrix_entry._native.native_decide.ax_*`. Recorded in the baseline; it doubles as the burndown list.

### Burndown feasibility (measured)

Kernel-checking these two lemmas was probed empirically before this PR:
- The loop kernels (`loopNTT`/`loopInvNTT`) use `while` loops → `Loop.forIn` → `partial`, so the kernel **cannot unfold them in principle** — this, not speed, is why the in-file comment says `decide`/`rfl` gets stuck. Any burndown needs a structural (e.g. `Vector.ofFn`/`Fin.foldl`) re-derivation of the transform matrices on the certificate side, keeping the loop kernels via the existing `@[implemented_by]` pattern.
- The dense 256×256×256 product identity costs ~180µs per modular multiply-add in the kernel (measured: 1M ops = 182 s) → **~49 min naive**, optimistically 5–15 min with an engineered Nat-packed certificate + sharded per-row lemmas. A real project, not a cleanup.
- The durable fix is the file comment's own option 2: prove the kernels implement the Vandermonde transform and derive inversion from ~256 geometric-sum identities (`∑ ζ^{jk} = 0`) instead of 16.7M multiplications.

Proposed as a follow-up issue; this PR only makes the trust dependency visible and gates against new ones.
- Spot-checked both directions against `#print axioms` (`MLDSA.Concrete.concreteNTTRingLaws` → native trust axiom present; `BR93.br93AsymmEnc.indcpa_bound` → `sorryAx` present); `--check` exit codes verified.
- Works fine with the module system (`module`/`public` sources).

## Notes

- Same tool as ArkLib's `alh/axiom-sweep` branch (multi-root variant); the copies are intentionally self-contained per repo, mirroring how other scripts are shared across the org. If it proves its keep everywhere, extracting a shared home is a possible follow-up.
- `lean4export` was considered and deliberately not used: it targets external kernel checking, has no source positions, and would add a per-toolchain pinned dependency; walking the environment yields the same information with line numbers and zero new deps.
- The per-declaration report is not committed anywhere; only the baseline is tracked.

## Adversarial review

Five independent adversarial reviewers (one per repo deployment) attacked the tool, wiring, and baselines before this PR was opened. Everything they confirmed is fixed in the follow-up commit on this branch:

- **Collector defect (HIGH, confirmed in 4 repos)**: the one-pass DFS finalized self-referencing constants (every inductive/ctor pair) prematurely, producing per-declaration rows that diverged from `#print axioms` in both directions, with constructible shapes where a new `sorry` passed `--check`. Fixed with a fixpoint-repair phase computing the true kernel closure (now strictly *more* accurate than `#print axioms`, whose own `CollectAxioms` wobbles inside mutual families). Post-fix rows spot-verified against ground truth; sorry-sets unchanged everywhere.
- Axiom *types* now traversed (`CollectAxioms` parity); duplicate `constNames` rows deduped; `native_decide` axiom names normalized to their owning declaration (the `ax_N_M` counters are `Elab.async`/toolchain-volatile); `--check`/`--update-baseline` mutually exclusive; bare `Lean.ofReduceBool`/`Lean.trustCompiler` are never baselinable (floor).
- CI steps now distinguish taint findings (report-only during soak) from infrastructure failures (always fail).
- Known blind spots documented: structure-field defaults and `example`s never enter any environment walk (`#print axioms` included); unimported files are invisible — paired with each repo's import-completeness gate.

What survived attack unchanged: baseline determinism (byte-identical regeneration), all headline numbers, spot-checks vs `#print axioms`, gate exit-code semantics, and the CI/step placement.

### DIFF
--- a/scripts/AxiomSweep.lean
+++ b/scripts/AxiomSweep.lean
@@ -44,10 +44,19 @@ The committed baseline (`scripts/axiom_baseline.json`) records the currently-kno
 `sorryAx`-tainted declarations and any declarations depending on non-standard axioms
 (anything beyond `propext`, `Classical.choice`, `Quot.sound` — so native trust axioms
 surface here too: `native_decide`-style tactics mint per-declaration
-`…._native.<tactic>.ax_*` axioms, recorded under their owning declaration). `--check`
-fails exactly when a declaration is tainted that the baseline does not cover. When gaps
-are closed, `--check` reports them and stays green; run `--update-baseline` to shrink the
-file in the same PR.
+`…._native.<tactic>.ax_<number>_<number>` axioms, recorded under their owning
+declaration). `--check` fails exactly when a declaration is tainted that the baseline
+does not cover. When gaps are closed, `--check` reports them and stays green; run
+`--update-baseline` to shrink the file in the same PR.
+
+The baseline is an allowlist for `sorryAx` debt only. Native trust is held to PolyFun's
+zero-debt rule instead: `neverAllowlistable` rejects any `._native.` axiom outside the
+explicit `grandfatheredNativeTrust` list, so no baseline edit can widen the trusted
+computing base. Exit codes are a contract with CI — `1` is a taint verdict, anything else
+an infrastructure failure — which is why `main` traps uncaught exceptions into `2`.
+
+`scripts/test-axiomsweep.sh` exercises all of this against the `AxiomSweepTestFixtures`
+library, whose fixtures carry synthetic taint of each shape this file reasons about.
 -/
 
 open Lean
@@ -65,11 +74,35 @@ def defaultRoots : Array Name :=
 /-- Axioms that carry no extra trust assumptions beyond Lean's standard foundation. -/
 def standardAxioms : List Name := [``propext, ``Classical.choice, ``Quot.sound]
 
-/-- Axioms that may never be baselined: bare native-compiler trust. A baseline edit
-cannot green these — remove the dependency instead. (Zero hits today; this floor keeps
-the baseline from ever becoming a second, laxer policy.) -/
+/-- Whether `a` names native-compiler trust: either a bare compiler axiom, or one of the
+per-declaration axioms `native_decide`-style tactics mint, which this toolchain names
+`Owner._native.<tactic>.ax_<n>_<n>` rather than routing through `Lean.ofReduceBool`. -/
+def isNativeTrust (a : String) : Bool :=
+  a == "Lean.ofReduceBool" || a == "Lean.trustCompiler" || (a.splitOn "._native.").length > 1
+
+/-- The native trust this repository has already accepted, listed by full axiom name so
+the acceptance is auditable in source rather than hidden in a JSON allowlist. Both come
+from the ML-DSA / ML-KEM NTT inversion certificates, whose loop kernels the kernel cannot
+unfold in principle (`while` → `Loop.forIn` → `partial`); see the burndown note in
+`LatticeCrypto/{MLDSA,MLKEM}/Concrete/NTT.lean`.
+
+Fails closed: if a private-name index or module path shifts, the entry stops matching and
+`--check` goes red until someone consciously re-accepts it. -/
+def grandfatheredNativeTrust : List String :=
+  ["_private.LatticeCrypto.MLDSA.Concrete.NTT.0.MLDSA.Concrete.invNTTMatrix_nttMatrix_entry._native.native_decide",
+   "_private.LatticeCrypto.MLKEM.Concrete.NTT.0.MLKEM.Concrete.invNTTMatrix_nttMatrix_entry._native.native_decide"]
+
+/-- Axioms that may never be baselined: native-compiler trust beyond what
+`grandfatheredNativeTrust` already accepts. Unlike `sorryAx` debt — honest work in
+progress, which the baseline is allowed to track — new native trust is a widening of the
+trusted computing base, so no baseline edit can green it; remove the dependency instead.
+
+This is the counterpart of PolyFun's zero-debt gate. PolyFun refuses a nonempty baseline
+outright; VCVio cannot, since it carries genuine `sorryAx` debt, so the floor applies the
+same "cannot pre-authorize future taint" rule to the part of the baseline where widening
+the TCB is at stake. -/
 def neverAllowlistable (a : String) : Bool :=
-  a == "Lean.ofReduceBool" || a == "Lean.trustCompiler"
+  isNativeTrust a && !grandfatheredNativeTrust.contains a
 
 /-- Phase 1: DFS. Compute, for every constant reachable from the work list, an
 under-approximation of the set of axioms it transitively depends on, memoised across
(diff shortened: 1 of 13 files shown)


dtumad:
I pushed a commit to this branch (`77edbe29`) mirroring Quang's follow-up on the PolyFun side. Context, since it lands on top of your work:

PolyFun#127 got one more commit after the copy here branched — [`e2be6869`](https://github.com/Verified-zkEVM/PolyFun/commit/e2be6869), "enforce zero-debt gate" — which upgraded the shared tool. Reviewing this PR I hit three of the same things independently, so rather than write them up I ported the fix:

- **The `neverAllowlistable` floor never fired.** It matched only `Lean.ofReduceBool` / `Lean.trustCompiler`, but this toolchain mints per-declaration `…invNTTMatrix_nttMatrix_entry._native.native_decide` axioms instead. So "zero hits today" was the predicate not matching the form native trust actually takes here, and all 20 native-trust declarations were greened by the baseline with `--check` passing. Quang deleted the floor upstream because PolyFun refuses a nonempty baseline outright; VCVio cannot do that with 51 honest sorries, so the split is now: baseline stays an allowlist for `sorryAx`, native trust gets the zero-debt rule via an in-source `grandfatheredNativeTrust` list. Widening the TCB can no longer be done by editing JSON.

- **`normalizeAxiomName` was collapsing too eagerly.** Any `X._native.Y....` became `X._native.Y`, so two distinct axioms could share one baseline key. Ported the upstream shape validation (`isDecimal` on the `ax_<n>_<n>` suffix, names that merely contain `._native.` preserved).

- **The fixture harness had not been ported.** Upstream ships `scripts/test-axiomsweep.sh` plus six fixtures; that is now here. The one that matters most is `Tainted/Mutual.lean`: `MutualRight` inherits its axiom only across the mutual-inductive cycle, which is the sole witness for the `repair` fixpoint pass — nothing in the swept libraries exercises that path, so the docstring's "strictly more accurate than `#print axioms`" had no test behind it. I checked the ported tests have teeth by reverting `normalizeAxiomName` to the loose form; the matrix fails, and passes again on restore.

One more, not from upstream: `main` returned exit **1** on an uncaught exception, colliding with the taint code that the CI step deliberately treats as report-only. Repro was `lake exe axiomsweep --out /nonexistent-dir/r.json`. `main` now traps into `2`.

I also flipped the CI step from report-only to enforcing, matching what Quang did upstream — `--check` is green on the real library today, and with the exit-code collision fixed the soak rationale is weaker. Happy to put the wrapper back if you would rather let it soak.

Separately, on the tool itself: I cross-checked it against `#print axioms` on 172 declarations (all 52 referenceable tainted ones plus 120 random clean theorems) and found **zero semantic disagreements** — the only differences are the sweep printing `_private.`-mangled axiom names where `#print axioms` de-mangles. The 20 `native_decide` declarations it surfaced are a real finding that nothing else in the repo would have caught.

alexanderlhicks:
Nice,  can you PR this to ArkLib / CompPoly too?

dtumad:
> Nice, can you PR this to ArkLib / CompPoly too?

Yeah i can take care of that (note these changes are mostly just porting work @quangvdao did in PolyFun)

scripts/AxiomSweep.lean
```diff
@@ -0,0 +1,426 @@
...
+  if !newNonstd.isEmpty then
+    failed := true
+    IO.eprintln s!"axiomsweep: {newNonstd.size} declaration(s) newly depend on \
+      non-standard axioms (not in {basePath}):"
+    for e in newNonstd do IO.eprintln s!"  {e.name} : {e.axioms}"
+  if failed then
+    IO.eprintln s!"axiomsweep: if intentional (new tagged sorry), refresh the baseline \
+      with `lake exe axiomsweep --update-baseline` and commit the diff."
+    return 1
+  if !fixedSorry.isEmpty || !fixedNonstd.isEmpty then
```
dtumad, line 329: This is the gap: a baseline row whose declaration is now clean prints advice and then returns `0`, so the row survives — and while it survives, the same declaration can regain `sorryAx` without failing the check. Making this branch a failure (same "refresh the baseline and commit the diff" instruction as above) would turn the baseline into exactly the current debt instead of an upper bound on it.

scripts/AxiomSweep.lean
```diff
@@ -0,0 +1,426 @@
...
+lake exe axiomsweep --check             # gate against scripts/axiom_baseline.json
+lake exe axiomsweep --update-baseline   # rewrite the baseline from the current build
+```
+
+The committed baseline (`scripts/axiom_baseline.json`) records the currently-known
+`sorryAx`-tainted declarations and any declarations depending on non-standard axioms
+(anything beyond `propext`, `Classical.choice`, `Quot.sound` — so native trust axioms
+surface here too: `native_decide`-style tactics mint per-declaration
+`…._native.<tactic>.ax_<number>_<number>` axioms, recorded under their owning
+declaration). `--check` fails exactly when a declaration is tainted that the baseline
```
dtumad, line 48: Worth being explicit here that the converse is not checked: a baseline entry that is no longer tainted does not fail, so the file can drift above the real debt. That is the property a "PR X closed this sorry" claim needs.

dtumad (commented):
I used this while auditing the large open PRs, and it did the job — being able to ask for a declaration's transitive taint at a specific head is exactly what makes "does this PR actually close a sorry" answerable. Two things I especially like: holding native trust to an explicit `grandfatheredNativeTrust` floor that no baseline edit can green, and `runUpdate` refusing to write a baseline that `--check` would reject anyway.

One gap in the ratchet, which matters precisely because this is meant to be the gate.

`runCheck` fails on `newSorry` and `newNonstd`, but a baseline row whose declaration is *no longer* tainted only produces the "good news — run `--update-baseline` to shrink the baseline" message and then `return 0`. So a stale row is accepted indefinitely, and while it sits there the gate will not notice a `sorry` reappearing under that same declaration name. That is the one direction where the ratchet does not hold: the baseline is supposed to be a non-increasing debt allowlist, but nothing forces it to shrink when the debt is paid, and an unshrunk entry is a standing licence to re-incur it.

Concretely this bites the PRs I was auditing. Several of them claim to close specific holes, and the way to hold them to it is that the baseline diff has to show the exact shrink in the same PR. If `fixedSorry` is only advisory, a PR can claim a closure, leave the row, and a later change can silently reintroduce the same taint under the same name with a green gate.

I would make `fixedSorry` / `fixedNonstd` non-empty a failure, with the same "refresh the baseline and commit the diff" instruction that the regression branch already prints — it is the same fix either way, and it makes the committed baseline exactly the current debt rather than an upper bound on it. If there is a reason to keep it advisory (say, transient taint differences across configurations), a flag to make it strict for CI would work too.

One smaller point worth recording somewhere in the module docstring: the sweep sees the compiled environment, so it does not cover `example`s, structure-default or autoparam holes, or declarations that nothing imports, and the build filters "declaration uses sorry" warnings. That makes the sweep the gate rather than the build — which is the right design, but it means a source-level scan is still a complementary check rather than a redundant one.

dtumad:
Closing in favor of #522, which merged carrying this PR's implementation and authorship: the branch's two commits (the sweep tool and the adversarial-review hardening) are the first two commits of #522's history, rebased unchanged, with the fixture matrix and namespacing added on top. Thanks for building this — the tool and its exit-code contract are exactly what landed.

PR #514, opened 2026-08-12
feat(Falcon): decode FPR to ℝ, and prove all six of its error bounds

alik-eth:
The `FPR` denotation went through `Float.ofBits` and `Float.toRat0`, whose chain bottoms out on runtime primitives the kernel cannot evaluate. Nothing about `toReal` was provable, and the five error bounds below it had sat `sorry` since.

This decodes the IEEE-754 fields from the word instead. `Float` no longer appears in the file, and `toReal` reduces in the kernel.

## That made the bounds refutable — and all five were false

- **`add_error`, `mul_error`** — two largest-finite normals overflow to the non-finite exponent, which denotes zero, so the error is the entire magnitude.
- **`div_error`** — overflow wraps to a *negative*, small, finite value.
- **`sqrt_error`** — needs no overflow at all: at `+∞` the hypothesis `0 ≤ toReal a` holds and the right-hand side is `0`, while `sqrt` returns `2^512`.
- **`expm_p63_error`** — constrained `x` but said nothing about `ccs`. At `ccs = 1` the fixed-point conversion wraps and the routine returns `0` against a true value of `exp(-x)`; above `1` the claim fails outright.

Each is now restricted to the domain where it holds. **Every conclusion is unchanged** — same `2^-52`, same `2^-51`, same shape. In-file witnesses show the hypotheses admit ordinary values, and that the result-range condition is *not* implied by the operand conditions.

## All six are now proved

`add_error`, `sub_error`, `mul_error`, `div_error`, `sqrt_error`, `expm_p63_error` — every one axiom-clean `[propext, Classical.choice, Quot.sound]`, no `bv_decide`, no `native_decide`. **`Extern/Falcon/FPRBridge.lean` carries no `sorry`.**

Each kernel is modelled as a pipeline record pinned to the kernel term by `rfl`, so the proofs are about the code that runs rather than a restatement of it. `add` and `mul` are straight-line. `div` and `sqrt` are not: each runs a fixed-length `for _ in [0:n]` loop, and neither body reads the loop index, so the fold is iteration of one step function — core's `forIn_eq_forIn_range'` reduces it to a `List.range'` fold, induction gives `forIn = f^[n]`, and a loop invariant becomes induction on the step count.

- **`div`** is exact restoring division: `yu·q + r = 2^55·xu`, remainder below `2·yu`, over 55 iterations. The step count has to be bounded — past 62 the quotient's doubling overflows `UInt64`.
- **`sqrt`** is a digit recurrence computing an exact integer square root: `q² ≤ 2^54·xu < (q+1)²`. Its bit weight hits zero on the last iteration, so the invariant holds for 53 steps and is extended by one; that final step is what turns it into a floor. Both exponent parities then collapse to a single formula, because the odd case's doubling of the significand exactly compensates the floor in the halved exponent.

`mul_error_combine` assembles `mul`, `div` and `sqrt` alike. Writing it abstractly rather than inlining it into `mul_error` is what let each later operation reuse it unchanged.

## Two places where the obvious bound was true but too loose

Both would have overrun the budget once composed with the `2^-53` rounding step, and both needed a sharper fact rather than more care:

- Sticky shifts **compose**, so `mul`'s two folds cost one ulp, not two.
- A **one-bit** sticky shift moves a value by one unit, not the two the generic bracket allows.

Numbers first, lemmas second: measured against exact rational arithmetic, `div`'s end-to-end error peaks at **0.498 of the `2^-52` budget** over 150k in-domain pairs. That is what said a sharper fact had to exist.

## The sixth: certifying a minimax polynomial

`expm_p63_error` is a different kind of obligation, and it lands in a new file, `Extern/Falcon/ExpmBridge.lean`. Its `2^-51` is very nearly saturated, so the budget has to be spent carefully. In units of `2^-63`, against 4096:

| | units |
|---|---|
| fixed-point pipeline | 10 |
| 15 Chebyshev certificates | 3574 |
| Taylor truncation, degree 19 | 80 |
| **total** | **3664** |

The fixed-point half is ordinary: `mulHi64` is the high half of the exact product, `mtwop63` is *exactly* `⌊2^63 · toReal x⌋` with no rounding of its own, and the twelve Horner floors contract because each step scales the previous error by `ζ < 0.694`. The loop's `UInt64` subtraction cannot wrap for a pleasingly cheap reason: `mulHi64 a b ≤ b` for **every** `a`.

The certification half is where the difficulty sits. `facctCoeffs` is a minimax fit, not a Taylor truncation, so `P − T` carries `O(1)` coefficients while being `O(2^-51)` — and any bound applying a triangle inequality to those coefficients discards exactly the cancellation the fit relies on. Measured: coefficient-wise against Taylor misses by ~38000×, and midpoint-plus-Lipschitz still misses by 730× at 256 subintervals, because bounding `sup |Q′|` coefficient-wise has the identical pathology. Subdividing restores locality; 15 Chebyshev certificates over **dyadic** breakpoints land at 3574. Dyadic matters: anchoring the split at a decimal bound for `log 2` produces 337-digit numerals, and the line-length linter cannot wrap a numeral.

The whole certification layer costs ~25s of build time, and `docs/agents/expm-certification.md` records the measurements behind every choice — including the three estimates that were wrong until checked.

## Companion PR: the sampler fix

The `berExp` argument-reduction fix that motivated part of this work has moved to #517 at @dtumad's suggestion — it is a runtime conformance change and does not belong in an error-bounds PR. What stays here is the bounds work it prompted but does not depend on:

- `expm_p63_error`'s domain at `x ≤ 694/1000` rather than `x < log 2`. Any caller reducing modulo `log 2` by rounding a floating-point quotient can land a few ulps **above** `log 2`, so a statement closed at `log 2` is unusable in practice. `0.694` is the contraction factor the Horner induction already ran on, and the certificates reach `89/128`, so this cost no constant churn.
- `mtwop63_neg`, `expm_p63_neg`, `expm_p63_error_abs` — the routine reads only its argument's magnitude, so it computes `exp(-|x|)`. That is proved here rather than asserted.

## Scope: the class contract, and the instance

@dtumad found that the `_valid` fields this PR adds were **unsatisfiable as specified**: with
`Valid := FPR.IsNormal`, `add_valid` is refutable at `1 + (-1)`, since the exact sum `0` is in
range by `InNormalMagnitudeRange`'s own `r = 0` disjunct while `FPR.add` returns `+0`. The open
item was the contract, not five missing proofs.

Repaired with `FPR.IsNormalOrZero` — normal and finite, or exactly `±0`, the domain binary64
arithmetic is genuinely closed under, still excluding subnormals and the non-finite encodings.
All six closure properties are proved against it, and every `_error` bound is now stated on the
same domain, so **the instance is live rather than commented out**. Getting there needed each
pipeline traced at an operand whose exponent field is `0` — exact cases, not approximate ones:

- `FPR.add` reassembles the other operand unchanged. Nothing is aligned in (`yu` vanishes), the
  renormalising count is forced to `8`, and the nine-bit sticky fold divides exactly, so the
  packing round-trips. One trace covers both argument orders, by stating it on the pipeline's
  ordered `x'`.
- `FPR.mul` and `FPR.div` carry an explicit flush guard that fires on a zero exponent field,
  clearing the significand *and* substituting the exponent `-1076`, which `make` packs as a bare
  sign bit.
- `FPR.sqrt` masks its significand against the same test, and `make_z` collapses on it.

`FPR.add` is not bit-preserving there — `(+0) + (-0)` is `+0` — so the widened statements are
phrased on `toReal` and on the decoded fields, never on the bit pattern.

**The contract also needed a non-degeneracy clause**, which is a defect this PR introduced and
nobody had flagged. Carrying a domain is what makes the class satisfiable at all, but it also
makes it satisfiable *for nothing*: `Valid := fun _ => False` discharges every `_error` and
`_valid` field vacuously, leaving only `interp_zero` / `interp_one` / `neg_exact` with content.
I built that instance and it is axiom-clean, so `[HasRealSemantics F ε]` alone was an assumption
that said nothing, and `compound_add_mul_error`, `horner_step_error` and the `butterfly_*` lemmas
could all be instantiated at it. Three fields close it: `valid_zero`, `valid_one`, and
`inRange_of_valid` — a value the format holds is a result it can round to, which for `FPR` is
`FPR.inNormalMagnitudeRange_toReal_of_isNormalOrZero`, exactly attained at both ends.

## The refutations are in the branch now

`Extern/Falcon/FPRBridgeRefutations.lean` carries the kernel-checked `¬ (∀ …)` for each bound
read without its hypotheses, at an explicit `UInt64` input, plus `not_add_isNormal_closed` —
@dtumad's `1 + (-1)`. It is downstream of the bounds, so it cannot weaken anything it refutes.

They share one argument and need no numeric estimate: dropping the restriction lets the routine
return a value at or below zero while the exact result is strictly positive, and the absolute
error is then at least the whole exact value, which no `ε < 1` multiple of it can cover. Three
kinds of real behaviour produce it:

- **overflow** — `add` reaches the non-finite exponent, which this denotation sends to `0`, and
  `mul` wraps straight past it to a small negative value;
- **underflow** — two ordinary normals whose exact *difference* is subnormal are mis-rounded by
  `add`'s alignment step into a large negative value, which is why the hypothesis restricts the
  exact result and not only the operands;
- **subnormal operands discarded** — `div`'s flush guard fires on a zero exponent field, so a
  subnormal numerator is dropped and the quotient returns `+0`.

`FPR.div`'s 55-iteration restoring-division loop and `FPR.sqrt`'s 54-iteration digit recurrence
do not reduce under `decide` — the `forIn` gets stuck, and `decide +kernel` runs for minutes. The
`sqrt` refutation goes through `sqrtPipeline_loopRes`, which trades the loop for `Function.iterate`
and does reduce; the `div` one sidesteps the loop entirely by using an input the flush guard
discards before the quotient matters.

## One thing to know when reviewing

The statements are textually identical to `main`, but `toReal` now means the decoder. Agreement with Lean's `Float` rests on a 30-pattern differential check, not a bridge lemma: the gap is closed by replacement.

## Follow-up: this file is now too big

At 7.4k lines `FPRBridge.lean` is more than double the repo's previous largest (`SimulateQ.lean`, 3.5k); `ExpmBridge.lean` adds a further 1.9k. It wants splitting along the seams it already has — the layering is acyclic, so the cut is mechanical:

`Decode` (decode, `toReal`, domain predicates) · `Kernel` (sticky folds, leading-zero count, rounding table, `make`/`make_z`) · `Add` · `Mul` · `Loop` (the `forIn` principle and the two loop pipelines) · `DivSqrt` · a thin re-export retaining `expm_p63_error` and the verifier bridge.

Deliberately **not** done here: git would render a 1→6 split as thousands of deleted plus thousands of added lines and bury the content worth reviewing. Better as its own PR whose whole claim is "pure code motion" — checkable by comparing `#print axioms` for the six bounds before and after, rather than by reading. Attribution on the new files is the file owner's call, so I would rather ask than guess.

## Verification

Full CI set green; CI-lint clean; extern and interop isolation OK. Every new declaration is axiom-clean — a sweep of all constants in the module found nothing else. No `bv_decide` and no `native_decide` anywhere: both emit custom trust axioms.

Each original statement is refuted by a kernel-checked `¬ (∀ …)` against its verbatim text, axiom-clean to the same standard, in `Extern/Falcon/FPRBridgeRefutations.lean`. That is the evidence that the added hypotheses are necessary rather than convenient.

The `FloatLike.HasRealSemantics FPR` instance is live and axiom-clean, so the six bounds now reach the compound lemmas rather than stopping at the file that proves them.


dtumad:
Reviewed by building the branch (3080 jobs, clean) and testing the claims rather than reading them. The central one holds up, and it is a good deal stronger than "the bounds needed tightening".

## Verified

**The decoder is right.** `FPR.decode` takes sign = bit 63, exponent = `(x >>> 52) % 2^11`, mantissa = `x % 2^52`, and `FPR.Bits.toReal` splits subnormal (`±m·2^-1074`), non-finite (`→ 0`, matching the old `toRat0` convention) and normal (`±(1 + m/2^52)·2^(e-1023)`). That is textbook binary64, and it reduces in the kernel as claimed.

**All seven bounds are axiom-clean.** `add_error`, `sub_error`, `mul_error`, `div_error`, `sqrt_error`, `expm_p63_error`, `expm_p63_error_abs` — every one `[propext, Classical.choice, Quot.sound]` under `#print axioms`. No `native_decide`, no `bv_decide`. `FPRBridge.lean` and `ExpmBridge.lean` are genuinely `sorry`-free (the two `sorry` hits in `ApproxArith.lean` are the word in prose).

**The refutation reproduces exactly.** For `sqrt_error` at `+∞`:

```
FPR.decode 0x7FF0000000000000  = { sign := false, exponent := 2047, mantissa := 0 }   -- toReal = 0
FPR.sqrt   0x7FF0000000000000  = 6913025428013711360
FPR.decode (FPR.sqrt inf)      = { sign := false, exponent := 1535, mantissa := 0 }   -- = 2^512
```

So the hypothesis `0 ≤ toReal a` holds, the right-hand side is `0`, and the old statement asserted `|2^512 − 0| ≤ 0`. The domain restrictions are necessary rather than defensive, and the non-vacuity witnesses show they still admit ordinary values. Restricting the domain while leaving every conclusion unchanged is the right call.

## Three things before merge

**1. New prose in `ApproxArith.lean` contradicts the PR.** The added "What remains open" block says:

> The bodies of `add_error`, `mul_error`, `div_error`, `sqrt_error` in `FPRBridge.lean` are `sorry` (the correctly-rounded three-stage pipeline proof is future work).

That text is added by this PR and is false as of this PR. Anyone reading the file after merge concludes the bounds are unproven. The rest of that block (the `_valid` fields, `neg_valid`, `sub_error`) is still accurate.

**2. Worth stating the scope in the description.** `HasRealSemantics` has no instances and no users outside `ApproxArith.lean` — the FPR instance is still commented out, blocked on the five `_valid` closure fields. So the six bounds do not yet reach any downstream Falcon theorem. That is fine and the class generalization is well argued, but the follow-up section only mentions the file split, and this seems like the more important half of "what is left". Relatedly, `Extern/Falcon/Instance.lean:140` still has `ofRawFPR x := ((Float.ofBits x).toRat0 : ℝ)`, so two denotations of a raw `FPR` word now coexist — intentional?

**3. A question about the `berExp` fix.** The head commit swaps the argument reduction from `rint` to `floor_`. That changes the sampler's output distribution, which is exactly what the GPV argument depends on, and the description does not mention it at all — it reads as a pure error-bounds PR. It deserves its own paragraph.

The substantive question: the saturation logic is written for `si ≥ 0` only. `(63 - sShift) >>> 26` detects `sShift ≥ 64` via the borrow bit, but not negatives:

```
si ≥ 0:   0→0, 1→1, 63→63, 64→63, 100→63     (saturates as documented)
si < 0:  -1→63, -2→62, -10→54                 (acceptance ≈ 0, not the correct >1 factor)
```

So `floor_` is right precisely when `x ≥ 0` holds at the call site. `x = diff²·dss − z0²·invSqr2Sigma0` is a difference of nonnegatives — true mathematically by the usual `σ ≤ σ₀` argument, but the docstring asserts the caller's guarantee without proving it, and the new regression only probes `berExpReduce` on `k/32` for `k ∈ [1,200]`, all positive. What rules out `x` dipping slightly below zero under rounding when the true value is near zero? If nothing does, it would be worth handling the negative branch explicitly rather than relying on saturation that was not designed for it.

Nothing above blocks the PR. (1) is a paragraph, (2) is description text, (3) is a question I would like answered before the sampler change lands.

Agreed on the file split being its own PR — "pure code motion, checkable by comparing `#print axioms` before and after" is the right framing.

alik-eth:
Thanks — all three land, and (3) turned up something I had not checked.

**1. `ApproxArith.lean` prose — fixed** in `fd87169f`. You're right that this PR introduced it and that it was false as of this PR. It now lists the `_valid` items that are actually open, plus the consequence you raise in (2), stated plainly: with the instance commented out, the bounds reach no downstream Falcon theorem.

**2. Scope — agreed, description updated.** On `Instance.lean:140`: not intentional, it is a real gap. That is the `FloatLike ℝ` ideal model, and `Extern/Falcon/FFT.lean:48,52` reads the GM twiddle table through it, so two denotations of a raw FPR word are live at once — `(Float.ofBits x).toRat0`, which is opaque, and `FPRBridge.toReal`, which is the proved decoder. They agree on every finite pattern and both send non-finite to `0`, so repointing should be semantically inert. But it changes the denotation the ideal FFT model is built on, and by your own framing that belongs in its own commit rather than folded into an error-bounds PR. I would rather do it separately.

**3. The `berExp` question.** Your saturation table is right; I reproduced it exactly:

```
sShiftOf [-1, -2, -10, 0, 1, 63, 64, 100] = [63, 62, 54, 0, 1, 63, 63, 63]
```

And it is worse than you put it: **on `x < 0`, `floor_` is worse than `rint` was.** At `x = -0.01`, `ccs = 0.5`, comparing the 64-bit acceptance threshold against the correct `ccs·exp(0.01) = 0.50503`:

```
rint  : si = 0,  sShift = 0   weight/2^64 = 0.49503     (off by 2%, right by accident)
floor_: si = -1, sShift = 63  weight/2^64 = 0           (always reject)
```

`rint` lands near the right answer because `si = 0` leaves `r = x` and sign-blindness reads `exp(-|x|)`. So the change is right exactly on `x ≥ 0` and strictly worse without it, which makes your question the right one.

What rules out `x` dipping below zero: nothing local. It rests on `σ ≤ σ₀`, a key-generation invariant. Whether float rounding can break it is a real question, because `dss = fl(fl(isigma²)·½)` is computed from `isigma` while `inv2σ0` is a hardcoded constant, so the two can disagree by an ulp even at `σ = σ₀` exactly. That is the only route to `x < 0`, since `z0 = 0` makes the second term exactly `0`. Measured:

- at the representable `1/σ` just **above** `1/σ₀` (`σ = 1.8204999999999998 ≤ σ₀`): `x = 0` **exactly**, for `r = 0, b = 0`, at every `z0 ∈ [0,25]`
- one ulp **down** (`σ = 1.8205000000000002 > σ₀`): `x < 0` for 25 of those 26
- 2M draws with `σ ∈ [σ_min, σ₀]`: min `x = 8.5e-13 > 0`

So the precondition holds for every `σ ≤ σ₀`, and the margin at the boundary is exactly zero. `fd87169f` records this on `samplerZLoop` and pins **both** sides in the test — the negative control is the point, since testing only `σ ≤ σ₀` would not distinguish a real invariant from a coincidence. It also pins `berExpReduce 0`, which the `k/32` sweep skipped.

On adding an explicit negative branch: I would rather not without your call. The reference has the identical structure and the identical *unproved* precondition — `sign_sampler.c:874` carries "We can use fpr_trunc() because x >= 0", and the loop justifies `S(z) ≤ G(z)` in prose only — so a guard would be the first deliberate divergence from bit-exactness with the C. It is cheap to add if you want it.

One thing I should flag rather than leave implicit: the end-to-end statement "the reduced argument is in range" is **not** proved, and it is harder than it looks. Sterbenz makes the subtraction exact, so `0 ≤ r` reduces to `fl(si·L') ≤ x`. But `invLog2Const · log2Const = 1 − 0.43·u`, below one by less than half an ulp, while the naive relative-error chain has to absorb `(1+u)²` — it misses by ~1.6u. And the fact is tight: scanning the smallest `x` per `si` for `si = 1..399`, the worst margin is **0 ulp**, with equality at `si = 1, x = fl(log 2)`. So it needs an exact rounding characterisation plus `toReal_ofInt` and a `floor_` bridge, neither of which exists yet. I did widen `expm_p63_error`'s domain to `x ≤ 694/1000` so that such a chain *could* close — a computed `r` lands a few ulps above `log 2` when `x` is near a multiple of it, so nothing stated at `log 2` would ever apply.

Extern/Falcon/ApproxArith.lean
```diff
@@ -237,34 +296,56 @@ theorem ieee754_machineEpsilon_lt_one : ieee754_machineEpsilon < 1 := by
...
+name — their statements already carry exactly the `IsNormal`/`InNormalMagnitudeRange`
+hypotheses this class's fields ask for. `div_error` needs its hypotheses supplied in a
+different order than `FPRBridge.div_error` uses, but is otherwise the same theorem.
+
+What remains open is the `_valid` half of the class, not the error bounds:
+
+- The `_valid` closure fields (`add_valid`, `mul_valid`, `div_valid`, `sqrt_valid`,
+  `sub_valid`) have no counterpart in `FPRBridge.lean` yet; each needs a bit-level
+  argument that a correctly-rounded, in-range result stays normal (no separate
+  `IsNormal`-losing failure mode beyond the overflow/underflow `InRange` already rules
```
dtumad: Exact cancellation is exactly such a failure mode, and `InRange` does not rule it out: `FPR.InNormalMagnitudeRange` is `r = 0 ∨ (minNormalReal ≤ |r| ∧ |r| ≤ maxFiniteReal)`, so `0` is in range, while `FPR.IsNormal` requires `exponent ≠ 0`. `1 + (-1)` refutes `add_valid` — see the main comment for the compiled refutation.

Extern/Falcon/ApproxArith.lean
```diff
@@ -237,34 +296,56 @@ theorem ieee754_machineEpsilon_lt_one : ieee754_machineEpsilon < 1 := by
...
+  `private` to `FPRBridge.lean`.
+- `sub_valid` has no bit-level pipeline of its own: `FPR.sub a b` unfolds to
+  `FPR.add a (FPR.neg b)`, so it should reduce compositionally to `add_valid` and
+  `neg_valid`, the way `sub_error` already reduces to `add_error` and `neg_exact`.
+
+Until these are discharged, this instance stays commented out, and so the error bounds
+proved in `FPRBridge.lean` do not yet reach any downstream Falcon theorem. -/
 -- instance : FloatLike.HasRealSemantics FPR ieee754_machineEpsilon where
 --   interp := Falcon.Concrete.FPRBridge.toReal
+--   Valid := Falcon.Concrete.FPR.IsNormal
```
dtumad: This name does not resolve: the declaration is `Falcon.Concrete.FPRBridge.FPR.IsNormal` (`FPRBridge.lean:332`, inside `namespace Falcon.Concrete.FPRBridge`). Same for `InNormalMagnitudeRange` on the next line. Adding the `open Falcon.Concrete.FPR in` above does not help.

Extern/Falcon/SamplerZ.lean
```diff
@@ -221,6 +243,14 @@ private def sigmaMinConsts : Array F := #[
   | some sigmaMin => sigmaMin
   | none => panic! s!"Falcon samplerZ does not support logn={logn}"
 
+/-- One rejection round: draw `z` from the bimodal half-Gaussian and keep it with probability
+`ccs * exp (-x)`, where `x = (z - r) ^ 2 / (2 * σ ^ 2) - z0 ^ 2 / (2 * σ₀ ^ 2)`.
+
+The `x ≥ 0` that `berExp` requires holds because `|z - r| ≥ z0` for either value of `b` when `r`
+is in `[0, 1)`, together with `σ ≤ σ₀`. That second half is a key-generation invariant rather
```
dtumad: This is the sentence I would like to see backed by something mechanical. `samplerZ` takes an unrestricted `isigma`, and by your own next sentence the margin is one ulp — so the boundary is reachable by a caller bug rather than only by a wild input, and crossing it silently changes the output distribution instead of failing.

dtumad (commented):
Follow-up to my 2026-08-14 pass, re-checked at `fd87169f`.

Both things I flagged are handled. The `ApproxArith.lean` status note no longer claims the four bounds are `sorry`, and `berExp`'s `x ≥ 0` is now *located* rather than asserted — `samplerZLoop`'s docstring derives it from `|z - r| ≥ z0` plus `σ ≤ σ₀`, and then volunteers that the margin is exactly one ulp wide. That is a better answer than the one I asked for.

One new blocker, in the part I had only looked at from the outside.

**The intended `HasRealSemantics FPR` instance cannot exist.** With `Valid := FPR.IsNormal` and `InRange := FPR.InNormalMagnitudeRange`, the `add_valid` field is refutable at `1 + (-1)`: both operands are normal, the exact sum is `0`, `0` is in range by the `r = 0` disjunct of `InNormalMagnitudeRange`, and `FPR.add` returns `+0`, which is not normal (`IsNormal` requires `exponent ≠ 0`).

```lean
theorem add_valid_false :
    ¬ (∀ a b : FPR.FPR, FPR.IsNormal a → FPR.IsNormal b →
        FPR.InNormalMagnitudeRange (toReal a + toReal b) →
        FPR.IsNormal (FPR.add a b)) := by
  intro h
  have hone : FPR.IsNormal FPR.one := by
    unfold FPR.IsNormal FPR.Bits.IsNormal FPR.decode; decide
  have hneg : FPR.IsNormal (FPR.neg FPR.one) := by
    unfold FPR.IsNormal FPR.Bits.IsNormal FPR.decode; decide
  have hsum : toReal FPR.one + toReal (FPR.neg FPR.one) = 0 := by
    rw [toReal_neg, toReal_one]; ring
  have hrange : FPR.InNormalMagnitudeRange (toReal FPR.one + toReal (FPR.neg FPR.one)) := by
    rw [hsum]; exact Or.inl rfl
  have hbad : ¬ FPR.IsNormal (FPR.add FPR.one (FPR.neg FPR.one)) := by
    unfold FPR.IsNormal FPR.Bits.IsNormal FPR.decode; decide
  exact hbad (h FPR.one (FPR.neg FPR.one) hone hneg hrange)

-- the computed sum is exactly +0
theorem add_one_negOne_decode :
    FPR.decode (FPR.add FPR.one (FPR.neg FPR.one)) = ⟨false, 0, 0⟩ := by
  unfold FPR.decode; decide
```

`#print axioms add_valid_false` → `[propext, Classical.choice, Quot.sound]`.

This matters more than a missing proof, because it changes what the remaining work is. The status block says the `_valid` fields need "a bit-level argument that a correctly-rounded, in-range result stays normal (no separate `IsNormal`-losing failure mode beyond the overflow/underflow `InRange` already rules out)". Exact cancellation is precisely such a failure mode, and `InRange` does not rule it out — it explicitly *admits* zero. So the open item is the contract, not five closure proofs: either let `Valid` include zero and state each operation's closure at the strength it actually has, or split operand validity from result finiteness/range. `add_error` itself is fine at this input (both sides are `0`); it is only `add_valid` that fails.

Worth fixing while you are in there: the commented block names `Falcon.Concrete.FPR.IsNormal`, which does not resolve — the declaration is `Falcon.Concrete.FPRBridge.FPR.IsNormal`. Uncommenting it today would fail on name resolution before it ever got to `add_valid`, which is a fair sign the block has not been elaborated since it was written. I would rather see the class shipped with a satisfiable contract than with a commented instance standing in for one.

On `berExp`, the residual concern is now narrow and is about enforcement rather than analysis. `samplerZLoop`'s docstring is explicit that `σ ≤ σ₀` is "a key-generation invariant rather than anything checked here", and that one ulp past `σ₀` gives `x < 0`. `samplerZ`'s public `isigma` is unrestricted, and the failure mode on the wrong side of that boundary is silent: the saturation has no negative branch, so a negative `si` truncates to `63 + si` and drives the acceptance weight to ~0 where the correct weight is above 1 — a wrong sampling distribution, not a crash. Given that this is the distribution the GPV argument rests on, could the invariant be carried by the caller's type, or checked defensively at the entry point? Splitting the runtime change into its own conformance PR would also be a reasonable answer.

None of this touches the six error bounds, which I think are solid and are the bulk of the PR.

alik-eth:
You're right, and it's my bug — `Valid`/`InRange`/the `_valid` fields are all added by this PR. I ran your refutation verbatim: it compiles and is `[propext, Classical.choice, Quot.sound]`, and `add_one_negOne_decode` reduces to `⟨false, 0, 0⟩` in the kernel.

Worse than a missing proof, as you say. The status note claimed there was "no separate `IsNormal`-losing failure mode beyond the overflow/underflow `InRange` already rules out" — exactly backwards, since `InRange` *admits* zero through its own `r = 0` disjunct. That sentence predates the branch but I carried it forward unchecked in `fd87169f`, which is the same way the stale `sorry` note survived until you caught it. I've stopped treating inherited prose as load-bearing-but-verified.

## The repair, in `90e0484d`

`FPR.IsNormalOrZero` — normal and finite, or exactly `±0`. `FPR.IsZero` is stated on the decoded fields rather than as `toReal x = 0`, since the latter also holds on `Inf` and `NaN` under this decoder's convention, which would readmit exactly the encodings that made `sqrt_error` false at `+∞`.

Proved against it:

- **`add_isNormalOrZero`.** The cancellation branch packs `±0` via `decode_make_z_of_zero`. The normal branch reuses `add_error`'s own exponent window: `decode_make_z_of_no_carry` puts the packed exponent at `(ex''' + 1076) + 1`, inside `[1, 2046]` on the window `-1076 ≤ ex''' ≤ 969`; the rounding carry adds `2` instead of `1`, and is ruled out at `ex''' = 969` by `addPipeline_no_carry`, so the result never reaches `2047`.
- **`sub_isNormalOrZero`**, compositionally, the way `sub_error` reduces to `add_error`.
- **`FPR.isNormalOrZero_neg`**, carrying both disjuncts — that's `neg_valid`.
- Your `1 + (-1)` is banked as a witness that the zero disjunct is reachable, so the widening isn't idle:

```lean
example : FPR.IsNormalOrZero (FPR.add FPR.one (FPR.neg FPR.one))
    ∧ ¬ FPR.IsNormal (FPR.add FPR.one (FPR.neg FPR.one))
```

All axiom-clean.

## What is still open, stated accurately this time

`mul_valid`, `div_valid`, `sqrt_valid` are **not** proved. Each needs `add_isNormalOrZero`'s exponent-window argument transposed to `FPR.make` and that operation's pipeline. Worth noting they are not *refuted*: a product, quotient, or square root of nonzero normals cannot cancel, so the `IsNormal` reading survives for those three — only `add` and `sub` were actually false.

Then the `_error` fields need widening from `IsNormal` to `IsNormalOrZero` operands, since `Valid` now admits zero. Every zero-operand case is exact rather than approximate (`x + 0`, `x * 0`, `sqrt 0`), and `div_error` already carries the `interp b ≠ 0` that case needs. That widening is independently wanted: Falcon feeds exact zeros constantly, so the current bounds miss roughly a tenth of Falcon-512's coefficients.

So the instance stays commented out for now, but the contract underneath it is satisfiable rather than false, and the block says so plainly instead of implying the opposite. I also fixed the name resolution you flagged — `Falcon.Concrete.FPRBridge.FPR.IsNormal`, not `Falcon.Concrete.FPR.IsNormal`.

## The sampler

Split out to #517, per your suggestion, and reverted here — `537d27e7`. This branch keeps only the bounds work it motivated but does not depend on: the `x ≤ 694/1000` domain and the sign-blindness results. The docstrings that named `berExpReduce` now describe the situation directly instead, since the declaration no longer lives on this branch.

Your enforcement question about `σ ≤ σ₀` is carried over to #517 and called out there as unresolved rather than answered.

alik-eth:
Both halves of your last pass are answered, and the second one turned up something you did not flag and I should have.

## 1. `add_valid` — repaired, and the instance is live

Your refutation is right and I ran it verbatim: it compiles and is `[propext, Classical.choice, Quot.sound]`, and `add_one_negOne_decode` reduces to `⟨false, 0, 0⟩` in the kernel. Both are now in the branch as `not_add_isNormal_closed` and `decode_add_one_negOne`.

`FPR.IsNormalOrZero` is the repair. Rather than stop at the closure fields, the `_error` bounds are widened to that same domain, so **`FloatLike.HasRealSemantics FPR` is uncommented and axiom-clean**. Your (2) from the pass before — that the six bounds reaching no downstream theorem was the more important half of what was left — is closed rather than deferred.

Widening needed each pipeline traced at an operand whose exponent field is `0`. Those cases are exact rather than approximate, so no new error analysis was involved:

- `FPR.add` reassembles the other operand unchanged. Nothing is aligned in (`yu` vanishes through `stickyShift 0`), the renormalising count is forced to `8`, and the nine-bit sticky fold divides exactly, so the packing round-trips. Stating it on the pipeline's ordered `x'` rather than on `a`/`b` makes one trace cover both argument orders — `addPipeline_swap_cases` then supplies the orientation from the `magKey` comparison.
- `FPR.mul` and `FPR.div` already carry a flush guard (`dzu`) that fires on a zero exponent field. It clears the significand *and* substitutes the exponent `-1076`, which `make` packs as a bare sign bit — so those two needed a mirror of `mulPipeline_dzu_eq_zero` rather than any new arithmetic.
- `FPR.sqrt` masks its significand against the same test and `make_z` collapses on it.

One thing worth knowing if you review that section: `FPR.add` is **not** bit-preserving at a zero operand — `(+0) + (-0)` is `+0` — so the widened statements are phrased on `toReal` and on the decoded fields, never on the bit pattern.

## 2. You were right that a commented instance was standing in for a contract — and the contract was still wrong

"I would rather see the class shipped with a satisfiable contract than with a commented instance standing in for one" turned out to cut deeper than the name-resolution slip you spotted. Carrying a domain is what makes the class satisfiable at all, but it also made it satisfiable *for nothing*:

```lean
noncomputable instance degenerate : FloatLike.HasRealSemantics FPR ieee754_machineEpsilon where
  interp := toReal
  Valid := fun _ => False
  InRange := fun _ => False
  add_error := fun _ _ h _ _ => absurd h not_false
  add_valid := fun _ _ h _ _ => h
  ...
```

Every `_error` and `_valid` field discharges vacuously, leaving only `interp_zero`, `interp_one` and `neg_exact` with any content. It typechecks and is axiom-clean. So `[HasRealSemantics F ε]` on its own was an assumption that said nothing, and `compound_add_mul_error`, `horner_step_error` and the `butterfly_*` lemmas could all be instantiated at it — the same failure mode as the unconditional version, one level up.

Three fields close it: `valid_zero`, `valid_one`, and `inRange_of_valid : ∀ a, Valid a → InRange (interp a)` — a value the format holds is a result it can round to. For `FPR` the last is `FPR.inNormalMagnitudeRange_toReal_of_isNormalOrZero`, and it is exactly attained at both ends (`2 ^ (-1022)` below, `maxFiniteReal` above), so it is not slack I gave myself.

## 3. The refutations are in the branch

You asked, in effect, for the evidence rather than the assertion. `Extern/Falcon/FPRBridgeRefutations.lean` carries a kernel-checked `¬ (∀ …)` for every bound read without its hypotheses, at an explicit `UInt64` input. It imports the bounds rather than the other way round, so it cannot weaken what it refutes.

They share one argument and need no numeric estimate: dropping the restriction lets the routine return a value at or below zero while the exact result is strictly positive, so the absolute error is at least the whole exact value and no `ε < 1` multiple of it can cover that. The three behaviours behind it are overflow (`add` to the non-finite exponent, `mul` wrapping past it), underflow (two ordinary normals whose exact *difference* is subnormal — the case that justifies restricting the exact result and not only the operands), and subnormal operands being discarded outright by `div`'s flush guard.

Two of them fought back, which is worth recording: `FPR.div`'s 55-iteration restoring division and `FPR.sqrt`'s 54-iteration digit recurrence do not reduce under `decide` — the `forIn` gets stuck, and `decide +kernel` ran for over two minutes without finishing. The `sqrt` one goes through `sqrtPipeline_loopRes`, which trades the loop for `Function.iterate` and does reduce; the `div` one sidesteps the loop by using an input the flush guard discards before the quotient matters. The module builds in 4.5s.

## 4. Still open, and unchanged

`Extern/Falcon/Instance.lean:140` and `Extern/Falcon/FFT.lean:48,52` still read the GM twiddle table through `(Float.ofBits x).toRat0`, so two denotations of a raw `FPR` word remain live. Repointing should be semantically inert but changes what the ideal FFT model is built on, so by your own framing it belongs in its own commit — I have not touched it here. The `FPRBridge.lean` split is likewise still deferred; the file is 7.4k lines now.

quangvdao (commented):
Thanks @alik-eth for this substantial and careful formalization. In particular, replacing the opaque `Float.ofBits` / `toRat0` path with a kernel-reducible decoder, finding that the original unrestricted statements were actually false, committing the counterexamples, repairing the exact-cancellation hole with `IsNormalOrZero`, and completing all six bounds is excellent work. I did an independent statement/API/consumer/axiom review at exact head `adb82a291d8114f29f9a335a2eb6a07f1bbd9f35`.

I did **not** find a P0/P1 mathematical or soundness defect in the six bounds. The principal theorems and the live `HasRealSemantics FPR` instance elaborate with coherent domains, and the checked theorem dependencies are exactly `[propext, Classical.choice, Quot.sound]`.

I do have the following requested changes and hardening findings.

### [P2] Rebase and validate against the current target

The PR is based on `ea9916d` / Lean 4.32.2, while current `main` is `3ecd552` / Lean 4.33.0. GitHub currently reports `CONFLICTING` / `DIRTY`, and current `main` adds the whole-library axiom sweep and a stronger canary-library build.

The textual conflict I found is the import at the top of `Extern/Falcon/FPRBridge.lean`. The likely resolution is to keep

```lean
import all LatticeCrypto.Falcon.Concrete.FPR
```

and drop the obsolete `public import Batteries.Data.Float.Rat`; then rerun the full current CI, especially `lake exe axiomsweep --check` and `lake build VCVioTest`. The current-target result needs review because the existing green checks are for 4.32.2.

### [P2] Reduce the exported proof-infrastructure surface

`FPRBridge.lean` places almost the whole 7.5k-line module under `@[expose] public section`. A repository consumer-graph check found 73 newly added declarations with no Lean reference outside this file. The problematic subset includes generic proof machinery such as `stickyShift`, `or_one_eq`, `or_and_self`, `uint32_add_mask_shiftRight_of_lt`, `roundQuarterTiesEven`, and `roundTableBit` plus their law families.

Please make proof-local helpers `private`. If some generic bit/rounding results are intended as reusable APIs, move them to the lowest appropriate `ToMathlib` owner with a named consumer and a direct canary. The public surface here should mainly be the decoder/denotation, domain predicates, structural laws meant for consumers, principal error/closure theorems, and verifier bridges.

### [P2] Correct the `HasRealSemantics` composition contract documentation

`ApproxArith.lean:25-32` and `:71-73` say that `_valid` lets compound expressions carry/re-derive both `Valid` and `InRange` through every intermediate result rather than assuming them at each step. The actual API propagates only `Valid`: `compound_add_mul_error`, `horner_step_error`, and both butterfly lemmas still take each exact-result `InRange` premise explicitly. The same overstatement appears in the `FPRBridge.lean:44-46` overview.

Please say that validity is propagated while exact-result range remains an algorithm-specific proof obligation. Alternatively, add genuine range-closure lemmas with sufficient magnitude hypotheses. The Lean statements are sound; this is a misleading foundational contract description.

### [P3] Make the raw `FPR.Bits` invariant explicit

`FPR.Bits` documents an 11-bit exponent and 52-bit mantissa but stores unrestricted `Nat`s. Thus the following malformed field triple is accepted as “normal”:

```lean
example : FPR.Bits.IsNormal ⟨false, 2048, 0⟩ := by
  norm_num [FPR.Bits.IsNormal]
```

Also, `FPR.Bits.abs_toReal_sub_of_succ_mantissa` describes adjacent representable values without bounding the original mantissa, so the successor can leave the 52-bit field.

The clean repair is `Fin (2^11)` / `Fin (2^52)`, or an explicit `WellFormed` predicate required by public raw-field theorems. At minimum, document `Bits` as an unchecked triple and avoid representability language without the bounds.

### [P3] Clean up stale documentation and unused private declarations

* `FPRBridge.lean:396-398` still says the conditional swap and later `FPR.add` stages “remain open”, although this PR proves the end-to-end pipeline.
* `FPRBridge.lean:7199` says `FPR.mul` returns zero when either operand is “one”; the theorem requires an operand to be **zero**.
* Twelve new private declarations have no non-self reference, including `addPipeline_n'`, `addPipeline_m`, `forIn_range_induction`, `divPipeline_quotient_lt`, `allOnes_toNat`, and `hornerMachine_le_coeff`. Please delete them unless they are wired into an actual principal proof or canary.

### Validation and prior feedback

At exact head `adb82a2` I independently built `FPRBridge`, `ExpmBridge`, `ApproxArith`, and `FPRBridgeRefutations`; checked the statement and axiom set of the six principal bounds; synthesized the live `HasRealSemantics FPR` instance; and ran all eight import-completeness checks, style lint, Extern/Interop isolation, agent-doc/generated-doc checks, and `git diff --check`. All passed. I found no added `sorry`, `admit`, axiom, unsafe declaration, `native_decide`, `bv_decide`, `Eq.rec`, or `HEq`.

The three older review requests are substantively handled: exact cancellation is repaired by `IsNormalOrZero` and committed refutations; the instance names now resolve; and the sampler invariant/runtime issue was split into #517, which is merged. Their threads remain unresolved/outdated in the UI and can be closed after the final rebase.

I am not charging the still-open composition from these local arithmetic bounds through `ffSampling` to the final sampler Rényi/security bound to this PR: `LatticeCrypto/Falcon/Security.lean` already records that as deferred work. Likewise, splitting the very large bridge and committing a durable decoder-vs-`Float` differential canary would be useful follow-ups.

Overall: the mathematical core looks strong. My request is to make it mergeable and current-target-clean, narrow the foundational API boundary, and correct the contract/invariant documentation before landing.

alik-eth:
Thanks — all five items are addressed, at head `0148bca5`, rebased onto `89e48210`. GitHub now reports the PR `MERGEABLE`.

## Something the rebase turned up that is worth recording

Replaying this branch onto current `main` **silently reverted #517**. The branch carried a commit reverting `SamplerZ.lean` and the sampler tests — correct when the `berExp` fix still lived here and was being split out, but #517 has since merged, so the revert replayed against a `main` that already had the fix and stripped it back out. The net diff showed `−69` lines on those files.

Nothing would have caught it: the pre-fix sampler compiles perfectly, every test still passes, and the PR is not otherwise about that file. It is fixed in `2153a220`, and I have added a standing check — this branch's net diff against `main` must not touch `SamplerZ.lean` or `LatticeCryptoTest/Falcon/Main.lean` at all, which is what the split intended and is now true.

## [P2] Rebase and current-target validation

Rebased onto `89e48210` / Lean **4.33.0**. The only textual conflict was the one you predicted, resolved as you suggested: keep `import all LatticeCrypto.Falcon.Concrete.FPR`, drop `public import Batteries.Data.Float.Rat` — the decoder is exactly what removed the `Float.toRat0` dependency that import served.

Three 4.33 breakages, all numeric tactics behaving differently:

- `not_sub_error_operands_only` ended `norm_num; positivity`; `norm_num` now closes that goal alone.
- `toReal_neg` and `toReal_minNormalSucc` drew the new diagnostic for declining to evaluate `2 ^ 1074` / `2 ^ 1022` against a default threshold of `256`. Binary64 reaches `2 ^ (-1074)` at its own subnormal scale, so the threshold sits below the format's exponent range; raised to `1100` on exactly those two declarations with the reason inline. Not a suppressed diagnostic — and not cosmetic either: with the larger budget `norm_num` closes `toReal_minNormalSucc` outright, which made its trailing `ring` redundant in turn.

Validated on the new target: full library build, `lake build VCVioTest`, **`lake exe axiomsweep --check` passed (no new axiom/sorry taint)**, style lint, and both isolation checks. The live instance and the principal bounds re-checked at `[propext, Classical.choice, Quot.sound]`.

## [P2] Exported surface

56 declarations are now `private` — the generic bit and rounding families you named (`stickyShift`, `roundQuarterTiesEven`, `roundTableBit`, and the `or_` / `and_` / `shiftRight_` / `toNat_` helpers). What stays public follows your criterion: decoder and denotation, domain predicates, structural laws meant for consumers, the principal error/closure theorems, and the verifier bridges. Both sibling modules reach this one through `import all`, so nothing downstream lost access.

I did not take the `ToMathlib` option for any of them. None currently has a consumer, and inventing one to justify the move seemed worse than keeping them local until a real second user appears.

Worth flagging: my first pass at this list was wrong in a way that looked authoritative. The scan reported 86 unreferenced declarations including `add_isNormalOrZero`, which the instance plainly uses — a negative lookbehind excluded a preceding `.`, so every fully-qualified `Falcon.Concrete.FPRBridge.foo` reference was invisible. Corrected, the count is 76 and the principal theorems drop out as they should.

## [P2] Composition contract

You are right and this was my error. Only `Valid` propagates; every compound lemma takes one `InRange` premise per exact intermediate, and no closure property could supply it, since it depends on the values an algorithm actually feeds in. Corrected in all three places — `ApproxArith.lean` twice and the `FPRBridge.lean` overview — stated as validity propagating while exact-result range remains an algorithm-specific obligation. I did not add range-closure lemmas; that would need magnitude hypotheses the callers do not have.

## [P3] `FPR.Bits` invariant

Documented as an unchecked triple, explicitly noting that out-of-range triples like `⟨false, 2048, 0⟩` inhabit it and satisfy `IsNormal`. Added `FPR.Bits.WellFormed` and `FPR.decode_wellFormed`, so the fact that nothing `decode` produces is out of range is now a theorem rather than an assumption. `abs_toReal_sub_of_succ_mantissa` carries the caveat that it is stated on a raw triple and does not bound the incoming mantissa, with a pointer to where the bound comes from on decoded values.

I did not convert the fields to `Fin (2 ^ 11)` / `Fin (2 ^ 52)`. That touches every `decode`-facing proof in the file and felt like the wrong change to fold into a rebase; happy to do it as its own PR if you would rather have the stronger form.

## [P3] Stale documentation and dead declarations

Both comments fixed — the one claiming the conditional swap and later `add` stages remain open, and the `mul` flush docstring that said "one" where the theorem requires zero. Removed 15 unreferenced private declarations (134 lines), including the six you named.

## Not done, deliberately

`Instance.lean:140` / `FFT.lean:48,52` still read the GM table through `(Float.ofBits x).toRat0`, and the `FPRBridge.lean` split is still deferred. Both remain follow-ups for the reasons given earlier in the thread.

Extern/Falcon/ApproxArith.lean
alik-eth: Addressed: exact cancellation is why `Valid` is now `FPR.IsNormalOrZero` rather than `FPR.IsNormal`, and your `1 + (-1)` is committed as `not_add_isNormal_closed` in `FPRBridgeRefutations.lean`. The status block no longer claims `InRange` excludes this failure mode.

Extern/Falcon/ApproxArith.lean
alik-eth: Addressed: the instance is live and the names resolve — `Falcon.Concrete.FPRBridge.FPR.IsNormalOrZero` / `.FPR.InNormalMagnitudeRange`. It elaborates and is axiom-clean, so the commented block is no longer standing in for a contract.

Extern/Falcon/SamplerZ.lean
alik-eth: Addressed by splitting the runtime change out to #517, which is merged. The `σ ≤ σ₀` enforcement question you raised is recorded there as open rather than answered — this branch no longer touches `SamplerZ.lean` at all.

alik-eth:
Correction to my last comment on the dead-declaration item — I called it done and it was not. Two things, now fixed at head `3f7faea4`.

**`hornerMachine_le_coeff`.** You named it explicitly and I missed it: it lives in `ExpmBridge.lean`, and my sweep only scanned `FPRBridge.lean`. Removed in `570cc8d5`. It was the `.1` projection of `hornerMachine_error` with no caller; the wrap-freedom fact it stated is still proved and used inside `hornerMachine_error` itself.

**Cascading dead code.** The more useful failure. I ran the sweep once, *then* privatised 56 declarations. Both steps change the reference graph — deleting a declaration orphans whatever only it used, and privatising turns a public-with-no-external-consumer into an unreferenced private — so a single pass cannot converge. Re-running to a fixed point removed nine more (77 lines), two of which had appeared in no list from either of us because they only became unreferenced once their sole callers were gone: `abs_addPipeline_yu_sub_lt` and `addPipeline_yu_toNat_mod_two`.

There are now no unreferenced private declarations in any of the four modules, verified at a fixed point rather than after one pass. The 20 public declarations without an external consumer are the intended keep-set: decoder and denotation, domain predicates, structural laws, and the verifier bridges.

Re-validated after the cleanup: full library build, `VCVioTest`, `lake exe axiomsweep --check`, style lint on all four modules, and both isolation checks.

I have also replied to and resolved @dtumad's three threads, which you noted could be closed after the rebase.

dtumad:
Deep semantic audit across the open PR set (semantics only, no style content). **Verdict: no semantic blocker — strongest PR in the lattice family; ready once CI is green on the current target.**

Verified directly at head `3f7faea`:

- **The decode is total and the non-finite escape is properly gated.** `FPR.decode`/`Bits.toReal` denote Inf/NaN as `0`, and every error bound is conditioned on `FPR.IsNormalOrZero`, which excludes `exponent = 2047` — so the `0` denotation can't silently leak into a bound. One consumer-facing trap worth a sentence in the class docstring: a downstream statement about `toReal (someOverflowingComputation)` written *without* the `Valid` gate is a theorem about `0`. Consumers must thread `Valid` everywhere.
- **The bounds are about the arithmetic the implementation actually runs**: `add`/`mul`/`div`/`sqrt` are Pornin's constant-time emulation ported from `sign_fpr.c` on `abbrev FPR := UInt64`, not a redefined arithmetic. `ε = 2^-52` is conservative vs correctly-rounded binary64; sound direction.
- The vacuous-`Valid` trap (`Valid := fun _ => False` satisfies every field) is closed by `valid_zero`/`valid_one`/`inRange_of_valid`, and the earlier refutable instance is now a committed counterexample (`not_add_isNormal_closed`). Good pattern — this is what the rest of the family should imitate.

Coordination:

- This **supersedes the `FPRBridge` cluster in #471** (which still records `interp_zero`/`neg_exact` as not-provable-as-stated) — #471 should drop that cluster entirely.
- Scope guard for readers: this PR does *not* close the ffSampling-composition gap that #466's `ε_step` needs (composing these per-op bounds through the sampler recursion to a TV/Rényi bound, [Jia+26] Thm 2 shape). It is deferred explicitly and correctly; nobody should cite this PR as establishing it.

alik-eth:
Both conditions from the 09-02 pass are met at head `0d8aa97e`:

- **CI is green on the current target**: 8/8 checks pass (Build Project, Lint Style, Library File Imports, Downstream Consumer Build, Interop TCB Isolation, Complexity Backend, Agent Documentation, summarize), against `main` on Lean 4.33.1 with the measure-first semantics.
- **The `Valid`-gate trap is now in the `HasRealSemantics` class docstring** (`Extern/Falcon/ApproxArith.lean`, `0d8aa97e`): a denotation is total, so the `FPR` instance sends Inf and NaN to `0`; a downstream statement about `interp` of an overflowing computation written without a `Valid` gate is therefore a true theorem about `0`, not about the value the algorithm produces. Consumers must thread `Valid`.

On the two coordination points: agreed that this PR does not establish the ffSampling-composition bound `#466`'s `ε_step` needs, and the scope guard in the module docstring says so; the `FPRBridge` cluster in `#471` is @alexanderlhicks's to drop.

Nothing else is open on this side; it is ready to land whenever #478 is in.

PR #515, opened 2026-08-17
feat: coordinate-wise special soundness implies knowledge soundness (ePrint 2023/846 §7)

dtumad:
Formalizes §2.9 and §7 of Fenzi–Moghaddas–Nguyen, *Lattice-Based Polynomial Commitments* ([ePrint 2023/846](https://eprint.iacr.org/2023/846)): coordinate-wise special soundness, and its two extraction lemmas.

Stacked on #525 and #526, which carry the general machinery. This PR currently contains their commits too; it will shrink when they land.

## Status against the paper

| Clause | Lemma 7.1 | Lemma 7.2 |
|---|---|---|
| expected query count | ✅ for the paper's algorithm | per level only |
| success probability | ✅ for the paper's algorithm | ✅ for the paper's recursion |
| output structure | ✅ for the paper's algorithm | ✅ for the paper's recursion |

Everything holds against a *fixed acceptance table*, which is the setting §7.1 needs — its own analysis reasons about `Xᵢ = |{x ∈ S : V (C x) (A (C x))}|` and uses `Pr[V = 1 ∣ Xᵢ = l] = l/N`, both of which presuppose that the adversary is a function of the challenge. `CoordinateFork/Realizability.lean` says where such a table comes from.

## The pieces

**Combinatorics** (`ToMathlib/Combinatorics/`). `SS(S, ℓ, k)` as `IsCoordSpecialSound`, with `le_card_of_hasCoordNeighbours` showing the cardinality conjunct is forced rather than assumed; the counting core `sub_div_le_div_card_filter`; the weighted column count `card_mul_sum_div_columnCount_le`; challenge trees; and §7.3's `pow_add_one_le_tValue`, the lower bound on the `t`-value that is why §7 needs its own extractor.

**The table model** (`CoordinateFork.lean`, `EvalDist/CoordinateFork.lean`). The bound averaged over an arbitrary distribution of acceptance tables. Only the *marginals* of that distribution appear on the left, which is what makes the transfer below cheap.

**Realizability** (`CoordinateFork/Realizability.lean`). For an adversary `A : (ι → S) → ProbComp Y`, `indepTable A` runs it independently at every challenge and `acceptRatio (acceptTable V (indepTable A)) = advSucc V A`. `probOutput_acceptTable_indepTable_eq_bernoulliTable` computes the *whole* joint law and finds the independent Bernoulli table — which is what makes the multi-round coupling realizable rather than assumed.

**The algorithm** (`CoordinateFork/Operational.lean`). `coordForkOp` is Figure 11 itself: sample a challenge, and on acceptance resample each coordinate without replacement until `k − 1` further accepting values turn up or the coordinate is exhausted. It succeeds with exactly the total-lookup core's probability — the draw order cannot matter, because it stops only on success or exhaustion — and returns a coordinate-wise `k`-special sound set of accepting challenges. Its expected lookup count is `1 + ℓ(k−1)`, by a column count rather than a conditional-probability argument.

**The recursion** (`CoordinateFork/MultiRoundOp.lean`). `multiForkOp` is §7.2. Its success probability *is* `multiSucc`, the analytic functional of `MultiRound.lean` — so the independent Bernoulli coupling that functional was defined by stops being a modelling decision: running the sub-extractor independently at each first challenge is what produces it.

## What is not proved

- **The multi-round query count composes only per level.** The paper's `(ℓ(k−1)+1)^μ` multiplies the per-round count by the sub-extractor's, which is Wald's identity over the resampling loop: `𝔼[∑_{i≤T} Xᵢ] = 𝔼[T]·𝔼[X]` needs `{T ≥ i}` independent of the `i`-th sub-run. That holds because the loop picks which challenge to look at *before* reading it — but the eager `Fintype.mPi` table, which is exactly what makes the coupling derivable for the success bound, cannot express "queried before read". Closing it means letting the loop consume its table through an oracle, at which point the count is `expectedQueries` rather than returned data.
- No security-parameter experiment, so this is not yet Definition 2.28 / Lemma 2.31 as stated.
- §7.3 combines a formalized lower bound on `tᵧ` with an upper bound *cited* from [AFR23]; the file says so twice.
- §8 (Fiat–Shamir) is out of scope here. Note for later: it is not self-contained either — Lemma 8.1's success half explicitly reuses a bound from Attema–Fehr–Klooß.

`docs/agents/forking.md` carries the full status and the reasoning above.

## Verification

Full build, `lake exe lint-style`, `scripts/check-agent-docs.py`, and four CI-wired regression files pass. Every headline result is axiom-clean (`propext`, `Classical.choice`, `Quot.sound`). No new sorries. The regressions include payload-sensitivity, coupling-sensitivity at equal marginals, a deliberately-wrong extractor, a pool where the closed form is false, and a tight concrete instance of the query bound.


dtumad:
## Adversarial audit follow-up

The follow-up changes are in [`36c93a9e`](https://github.com/Verified-zkEVM/VCVio/commit/36c93a9ecc67254364f79559c9436a8baa61b682). The title and description now use a narrower coverage claim: this PR formalizes ingredients for the coordinate-wise forking analysis in ePrint 2023/846 §7, not a complete oracle-level extractor or the paper's full knowledge-soundness result.

### What changed

- Made the extractor statement an explicit input, so extraction cannot mix transcripts from different statements.
- Narrowed the single-round result to the finite table model actually formalized: exact output characterization plus the corresponding success-probability bound.
- Described `multiSucc` as an analytic recurrence under an independent full-table Bernoulli coupling; it is not presented as an executed multi-round extractor.
- Corrected the AFR interpretation: the proved structural estimate is the expected-query **upper** bound `2 * tΓ - 1`, not `2 ^ tΓ - 1`. The t-value lower bound alone does not imply a runtime lower bound.
- Made the challenge-tree claim explicit: the current tree records only the challenge projection, not full commits/responses or extractor state.
- Removed unrelated positional-supply/`ToPolyFun` changes and the incomplete negative-hypergeometric track from this PR.
- Added adversarial regressions for a deliberately bad extractor, equal marginals with different joint fork behavior, nonconstant tables, `k = 0/1`, empty coordinates, `k > |S|`, the `μ = 2` recurrence, and payload sensitivity. Both forking regression modules now run explicitly in CI.

### Current proof boundary and limitations

- **Single round:** table success/output theorems only. There is no implemented oracle adversary/extractor, transcript replay semantics, or oracle-query-cost theorem.
- **Multiple rounds:** `multiSucc` is a closed analytic recurrence using independent Bernoulli full-table coupling. There is no executed multi-round extractor or theorem identifying its output distribution with that recurrence.
- **Special soundness:** fixed-statement, extensional extraction-success inequality. It does not yet include a security parameter, a joint bad-event/collision term, a PPT extractor, or the complete knowledge-soundness reduction.
- **Trees:** only challenge projection is represented. Full transcript payloads and adaptive execution state are not encoded by `ChallengeTree`.
- **Query complexity:** `tValue` has structural monotonicity/cardinality lemmas and an expectation upper bound. These do not prove an algorithmic lower bound on actual extraction runtime.

Clear next paths are to connect the finite table model to a concrete replaying `OracleComp` extractor, prove the required coupling/distributional bridge for the multi-round recurrence, and then add the missing security-parameter, bad-event, and complexity hypotheses needed for the paper-level knowledge-soundness statement.

### Validation

- Full non-test library build, `VCVioTest`, targeted forking modules, smoke test, style lint, documentation checks, import/isolation checks, and generated-import checks all pass locally.
- The headline declarations use only Lean/Mathlib's standard logical axioms (`propext`, `Classical.choice`, and `Quot.sound`). No `sorry`, `admit`, new axiom, or `unsafe` declaration was added in the changed files.
- All PR checks are green, including the full project build and downstream no-submodule consumer build.

The PR remains a draft because the remaining paper-level bridges above are real scope gaps, not merely presentation work.

quangvdao:
Draft review at exact head `e0cc7f679b526e6ccc8e284bb72bf810f4d1fe18`. No source edits or ready-for-review transition made.

### Findings

**[P2] Do not identify the returned weighted charge with the paper's oracle-query clause yet.** [`MultiRoundOp.lean:269–276`](https://github.com/Verified-zkEVM/VCVio/blob/e0cc7f679b526e6ccc8e284bb72bf810f4d1fe18/VCVio/CryptoFoundations/CoordinateFork/MultiRoundOp.lean#L269-L276) eagerly executes a recursive extraction at *every* first challenge through `Fintype.mPi`, then returns only the charges of the entries the outer fork selects. The theorem at lines 383–390 bounds that returned projection; it does not instrument the preceding eager computation or establish an equivalent lazy oracle implementation. For example, at one coordinate, four challenge values, depth two, and an always-rejecting table, the eager recursion contains 16 base-level calls while the returned charge is 1. This is a source-level operational counterexample to equating the two costs, not a counterexample to the proved weighted inequality.

[FMN Lemma 7.2](https://eprint.iacr.org/2023/846.pdf#page=72) asks for an oracle algorithm with an expected bound on queries to the adversary. The file's headline “all three clauses … hold of the same computation” and “This is the expected-query clause” therefore overstate the formal connection. The caveat that this is only a charge helps, but does not discharge the missing operational theorem. Retain the useful weighted bound, describe it consistently as a table-model charge bound, and reserve the oracle-query claim for an instrumented lazy extractor plus a joint output/cost correspondence. Merely equating first-component distributions is insufficient for the cost claim.

**[P2] Independent per-challenge resampling is not generally the original prover's shared fixed-coins table.** [`Realizability.lean:19–23,61–64`](https://github.com/Verified-zkEVM/VCVio/blob/e0cc7f679b526e6ccc8e284bb72bf810f4d1fe18/VCVio/CryptoFoundations/CoordinateFork/Realizability.lean#L19-L23) describes `indepTable A := Fintype.mPi A` as the table obtained by fixing the adversary's coins once and for all. A prover that ignores its challenge and returns one fair random bit gives a simple distinction: with one shared fixed tape, its answers at two distinct challenges are always equal; this product construction gives unequal answers with probability 1/2. Both have the same single-challenge marginals. The proved marginal transfer and Bernoulli joint law are valid for the *constructed independent table*, but they do not identify the original shared-tape rewind coupling. Please describe the construction as independent randomness pre-sampled per challenge (or an explicitly transformed adversary), not the original prover with its tape fixed. The caveat at lines 42–45 already points in the right direction and should be used consistently.

There is also status drift to reconcile while doing this: the PR body still says only per-level cost is proved and §8 is out of scope, whereas the head adds `multiForkOpW` and `SamplingGame`; the agent guide advertises all clauses and later retains older “missing” claims. #525 and #526 are already merged, so the stack paragraph is historical too.

### What the reviewed contracts do establish

The extensional special-soundness predicate fixes both statement and commitment. `Extracted` tests a returned valid witness, not mere termination; the concrete bad-extractor and payload-sensitive controls are meaningful. The multi-round result constructs accepting **challenge** trees, with the transcript/prover-message projection limitation explicitly represented by `IsChallengeTree`. The weighted accounting proof uses a conditional table bound and averaging; I found no reason to weaken that mathematical result. The existing equal-marginal/different-coupling tests correctly demonstrate why joint laws matter.

This pass examined the security/extraction interfaces, single/multi-round operational and weighted definitions, recurrence/realizability bridges, challenge-tree shape, principal regressions, relevant documentation, paper clauses, and prior audit follow-up. It is not a line-by-line certification of every supporting combinatorial proof. Local documentation checks passed. I did not rebuild this stale draft or reproduce its historical axiom/build claims; the current head exposes only a successful summary check, not a current green full build. A current-main port, full validation, and the operational/coupling limitations above remain necessary before merge readiness.

PR #516, opened 2026-08-17
refactor(SigmaProtocol): split the interaction out as ChallengeVerifyProtocol

BoltonBailey:
Human summary: This PR is to split out the simulation/extraction component of the SigmaProtocol structure as an extension of a simpler class of protocols that may not have these features.

----

Addresses @dtumad's review point on #450: rather than a parallel protocol hierarchy, factor the shared interaction out of `SigmaProtocol` so both #450 and #418 can build on one record.

## What changes

Adds `ChallengeVerifyProtocol` — the bare commit–challenge–response interaction (`commit` / `respond` / `verify`) — and makes `SigmaProtocol` extend it with the existing `sim` and `extract` fields. Consumers that only run a protocol can take the base record; those that extract keep taking `SigmaProtocol`.

Properties concerning the interaction alone move to the base: `PerfectlyComplete`, `realTranscript`, `HVZK`, `PerfectHVZK`, `perfectHVZK_iff_hvzk_zero`, `simCommitPredictability`, `simChalUniformGivenCommit`, `UniqueResponses`, and `toIdenSchemeWithAbort` with its `Coe`. `SpeciallySoundAt` / `SpeciallySound` / `extract_sound_of_speciallySoundAt` stay on `SigmaProtocol`, since they consume `extract`.

No fields are removed and no proofs change — this is purely a restructuring.

## Why the churn is small

Lean's dot notation resolves through the parent projection, so `σ.commit`, `σ.HVZK`, `σ.simCommitPredictability` and friends keep working unchanged on a `SigmaProtocol`. Fischlin, the whole Fiat-Shamir Σ-layer (`CmaToNma`, `Fork`, `Reductions`, `Security`, `Chain`), and every `variable (σ : SigmaProtocol ...)` site are untouched. Only explicitly qualified references needed updating:

- one `simp` lemma name in `FiatShamir/Sigma/Stateful/Hops.lean`
- unqualified references in the Schnorr example, switched to dot notation

## Scope

Deliberately kept minimal and on `ProbComp`:

- **not** generalizing over a monad — that stays #418, which rebases onto this
- **not** removing the currently-unused `sim` field — separable cleanup
- **not** adding new properties — #450 adds its soundness notions to the base on rebase

Then #450 can delete its own parallel copy of this structure and instantiate the shared one.

## Verification

`lake build VCVio Examples` passes: 0 errors, no new warnings, no new `sorry`s.

### DIFF
--- a/Examples/Schnorr/SigmaProtocol.lean
+++ b/Examples/Schnorr/SigmaProtocol.lean
@@ -100,7 +100,7 @@ omit [Fintype F] [DecidableEq F] in
 /-- Perfect completeness: an honest prover with a valid witness always produces
 an accepting transcript. Follows from `add_smul` and `mul_smul`. -/
 theorem sigma_complete (g : G) :
-    PerfectlyComplete (sigma F G g) := by
+    (sigma F G g).PerfectlyComplete := by
   intro pk sk h
   have h_eq : sk • g = pk := of_decide_eq_true h
   simp only [sigma, monad_norm]
@@ -140,7 +140,7 @@ omit [Fintype F] [DecidableEq F] in
 The proof swaps sampling order and uses uniformity of `F` to reindex via the bijection
 `r ↦ r + c * sk`. -/
 theorem sigma_hvzk (g : G) [Finite F] :
-    PerfectHVZK (sigma F G g) (simTranscript F G g) := by
+    (sigma F G g).PerfectHVZK (simTranscript F G g) := by
   let _ : Fintype F := Fintype.ofFinite F
   intro pk sk h_sk
   have h_eq : sk • g = pk := of_decide_eq_true h_sk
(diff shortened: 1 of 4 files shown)


dtumad (approved):
Final review before merge. The factoring is faithful — no fields removed, interaction-only properties (`PerfectlyComplete`, `realTranscript`, HVZK/PerfectHVZK, `UniqueResponses`, the `IdenSchemeWithAbort` coercion) move to the base record, and `SpeciallySound`/`extract_sound_of_speciallySoundAt` stay on `SigmaProtocol` where `extract` lives. Verified no library outside VCVio/Examples consumes the moved `Coe`, and `lake build VCVio Examples` passes clean on a current-main merge. The module docstring's framing of the two literature conventions (existential extractor vs bundled data) is a nice touch. This unblocks #450 and #418 as intended.

PR #511, opened 2026-08-08
feat: define VCVio’s public boundary over PolyFun

dtumad:
## Stack

- base: #510
- PolyFun dependency: Verified-zkEVM/PolyFun#114 (`134eca434dc21d692737f5daa2b05aa189ca2dfa`)

This PR is intentionally stacked on the module-system migration so its diff is limited to the long-term PolyFun/VCVio boundary work.

## Summary

- add generic `PFunctor` probability semantics in `VCVio.EvalDist.PFunctor`, with `OracleComp` delegating through the existing VCVio façade
- move query-handler instrumentation to PolyFun and retain thin `QueryImpl` aliases for VCVio consumers
- replace downstream reliance on PolyFun implementation imports with public laws, including interaction and UC clients
- document the intended boundary: expose foundational `PFunctor` concepts where they are useful, while keeping probability and cryptographic policy in VCVio
- add a compile-time façade canary and CI check that rejects `import all PolyFun...`
- pin the paired PolyFun draft commit so the stack is reproducible

## Design direction

This takes a hybrid approach rather than making VCVio uniformly thinner or uniformly more opaque:

- generic free-monad operations and handler instrumentation belong in PolyFun
- VCVio should expose those concepts when they improve interoperability
- VCVio continues to own `OracleSpec`, `OracleComp`, probability-spec policy, notation, and compatibility aliases used by cryptographic developments
- downstream code should depend on public laws, not the reducibility of PolyFun internals

## Validation

- `lake build` (3,007 jobs)
- `lake build ToMathlib VCVio LatticeCrypto Extern HashSig Examples VCVioWidgets` (9,214 jobs)
- `scripts/check-polyfun-boundary.sh`
- `lake env lean VCVioTest/PFunctorFacade.lean`
- `lake env lean VCVioTest/Smoke.lean`
- `lake env lean LatticeCryptoTest/Falcon/Main.lean`
- `git diff --check`

Issue #145, opened 2026-03-10
Proof obligation for `hashedElGamal_indcpa_bound` in `Examples/HashedElGamal.lean`


quangvdao:
A proof in `Examples/HashedElGamal.lean` contains a `sorry`.

**Goal:** Replace the `sorry` with a complete proof.

[Link to the sorry on GitHub](https://github.com/Verified-zkEVM/VCV-io/blob/master/Examples/HashedElGamal.lean#L167)

**Code Snippet:**
```lean
theorem hashedElGamal_indcpa_bound
    (adv : AsymmEncAlg.IND_CPA_Adv (hashedElGamal F g hash)) :
    (AsymmEncAlg.IND_CPA_OneTime_AdvantageENN (encAlg := hashedElGamal F g hash) adv).toReal ≤
      ddhDistAdvantage g (ddhReduction (F := F) (hash := hash) adv) +
      EntropySmoothing.advantage F g hash (esReduction (F := F) (g := g) adv) := by
  sorry
```

PR #521, opened 2026-08-18
chore: update to Lean 4.33 and align the PolyFun/VCVio API boundary

dtumad:
## Summary

- update Lean, Mathlib, cslib, and the dependency manifest to the 4.33 line
- pin the coordinated PolyFun `v4.33.2` release containing the public module, realizability, UC-family, representation-invariance, and codability foundations
- define VCVio's public PolyFun boundary, including the `EvalDist.PFunctor` façade, import canaries, CI boundary checks, and module-system documentation
- migrate OracleComp, Fiat–Shamir, Merkle tree, UC, program-logic, example, and widget clients to public APIs
- expose stable `SetM`, `SPMF`, WriterT, query-cache, and query-seed equation surfaces
- consume upstream `DynComputation.unroll_wrap` and migrate seeded handlers/contracts from raw `Function.update` to public `QuerySeed` laws

## Reconstruction

This branch is reconstructed directly on current `main`. It does not replay the API, lattice, axiom-sweep, or MLDSA changes already merged through #519, #520, #522, #528, and #529. It preserves the unique 4.33/module-boundary work from the old stack and resolves the previously conflicting PR in place.

This also incorporates and supersedes the boundary work from #511; after this PR merges, #511 can close without losing its canaries, scripts, or documentation.

## Review follow-ups

The [downstream audit](https://github.com/Verified-zkEVM/VCVio/pull/521#issuecomment-5376790615) against [ArkLib #791](https://github.com/Verified-zkEVM/ArkLib/pull/791) found four defects and one API-shape problem. All are addressed:

| Finding | Resolution |
| --- | --- |
| `grind` regressions in `VCVioTest/MonadProbability.lean` and `VCVioTest/GrindFailFast.lean` | `attribute [implicit_reducible] OracleSpec`. Both canaries pass **unmodified**. |
| CI missed them | `lake build VCVioTest` replaces five hand-maintained `lake env lean` canary steps. |
| Append/simulation laws pinned to `Type 0` | `Append.lean` generalized to arbitrary index universes over a shared response universe; the two missing `addLift` laws added; nonzero-universe canaries added in-repo and in the downstream scratch consumer. |
| PolyFun's generically named axiom-sweep fixtures | Fixed upstream in [PolyFun #145](https://github.com/Verified-zkEVM/PolyFun/pull/145); this PR re-pins to the resulting `v4.33.2`. |
| Append laws straddling root and `QueryImpl` | Hard cutover into `QueryImpl`. No deprecated aliases — `gotchas.md` §18, and there were no call sites outside `Append.lean`. |

Three corrections to the audit report, detailed in the [reply](https://github.com/Verified-zkEVM/VCVio/pull/521#issuecomment-5376790615):

- the `grind` failures are caused by Lean 4.33's canonicalizer unfolding reducibles past the non-reducible `OracleSpec` wrapper, **not** by the `instance` → `abbrev` façade demotion;
- `GrindFailFast.lean` failed at three sites, not one, and the third is an unguarded bare `grind`;
- the `change`-glue cleanup landed at 2 of 24 sites; the remaining 22 are definitional steps under a dependent `hstep` binder and cannot be moved into a statement.

## API-boundary audit

- no `allowUnsafeReducibility`
- no new `sorry` or `admit`
- no downstream redeclaration of the now-upstream `unroll_wrap`
- wrapper normalization is behind named public equations rather than `change` or rewrite-then-`rfl` at client seams
- query-seed clients use `QuerySeed.update`, `update_comm`, and append/update laws rather than raw dependent `Function.update`
- sum-spec reducibility remains declared once at its public instance
- public simulation-composition laws are universe-polymorphic and canaried from outside the package
- `VCVio/EvalDist/PFunctor.lean` opens `@[expose] public section`, matching the rest of that layer

## Validation

- `lake build ToMathlib VCVio LatticeCrypto HashSig Examples VCVioWidgets Extern VCVioTest` — 9,286 jobs
- `lake exe lint-style ToMathlib VCVio Extern LatticeCrypto Examples VCVioWidgets Interop`
- PolyFun boundary, Interop TCB isolation, and Extern isolation checks
- axiom-sweep fixture matrix
- `lake exe axiomsweep --check` — 14,192 declarations across 444 modules, no new taint
- smoke, round-by-round, without-replacement, Merkle batch, computability, PFunctor-façade, universe-polymorphism, probability-tactic, long-chain, monad-probability, and grind fail-fast canaries — now all covered by the `VCVioTest` library build rather than an enumerated list
- PolyFun `v4.33.2` validated upstream with `--lint --test --axioms`
- ArkLib #791 rebuilt against this head (`63f449fd`) with all six local `simulateQ_*` adapters deleted — full build 4,468 jobs, `validate.sh --axioms` clean at 10,486 declarations


alexanderlhicks:
Thanks!

dtumad:
I tested the current head, `16628cbdee8e8aff28a065063426e52e79390514`, as the VCVio dependency of a complete ArkLib Lean 4.33 baseline. The baseline pins PolyFun `ee25209d0b74bb9c6afc4220789070b5dfbc9c13` (v4.33.1) and mathlib `db584cd6d46c92f209a44c0f1c829460d327499d`.

The good news is that the downstream migration completes: ArkLib's full build, warning gate, axiom regression sweep, and docs build pass. Several downstream proofs became shorter by using the new public laws (notably FFT subdomains, Johnson-bound reuse, Batched FRI cancellation, and VCVio lifting laws). I also ran VCVio's libraries/canaries, axiom sweep, style lint, PolyFun boundary/interop/isolation checks, and native KATs; those pass. PolyFun's complete `--lint --test --axioms` validation passes at the pinned revision.

I found these issues at the exact PR head:

1. `lake env lean VCVioTest/MonadProbability.lean` fails at line 178:

   ```lean
   example : Pr[⊥ | (failure : OptionT ProbComp Bool)] = 1 := by grind
   ```

   The `simp` version immediately above succeeds, but `grind` reports that it cannot canonicalize `OptionT.instMonadLiftTSPMF ...` / synthesize `MonadLiftT (OptionT (PFunctor.FreeM ...)) SPMF`. I count loss of the one-line proof as a regression even though there is an easy workaround.

2. `lake env lean VCVioTest/GrindFailFast.lean` fails at line 111. The intended fail-fast canary still verifies that bare `grind` fails, but `grind [probEvent_eq_zero_iff']` now also fails. Its diagnostics similarly fail to canonicalize the transitive `MonadLiftT FreeM PMF SPMF` instance.

   Both modules are imported by `VCVioTest.lean`, but the current green CI checks the generated import file and selected individual canaries without building the `VCVioTest` library/umbrella. Building that target (or adding these two files as explicit canaries) would catch both regressions.

3. The new append/simulation exposure layer is not universe-polymorphic. A minimal consumer with `A : Type uQuery` fails for both
   `QueryImpl.simulateQ_add_liftM_left` and `simulateQ_addLift_add_liftM_left`: Lean expects `OracleComp.{0,0,0}`. `VCVio/OracleComp/SimSemantics/Append.lean` declares the relevant `ι`, `α`, and monads as `Type` / `Type → Type`, despite the file-level universe declaration. ArkLib needs local universe-polymorphic versions of the left/right and three-way laws as a workaround. Please generalize these public lemmas and add a nonzero-universe consumer canary.

4. There is a transitive integration collision from the new PolyFun pin: PolyFun exposes a Lake library and namespaces named generically `AxiomSweepTestFixtures`, which collide with an ArkLib fixture of the same name during `--axioms`. I fixed ArkLib by package-prefixing its fixture as `ArkLibAxiomSweepTestFixtures`, but PolyFun should also package-prefix its non-default fixture so a second downstream package cannot hit the same problem.

One API-shape suggestion from the migration: the two-way append laws live under `QueryImpl`, while `simulateQ_addLift_add_liftM_left/right` are at root. Long-term, I think all implementation-composition laws should live consistently under `QueryImpl`; if these root names move, retaining deprecated forwarding aliases for a release would avoid another downstream hard rename. More generally, the best v4.33 changes were the ones that exposed stable equations while keeping representations opaque/minimally reducible.

Current PR checks are green, so the two failing test modules and the universe consumer are useful additions to the matrix before merge.

dtumad:
Stack update: #531 now uses the generic polynomial semantics and public API boundary established
here to implement the accepted measure/kernel design from #532. No additional change is needed in
this PR: it remains the prerequisite, while all probability policy and compatibility work stays in
the follow-up.

dtumad:
Thanks — building ArkLib against the exact head was the right way to find these. All four defects
and the API-shape point are fixed. Three of your diagnoses need correcting, and one item landed
much smaller than it should have; details below.

## 1 & 2 — the `grind` regressions

**The cause is not the façade demotion.** Synthesis of `MonadLiftT ProbComp PMF` still works fine
through `PFunctor.FreeM.instMonadLiftTPMF`; the `instance` → `@[deprecated] abbrev` change at
`VCVio/OracleComp/EvalDist.lean` is not load-bearing here. On `main` the old
`OracleComp`-keyed instance would have hit the identical wall.

What actually happens is the toolchain bump. Lean 4.33 reworked `grind`'s canonicalizer
(`Lean/Meta/Sym/Canon.lean`) so instance arguments are re-synthesized *from the canonicalized
type*, and `Sym.unfoldReducible` first unfolds every `@[reducible]` declaration in the goal. Every
link is reducible:

```
OptionT ProbComp
  → OptionT (OracleComp unifSpec)                                    -- ProbComp is `abbrev`
  → OptionT (PFunctor.FreeM unifSpec.toPFunctor)                     -- OracleComp is @[reducible]
  → OptionT (PFunctor.FreeM { A := ℕ, B := fun n ↦ Fin (n + 1) })    -- toPFunctor/ofFn/unifSpec reducible
```

At that erased type synthesis dies:

```
failed to synthesize  { A := ℕ, B := fun n ↦ Fin (n + 1) }.IsProbabilitySpec
failed to synthesize  MonadLiftT (OptionT { A := ℕ, B := fun n ↦ Fin (n + 1) }.FreeM) SPMF
```

The only route back to the registered `IsUniformSpec unifSpec` is
`IsUniformSpec.toIsProbabilitySpec : PFunctor.IsProbabilitySpec (OracleSpec.toPFunctor ?spec)`, and
assigning `?spec := (fun n ↦ Fin (n+1))` requires unfolding `OracleSpec` — a plain non-reducible
`def`. Measured `isDefEq` for `toPFunctor (fun n ↦ Fin (n+1)) ≟ toPFunctor ?spec`: `false` at
`.instances` (what TC synthesis uses), `true` at `.default`. That gap is the whole bug, and it is
why `simp` copes and only `grind` breaks. Corroborating asymmetry: the `SetM` side survives, because
`instMonadLiftTSetM` has no `IsProbabilitySpec` premise and never has to route back through
`toPFunctor ?spec`.

Fix is one attribute on the wrapper (`VCVio/OracleComp/OracleSpec.lean`), no
`allowUnsafeReducibility` needed. This branch already used exactly that idiom locally at
`RandomOracle/EagerTable.lean:41`; promoting it to the definition made that local copy redundant and
it is removed. Both canaries pass **unmodified**, and `GrindFailFast`'s `fail_if_success grind`
guards still hold.

**Second correction: `GrindFailFast.lean` failed at three sites, not one** — 111, 115, and 118. 118
is a bare `grind`, so no guard was hiding it.

## CI gap

You were right that this is the real problem. `mk_all --lib VCVioTest --module --check` only checks
that the umbrella's import list is textually complete — it never elaborates, so `VCVioTest` *looked*
covered. Five of eleven modules were never compiled anywhere in CI: `GrindFailFast`,
`LongChainPrograms`, `MonadProbability`, `ProbabilityTactics`, `SampleableType`. All five open with
`public import VCVio`, which is exactly the surface this branch rewrites.

Rather than add those two files as canaries, CI now runs `lake build VCVioTest` and the five
enumerated `lake env lean` steps are deleted as subsumed. A canary added in a future PR is covered
the moment `update-lib.sh` registers it, and the work is now cached — `lake env lean` writes no
oleans, so those steps were paying full elaboration cost every run and discarding it.
`scripts/build-project.sh` got the same target, since the documented local workflow had the same
blind spot.

Deliberately **not** in scope: adding `VCVioTest/**` to the lint path filter and the warning budget.
Both would open a backlog in files that have never been linted. Filed as a follow-up.

## 3 — universe polymorphism

Generalized. `Append.lean` declared `universe u v w` but pinned four `variable` blocks to `Type` /
`Type → Type`; those now take arbitrary index universes over one shared response universe with `α`
in it, matching what `parallelStateT` / `addReaderT` / `parallelWriterT` / `StateSeparating`
already do. No proof changed — `simulateQ_liftTarget`, `simulateQ_compose`, and `liftComp_eq_liftM`
were already general enough, which your local copies had effectively already demonstrated.

Also upstreamed `QueryImpl.simulateQ_addLift_liftM_left` / `_right`, the two-way `addLift` case with
the components in distinct source monads, which VCVio lacked entirely. To be accurate about it:
ArkLib has **no call sites** for those two — they were added to complete the family, not because
anything needed them.

Canaries added at both ends, as you asked: `VCVioTest/UniversePolymorphism.lean` in-repo, and a
nonzero-universe example in the downstream scratch-consumer CI job so the generality is checked from
*outside* the package.

`docs/agents/gotchas.md` §9 said "keep `α β : Type`", which is what produced this. It now separates
local definitions (pinning is fine) from public reusable laws (a `Type 0`-only statement is not a
replacement for the adapter a consumer would otherwise write).

## 4 — PolyFun fixture collision

Fixed upstream rather than worked around: [PolyFun #145](https://github.com/Verified-zkEVM/PolyFun/pull/145)
prefixes the fixtures library, the sweep module root, and the executable. This PR re-pins to the
resulting `v4.33.2`. Only the fixtures rename fixes an observed break — `lake exe` resolves
root-package-first, so the executable/module renames are precautionary, and the lakefile records
why so they don't get dropped later as noise.

Left alone deliberately: VCVio's own `axiomsweep` executable and `scripts/AxiomSweep.lean` root are
unprefixed too, and so are ArkLib's. That is the same latent class, but nothing is broken by it
today and renaming would churn a documented command in CI, the README, and `AGENTS.md`. Worth doing
in its own PR if you want the convention uniform.

## 5 — namespace consistency

Done as a hard cutover into `QueryImpl` rather than with forwarding aliases:
`gotchas.md` §18 bans compatibility shims, and grep found no call sites outside `Append.lean` —
ArkLib was using its own local copies, not these. Moved `simulateQ_addLift_add_liftM_left/right` and
`simulateQ_optionT_liftM_run_eq_of_query`.

## Where I under-delivered

You asked for the `change`-glue cleanup in full. It landed at **2 of 24 sites**, and the reason is
worth recording rather than glossing.

The two that moved are public API — the `Fork.mem_support_…_iff` characterizations now state their
outcome binder at `Chal` instead of `(unifSpec + _).Range (Sum.inr mc)`, which removes the ascription
for consumers as well as in the proof.

The other 22 are not reachable by restating a lemma. They sit under `preservesInv_layered`, whose
`hstep` quantifies `∀ t, ∀ z ∈ support (… ((unifForward + roImpl) t).run s …)`, so `z`'s type is
dependent on the universally quantified `t`. After `cases t`, reducing
`(unifSpec + _).Range (Sum.inl n)` to `unifSpec.Range n` is a definitional step with nowhere to go in
the statement. I tried restating `support_step_inl` / `_inr` at reduced types; it broke ten `rw`
sites, because `rw` matches syntactically against hypotheses the skeleton produced at the unreduced
type. Reverted. Net on the branch: 71 → 69 added `change`s.

I also mis-called `Fork.lean`'s private `simulateQ_add_run_query_bind` a duplicate of this PR's
`simulateQ_add_query_bind_*`. It is not — it is the `StateT`-run form, which had no public
counterpart at all. It is now `OracleComp.run_simulateQ_query_bind` in
`SimSemantics/StateT/Basic.lean`, the step lemma any `inductionOn` over a stateful simulation needs,
and Fork.lean applies it.

## Two latent defects found on the way

- `VCVio/EvalDist/PFunctor.lean` was the only file in its layer on a bare `public section` rather
  than `@[expose] public section`, so the new PFunctor-level instance bodies were unexported and the
  `@[reducible]` on `IsUniformSpec.ofFintypeInhabited` was inert downstream. Fixed.
- `PFunctor.FreeM.instEvalDistCompatible` is reachable only explicitly, via
  `IsUniformSpec.toPFunctor` at the oracle façade — never by synthesis, since that conversion is
  deliberately a `def`. Correct as designed, but it would become a live ambiguity with
  `OracleComp.instEvalDistCompatible` the day anyone promotes it, so that is now recorded at the
  declaration.
- `OracleComp.support_eq_evalDist_support` is public. It is the named bridge between the two
  semantics, and keeping it private is what forced a `change support _ = SPMF.support 𝒟[_]` at the
  façade.

## One thing the round trip surfaced

VCVio states its routing laws over `impl₁ + impl₂` (`HAdd`), and ArkLib's local copies were stated
over `QueryImpl.add impl₁ impl₂`. Two `change` blocks in ArkLib spelled the handler in `.add` form,
so the upstream `rw` did not fire until they were rewritten to `+`. `QueryImpl.add_eq_hAdd` — new in
this PR — is the intended bridge, and it is what VCVio's own `simulateQ_addLift_add_liftM_*` proofs
use.

Tagging `add_eq_hAdd` `@[simp]` would make `.add` normalize to `+` automatically and would have made
those two ArkLib proofs work untouched. I did **not** do it here: it shifts a global simp normal form
on a PR that is otherwise ready, and it wants its own build. Flagging it as the obvious follow-up if
you want `.add`-shaped goals to route without a manual bridge.

## Round trip

ArkLib #791 is updated against this head with all six local `simulateQ_*` adapters at
`ArkLib/OracleReduction/Basic.lean:434-563` deleted and call sites pointed at `QueryImpl.*` — which
is the actual test of finding 3.

Validation on this branch: 9,286 build jobs including `VCVioTest`; lint-style, PolyFun boundary,
Interop and Extern isolation, and `mk_all` all clean; axiom sweep `check passed (no new axiom/sorry
taint)` over 14,192 declarations in 444 modules.

PR #530, opened 2026-08-21
docs: probability-semantics design record — landscape, verification log, and Mathlib integration shape

dtumad:
## Summary

The written design record for VCVio's probability semantics: what the options were, what is
actually true upstream, and what the resulting statements should look like so that Mathlib's
library applies to them.

Docs only. No Lean changes.

## Contents

| Document | Question it answers |
|---|---|
| `probability-semantics-landscape.md` | What are the options, and what is true upstream? The evidence ledger; §19 is its verification log. |
| `mathlib-integration-shape.md` | What should our probability statements *look like* for Mathlib's library to apply? Short- and long-term shape. |
| `README.md` | Reading order, and the two conventions that keep these documents honest. |

Two further records — the spike findings and the accepted design — describe Lean files that are
not on `main`, so they travel with the measure-semantics work rather than landing here. A reader on
`main` is never pointed at a design whose implementation is absent; the README says so explicitly
rather than leaving the gap to be inferred.

## The case, in one paragraph

`PMF.support_countable` is a theorem, so no `PMF` can carry a measure over an infinite trace space
or a continuous distribution — which is why the coinductive layer is confined to fuel bounds and
why `DiscreteGaussian.lean` hand-rolls what Mathlib already ships. Divergence work written against
`PMF` also cannot be upstreamed now that the type is being retired. Mathlib's timing sets our
schedule, not our case: `PMF` is untouched at `v4.34.0-rc1`, so we convert on our own terms.

## What the integration-shape document concludes

- **Two dated liabilities.** There is no `lintegral` anywhere in `VCVio/EvalDist/` against 555
  `tsum` mentions, so Mathlib's convergence theory is unreachable from a VCVio expectation. And
  `probEvent` is defined through `PMF.toOuterMeasure`, which leanprover-community/mathlib4#42821
  deprecates.
- **Bridge or replace is decided by measurement, not inspection.** Two surfaces went against
  expectation: `tvDist` has 760 uses across 26 files despite living in one file, and `Fintype.mPi`
  has no users outside its own file at all.
- **Against the obvious move on normalization**: do not flip the `tsum` simp set to `lintegral`.
  Finite cryptographic goals genuinely want sums, and Mathlib states its own discrete results the
  same way. The breakage budget belongs on expectation and independence, not on `probOutput`.

## Evidence discipline

§19 records, for each claim, *how it was checked* — the failure mode here is asserting that
upstream provides something on the strength of a name matching. It also now records its own
corrections: an earlier revision's per-file counts were produced with `grep -c`, which counts
matching lines rather than occurrences, and the PolyFun tag reference has gone stale twice in a
single day.

## Validation

- `python3 scripts/check-agent-docs.py`
- All 58 relative links across the reading directory resolve


dtumad:
Context for this survey: #532 (the proposal it supports), and #531 (the prototype it was written to evaluate).

dtumad:
Implementation follow-up: #531 now exercises the survey's main alternatives and records the
resulting accepted baseline. The outcome is stratified rather than a blanket carrier replacement:
closed programs use `Measure`, indexed/stateful semantics use `Kernel`, explicit effects retain
enriched total outcomes, and possibly nonterminating computations expose returned-value
subprobability measures. Existing `Pr[...]` proofs remain the discrete compatibility façade.

I updated this PR's description so it is clear that the survey remains the evidence ledger; the
implementation and final design live in #531 and the roadmap in #532.

quangvdao (approved):
Reviewed at `4957e1f9ebdef2379cf54c7a999bd4f64ac7e78e`.

No blocking findings. I checked the full four-file documentation delta, the distinction between snapshot facts and accepted design, the evidence/verification ledger, the reading-index links, and the proposed Mathlib integration shape. Volatile upstream facts are explicitly date-scoped rather than presented as timeless repository state, and the two not-yet-landed companion records are deliberately not linked as present files.

Validation: exact-head hosted CI is green, including the clean project build and documentation gate; I also ran `scripts/check-agent-docs.py` against the frozen head locally and it passed. This is ready to merge.

PR #154, opened 2026-03-11
feat: add proof-driven game-hop widgets

quangvdao:
## Summary
- add a `VCVioWidgets` Lean library for infoview game-hop panels with source extraction, reveal anchors, and deterministic rendering
- move game-hop authorship next to the proofs and replace hidden widget-side diagram tables with `@[game_hop_root]` registration plus root-driven diagram inference
- keep the panel responsive during elaboration and add a small proof-local hint vocabulary for later refinement without reintroducing central hardcoded diagrams

## Test plan
- [x] `lake build`
- [ ] Open `Examples/OneTimePad.lean` in the infoview and verify the inferred game-hop chain and clickthrough targets
- [ ] Open `Examples/ElGamal.lean` in the infoview and verify the inferred game-hop chain and clickthrough targets
- [ ] Confirm GitHub Actions checks pass on the updated PR

Made with [Cursor](https://cursor.com)

PR #542, opened 2026-08-24
fix(oraclespec): restore coproduct transparency on Lean 4.33.1

### LINKED ISSUE #541: Lean 4.33 migration: composed OracleSpec Range does not reduce, blocking four PRs
`main` is on Lean 4.33.0 as of `89e48210`. Four open PRs — #466, #478, #507, #508 — still build against 4.32.2, so their green checks certify a toolchain the repo no longer uses. Merging `main` into each is **clean**; no textual conflicts anywhere. The proofs are what break.

I migrated as much as is mechanically fixable. **19 fixes across two branches**, and all four PRs then converge on **one** blocker that is a design question rather than a migration patch.

## Root cause

`OracleSpec` addition is `Sum.elim`, so `((a + b) + c).Range (Sum.inl (Sum.inr t))` reduces to `b.Range t` only through unfolding that 4.33 no longer performs at the transparency tactics use. Once a proof unfolds a composed handler, the goal is **ill-typed at implicit transparency**, and tactics stop matching patterns that are plainly visible in the printed goal — `rw` reports "did not find the pattern" against a hypothesis displaying exactly it.

The symptoms look unrelated but are all this one thing: `unsolved goals`, `tauto failed`, `rewrite failed`, `subst failed`, `application type mismatch`.

`main` itself builds fine, so the core layer is healthy in isolation. It is specifically proofs that unfold *through* the sum.

## The blocker: `Range` leaking into a lambda's expected type

Everything else has a tactic-level fix. This does not, because no tactic can rewrite the expected type of a lambda being passed to a function.

#478 and #466, `GameIdentification.lean:179` — a handler passed to `appendInputLog`:

```
fun msg ↦ gpvRealImpl psf hr M Salt pk sk (Sum.inr msg)
  has type      (msg : (ofFn …).Domain) → StateT … (((unifSpec + …) + ofFn …).Range (Sum.inr msg))
  but expected  QueryImpl (ofFn fun x ↦ Salt × Domain) (StateT …)
```

#507, `BodyBounds.lean:391` — a `probEvent` predicate:

```
fun z ↦ z.2.2 = true
  has type      Chal × GhostState M Commit Chal → Prop
  but expected  (…).Range (Sum.inl (Sum.inr mc)) × GhostState M Commit Chal → Prop
```

Same shape, different position. Fixing it means either making the composed `Range` reduce at the transparency tactics use, or phrasing these handlers so it does not leak into a `QueryImpl` / predicate position. Both are decisions about the `OracleSpec` layer, with callers, so I would rather not guess.

## What is already fixed

| branch | fixes | clears |
| --- | --- | --- |
| `gpv188-generic-r433` (`3dc6349b`) | 14 | `GameRuns`, `FlagHandlers`, `CombinedHandler` — for #478 and #466 |
| `fswa-cma-nma-engine-r433` (`6ea617eb`) | 5 | `GhostLayer`, `Projections` — for #507 and #508 |

Six recurring shapes, all mechanical:

1. composed handler will not unfold → `change` to the explicit form, then the named rung (`QueryImpl.simulateQ_add_liftM_query_right`); this is the idiom `gpvRealImpl_run_sign` already used
2. `id` survives under `do` binders → append `rfl`
3. `tauto` cannot refute a tuple's `false = true` → project the component out, since `simp` cannot destructure a pair whose type carries the unreduced `Range`
4. `mem_support_bind_iff` applied to a map → `mem_support_map_peel` first
5. on a uniform query the support is `univ`, so `f '' univ` now normalises to `Set.range f` → the destructuring loses a binder. Only uniform branches; sign branches are genuine images and keep three
6. rewrites by `rfl`-lemmas that no longer match are usually **removable** — whatever they rewrote was definitional anyway

One ordering constraint worth knowing, which cost the most time: in `FlagHandlers`' random-oracle branch the map must be flattened **before** rewriting with `gpvRealImpl_run_read`. That lemma's sides are defeq but not syntactically equal — LHS type `Range (Sum.inl (Sum.inr mc))`, RHS `Range mc` — so rewriting first strands the goal and no simp-set tuning recovers it. Reordering two lines fixed it.

## Hypotheses tested and refuted

So nobody repeats them:

- **`Sum.elim` being semireducible is not the cause.** Tested with `attribute [local implicit_reducible] Sum.elim`; the attribute took and the errors were unchanged.
- **Type-ascribing the right-hand side does not help.** Elaboration discards the ascription.

## Caveat on the counts

Lake halts a module at its first error, so each module's breakages only surface once the previous one is fixed. The counts above are what surfaced, not a proven total — there may be more behind the blocker in each branch.

dtumad:
## Summary

- Upgrade the Lean, Mathlib, and inherited cslib baseline coherently to v4.33.1, and pin the merged PolyFun coproduct fix from Verified-zkEVM/PolyFun#154.
- Restore nested dependent coproduct normalization at implicit transparency by making the `OracleSpec` `HAdd` a plain instance that calls `PFunctor.sum` directly.
- Add exact regression canaries for the outer-right `QueryImpl.appendInputLog` and middle-branch probability-notation failures reported in #541, alongside the proof normalization and explicit import adjustments required by the 4.33.1 dependency surface.
- Document the PolyFun–OracleSpec boundary, Lean module visibility versus transparency, and related Lean core and Mathlib patterns.

## Root cause

`OracleSpec` previously marked its `HAdd` instance `implicit_reducible`, while PolyFun represented coproduct directions through the ordinary semireducible `Sum.elim` wrapper. Nested dependent response types therefore had to cross both `HAdd` projections and `Sum.elim` before reaching the primitive recursor, and could stop normalizing at `.implicit` transparency on Lean 4.33.

PolyFun #154 moves that normal form to primitive `Sum.rec`. On the VCVio side, leaving `HAdd` at the ordinary `instance_reducible` status supplied by `instance` gives implicit and instance checking the intended reduction without making combined specifications unfold during ordinary tactic matching. `OracleSpec` remains a lightweight dependent-family facade over `PFunctor` rather than becoming a wrapper structure.

## Dependency

Verified-zkEVM/PolyFun#154 is merged. This branch pins its canonical merge commit, `bb7e1003544b8461e3d484fc51c61b2ff865250f`; that commit has the same source tree as the previous pre-merge pin. There is no post-fix PolyFun release tag yet, so the commit pin remains intentional and reproducible.

Closes #541.

## Validation

- `lake update --keep-toolchain` (idempotent) and `lake exe cache get`
- `lake build ToMathlib VCVio LatticeCrypto Extern HashSig Examples VCVioWidgets`
- CI-equivalent fresh production build plus `check-warning-log.py` (`No repository non-sorry warnings found`)
- `lake build VCVioTest`
- `lake env lean VCVioTest/PFunctorFacade.lean`
- `mk_all --module --check` umbrella checks for every production library and `VCVioTest`
- `lint-style` across the active libraries
- PolyFun/PMF boundary and Extern/Interop isolation checks
- `scripts/test-axiomsweep.sh`
- `lake exe axiomsweep --check` (14,329 declarations across 459 modules; no new axiom or `sorry` taint)
- `git diff --check`

### Issue #541 downstream reproductions

- GPV/Falcon: composed the finalized PR with `gpv188-generic-r433` (`3dc6349b81b6ed6b4eb0dfdd6573dc6efcb0a814`) and with #466. The actual `gpvRealImpl ... (Sum.inr msg)` lambda expected by `QueryImpl.appendInputLog` typechecks in both compositions.
- FSWA/ML-DSA: composed the finalized PR with `fswa-cma-nma-engine-r433` (`6ea617ebe72636250f3ed12940cf2a38c896f765`) and with #508. The original middle-branch `.inl (.inr mc)` probability expression typechecks unchanged in both compositions; the containing `BodyBounds` module builds after deleting one now-redundant trailing `rfl` in the disposable 4.33.1 test branches.

The broader downstream branches still contain separate Lean 4.33.1 proof-script migration residuals: GPV/Falcon stop in `GPVHashAndSign/FlagHandlers.lean` on tactic/projection/support proof shapes, while FSWA/ML-DSA stop later in `WithAbort/Security/NMAReduction.lean` on a rewrite that no longer matches. Neither residual reproduces the nested coproduct expected-type failures fixed here.


dtumad:
@alik-eth Can you verify yourself that this resolves your issue in #541

alik-eth:
Verified. **#542 resolves #541** — both blockers I reported there are gone, and I checked against the *original unmigrated proofs* rather than my patched branches, so the result isolates your fix.

## Method

Worktree at `6e1461ea` (this PR's head, Lean 4.33.1, PolyFun #154 pinned), merged with the unmodified heads of #478 and #507. The GPV files I had migrated were reverted to their pre-migration state first, so nothing of mine was masking a failure.

## Result

Both type-position blockers from #541 are resolved:

- **`GameIdentification.lean:179`** — the handler lambda passed to `appendInputLog`. Builds clean.
- **`BodyBounds.lean:391`** — the `probEvent` predicate lambda. Gone; the only failure left in that file was at line 710 and is unrelated (below).

The clearest single piece of evidence is `FlagHandlers.lean:577`. Before your fix the hypothesis there was

```
hx : x ∈ support ((fun p ↦ (p.1, p.2, false)) <$> (liftM (query n) >>= pure ∘ …))
```

carrying an unreduced `Range (Sum.inl (Sum.inl n))`, which is why `rw` reported "did not find the pattern" against a goal that visibly displayed it. After your fix it is simply

```
hx : x ∈ support (liftM (OracleSpec.query n))
```

Clean type, and `simp` now normalises the bind away entirely, so the destructuring that used to be needed is just redundant.

## What still needs doing, and what it is not

Four small edits were required on top of your fix, none of them transparency-related — all are ordinary 4.33 tactic-strength drift:

| site | fix |
| --- | --- |
| `FlagHandlers:338` | `rfl` now reports "no goals to be solved" |
| `FlagHandlers:577` | destructuring no longer needed; `simp` normalises the bind away |
| `AppendQuery:228` | trailing `rfl` after `erw` now redundant |
| `BodyBounds:710` | same shape |

With those, `VCVio` builds through both `GameIdentification` and `BodyBounds`. The next failure is `Security/NMAReduction.lean:370` — a `rw` pattern miss deeper in #507's own tree, which is that PR's migration work rather than anything to do with #541.

## Correction to my own writeup

In #541 I said `Sum.elim` being semireducible was "not the cause", on the basis that `attribute [local implicit_reducible] Sum.elim` changed nothing. That was too strong: your root-cause analysis has `Sum.elim` squarely implicated, and my experiment was the wrong intervention rather than a refutation. Loosening transparency on one link does not help when the path has to cross both the `HAdd` projections and the wrapper; moving the normal form to `Sum.rec` does.

The practical consequence is that most of the 19 patches I had banked were treating symptoms. Only the four above survive your fix, so I would rather those land as part of whichever PR owns them than carry my branches.

PR #556, opened 2026-08-27
feat(scripts): count SPMF in the boundary guard, and report the trend

dtumad:
## What and why

`scripts/check-pmf-boundary.sh` counted bare `PMF` only, and said so in its own header: *"`SPMF` and `FinRatPMF` do not count."* But `SPMF := OptionT PMF` (`ToMathlib/ProbabilityTheory/SPMF.lean:58`), so every `SPMF` use is a transitive PMF dependency. The guard was measuring **935 occurrences across 78 files** while real exposure was **2146 across 120** — it saw 44% of the surface it exists to bound.

**42 files carried coupling the guard could not see at all**, including:

| File | Was | Now |
|---|---:|---:|
| `VCVio/EvalDist/Monad/Basic.lean` | 0 | 87 |
| `VCVio/EvalDist/Instances/OptionT.lean` | 1 | 39 |
| `VCVio/OracleComp/QueryTracking/Tracing.lean` | 0 | 34 |
| `VCVio/CryptoFoundations/KEMDEM.lean` | 0 | 29 |
| `VCVio/OracleComp/Coinductive/Responder.lean` | 0 | 21 |

Upstream is retiring `PMF` construction by construction — the pinned Mathlib already deprecates `PMF.bernoulli` and `PMF.binomial` for measure-valued replacements in `ProbabilityTheory`. The guard should make that retirement visible and monotone, not merely freeze it.

## Three modes

- **ceiling** (default, blocking) — unchanged semantics, now over both tokens. A file absent from the baseline still has an allowance of zero, so new coupling cannot appear without an explicit baseline change.
- **report** (pull requests, advisory) — aggregate and the 12 largest per-file deltas against the base ref.
- **ratchet** (opt-in, *not* in CI) — every touched file still carrying coupling must decrease, unless `scripts/pmf_boundary_holds.tsv` records a reason.

## Why the ratchet is not the CI gate

Decrease-on-touch was built first and tested against the open stack **before** being wired in. It would have failed every ready pull request:

| PR | Would stall on |
|---|---|
| #542 | `OracleComp/OracleSpec.lean` — touched only because it carries one incidental `PMF` mention |
| #540 / #544 | `RenyiDiscrete.lean` 26→55, because 27 occurrences move *out* of `RenyiDivergence.lean` and 29 *into* it while a `sorry` is closed |
| #547 | `FujisakiOkamoto/{T,U}Transform.lean` — unchanged counts on files touched to delete unsound statements |

A gate that turns unrelated work into a boundary failure gets neutralised by holds, and the holds then rot. The ceiling stays the hard invariant; the report keeps the trend visible on every PR; the ratchet remains for a deliberate reduction pass.

## Incidental fixes

- **The comparison only ever ran on CI.** It used bash associative arrays (`declare -A`), and the macOS system bash is 3.2, which has none — so a maintainer could not run this locally. Moved into `awk`, which has them natively.
- **`scripts/test-pmf-boundary.sh`** fixtures all three modes in a throwaway git repository, following the existing `test-polyfun-boundary.sh` convention. It asserts explicitly that `FinRatPMF` is not counted and that the `PMF` inside `SPMF` is not counted twice — the two ways the token boundary could silently go wrong.

## Sequencing

This rewrites `scripts/pmf_boundary_baseline.tsv` wholesale, and #540/#544/#548 all touch that file. **Land this after #540 and #544**, then regenerate with `--update-baseline`. Landing it first would force a conflict resolution plus a baseline re-run on those PRs.

## Validation

- `scripts/test-pmf-boundary.sh` — all three modes, both token-boundary hazards
- `scripts/check-pmf-boundary.sh` — clean on `main` content at the regenerated baseline
- `scripts/check-agent-docs.py` — passes
- Ceiling verified to still reject a new file introducing coupling, and to accept `#542`/`#547`-shaped changes

Baseline goes 78 → 120 files, 935 → 2146 occurrences. That is a change in what is measured, not a regression.


quangvdao:
Review findings and resolution

I reviewed the boundary checker after the `main` merge and found four issues:

1. `--report` read the base branch's ceiling TSV rather than counting the base tree, so a stale ceiling could be reported as an actual reduction.
2. `--ratchet` also compared against the ceiling rather than the actual base-tree count, allowing a touched file to pass without decreasing.
3. The raw grep counted comments and docstrings, so prose consumed the migration budget and could make a semantics-preserving refactor fail.
4. With ripgrep installed, a completely clean tree exited 1 because zero matches propagated through `pipefail`.

Fixed in `ca11c16d`:

- Both report and ratchet now materialize the comparison ref and scan its actual Lean sources with the same counter used for the current tree.
- Added a regression fixture with actual base count 1 and stale ceiling 2. It now reports no movement and the ratchet correctly rejects a touch without a decrease.
- Replaced the split grep/fallback behavior with one portable Lean-aware lexical pass that ignores line comments, nested block/doc comments, and string literals.
- Added a zero-occurrence fixture.
- Regenerated the baseline against the exact post-merge head. The corrected metric records 1,821 explicit source identifiers across 106 files; the earlier 2,146/120 figure included comments and strings.
- Updated the policy docs to state explicitly that this is a syntactic migration proxy, not semantic dependency analysis.

Validation passed:

- PMF/SPMF fixtures, ceiling, actual-base report, and ratchet
- portable `/usr/bin:/bin` fixture run
- PolyFun boundary fixtures/check
- interop isolation
- agent docs
- shell syntax and `git diff --check`

I also restacked #557 onto this repaired head. It is now exactly two commits above #556, changes only `VCVio/OracleComp/Coinductive/Responder.lean`, and reports no PMF/SPMF source-count movement. I did not wait for CI.

PR #557, opened 2026-08-27
refactor(Responder): derive the stateful-responder bridge from fold naturality

dtumad:
## What

`ProbResponder.run_simulateQ_toQueryImpl_ofStateQueryImpl` was a bespoke 17-line `OracleComp` induction. Its own docstring said why:

> It is proved here by a bespoke `OracleComp` induction **for want of a generic *naturality of the universal fold along a monad morphism* law**: `simulateQ'` is already a bundled `OracleComp spec →ᵐ r`, but there is no `φ ∘ₘ simulateQ' impl = simulateQ' (φ ∘ impl)`, nor a `StateT`-functoriality lemma for `→ᵐ`.

That law now exists upstream. PolyFun's `FreeM.run_liftM_mapHom` is precisely it, transported through `StateT`, and every piece it needs is already present here:

| Needed | Supplied by |
|---|---|
| `simulateQ` is the universal fold | `simulateQ_def` (`rfl`) |
| `𝒟` as a bundled `ProbComp →ᵐ SPMF` | `MonadHom.ofLift ProbComp SPMF` (already used in `EvalDist/Instances/`) |
| post-composition of a stateful handler by a morphism | `StateT.mapHom` |
| `ofStateQueryImpl` post-composes each query with `𝒟` pointwise in the state | its own definition |

So the whole theorem is one application. The statement is unchanged; the docstring is rewritten to describe it intrinsically rather than to record the absence of the law.

## Two claims in the old docstring that did not survive checking

It named two siblings as "each re-derived by hand" for the same reason. I tried both:

- **`evalDist_simulateQ_run_congr`** *is* an instance of the law — but it is already a one-line `simp_all` induction, and routing it through the generic law takes ~8 lines of explicit rewriting, because `𝒟[…]` does not syntactically match `(MonadHom.ofLift …).toFun α x` and `rw` cannot find it. That is a strictly worse proof, so it is left alone. Reverted rather than shipped.
- **`evalDist_simulateQ_run'_eq_evalDist`** is **not** an instance at all. It is stated over `run'`, which discards the state and so is not a monad morphism on `StateT`, and its content is a uniformity hypothesis on the implementation (`𝒟[(so t).run' s] = OptionT.lift (PMF.uniformOfFintype …)`) rather than naturality.

So the honest score is one collapse, not three. The docstring no longer claims otherwise.

## Note on `by exact`

The proof is `by exact …` rather than a term-mode `:=`. Term mode fails with a type mismatch: matching the generic law needs the unfoldings of `simulateQ`, `toQueryImpl`, `ofStateQueryImpl`, `StateT.mapHom`, and `𝒟`, which are definitional but not syntactic. Recorded in a comment at the site.

## Validation

Full `lake build`: 0 errors. Both downstream consumers (`Responder.lean`'s `run_map_…` and `CryptoFoundations/AsymmEncAlg/INDCPA/Oracle.lean`) build unchanged.

### DIFF
--- a/VCVio/OracleComp/Coinductive/Responder.lean
+++ b/VCVio/OracleComp/Coinductive/Responder.lean
@@ -193,8 +193,8 @@ game between its `ProbComp` presentation (where probability reasoning happens) a
 
 Structurally it is the naturality of `simulateQ` along the monad morphism
 `𝒟 : ProbComp →ᵐ SPMF`, transported through `StateT σ` — i.e. `𝒟 ∘ simulateQ impl =
-simulateQ (𝒟 ∘ impl)`. That is exactly `PFunctor.FreeM.run_liftM_mapHom` at
-`φ := MonadHom.ofLift ProbComp SPMF`: `simulateQ` is the universal fold
+simulateQ (𝒟 ∘ impl)`. That is exactly `PFunctor.FreeM.run_liftM_mapHom` at the bundled
+evaluation-distribution morphism: `simulateQ` is the universal fold
 (`simulateQ_def`), `ofStateQueryImpl` post-composes each query with `𝒟` pointwise in the
 state, and `StateT.mapHom` is that post-composition as a morphism, so the whole statement
 is one instance of the generic law rather than an induction over `OracleComp`. -/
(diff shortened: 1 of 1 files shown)


quangvdao:
Review/update summary

I reviewed the unique #557 slice against its parent #556 and current `main`.

Findings:
- The Lean theorem statement is unchanged, and `PFunctor.FreeM.run_liftM_mapHom` is the correct stateful naturality instance here. I found no mathematical correctness issue in the refactor.
- The failing **Check Library File Imports** job was actually #556's new PMF/SPMF ceiling check: this slice raised `Responder.lean` from 21 to 23 lexical occurrences.
- The PR was based on `main` even though it included #556, so the review surface also exposed the parent infrastructure diff. The stack is now explicit.
- One parent-owned follow-up remains in #556: report mode compares the current count with the base ceiling TSV rather than recomputing the base tree's actual count. If a file shrinks below a stale ceiling, the reported trend can be inverted. That should be repaired in #556.

What changed:
- Merged current `main` (`2bdbf9ba`) into #556 and pushed merge commit `c8403e40`.
- Restacked #557 linearly on the updated #556 head and changed this PR's base to `dtumad/pmf-boundary-spmf-ratchet`.
- Preserved the fold-naturality refactor (`eb5b56f4`) and added `77891e2f`, which lets Lean infer the target monad and phrases the docstring intrinsically, restoring `Responder.lean` to the 21-occurrence ceiling.

Local validation passed:
- PMF/SPMF fixtures and ceiling
- PolyFun boundary fixtures and check
- interop isolation
- agent docs
- `git diff --check`

I did not wait for CI. The previous #557 head had already passed Build Project, downstream consumer, lint, docs, and isolation; only the lexical boundary gate failed.

PR #572, opened 2026-08-29
feat(uc): add reassociation-stable proportional scheduler

quangvdao:
Binary scheduling that chooses each subtree with probability one half changes the distribution when parallel components are reassociated. This PR samples a uniform slot from the combined positive scheduler masses and proves the swap, flat-choice, and reassociation laws required by PolyFun's scheduled sampler semantics.

The integration pins merged PolyFun `988a1ab00bf3fe8da73c648757033548586f45d4`, including constructor normalization and finite-prefix resource accounting, with its published cslib prerequisite. It adapts the scheduler to positive naturals and updates the affected replay, probability, one-time-pad, and optional-backend proofs.

Validation: the complete local build, environment and source lint, tests, boundary checks, and axiom sweep pass, including the intervening merged main changes. The optional complexity package also passes its exact-machine canaries, guarded trust checks, and compatibility preflight. No new axiom or sorry debt. The scheduler delta is unchanged by the stack rebase.

Stacked on #696 to share the already reviewed measure foundations; the PR delta contains only scheduler and dependency-integration changes.


dtumad:
Deep semantic audit across the open PR set (semantics only). The coherence facts proven here are real (positive masses intrinsic, and the falsifier on the PolyFun side shows `IsFlat` is substantive), but the PR is not close to mergeable in its current shape, for structural reasons:

1. **BLOCKING: the PolyFun pin predates the review rework of its own dependencies.** This file uses `ScheduleMass` nine times, but PolyFun#174's head is now `52d85bd` ("refactor(uc): use PNat and expose scheduled APIs"), which **deleted `ScheduleMass` in favor of Mathlib's `ℕ+`** — and #176's public signatures moved with it. This branch cannot build against the reviewed heads of the PRs it depends on; `MERGEABLE` here only means no git conflict. Migrate to `ℕ+` and re-pin.
2. **The layer split is right, but the PolyFun side it depends on forks PolyFun `main`.** `PolyFun/Interaction/UC/OpenProcessSamplerFactorization.lean` on PolyFun main already has `sourceDraw`/`leftDraw`/`rightDraw`, their `_bind` laws, `parLeftPathEquiv`, and `schedulerFlip`; PolyFun#174/#176 re-introduce the same nine declarations in a `BinaryScheduler` namespace differing only in mass-indexing. The expected move is to generalize the existing `OpenProcessFactorization` API to a mass-indexed sampler in place, not to grow a divergent twin — I'll take that up on the PolyFun side, but this PR should not land against the fork.
3. **BLOCKING (payoff gap): "reassociation-stable" is currently proven only at the scheduler-draw level, and nothing downstream can consume it.** PolyFun main's process-level results (`openTheory_plug_par_left_sampler_equiv` and friends) each take exactly the hypothesis shape `R.rel (sourceDraw σ) (leftDraw σ)` that `ProportionalScheduler.isCoherent` provides — but no analogue of those theorems exists for `scheduledOpenTheory`, which also has no `Semantics` instance, no `Emulates`, no `AsyncSecurity`, and hence no consumer anywhere in VCVio. The doc line this PR adds to `docs/agents/interaction.md` ("proves that the resulting output distribution is unchanged by swapping or reassociating component frontiers") overstates accordingly: what's proven is that the *scheduler draw* is unchanged for one three-leaf node, not any process- or boundary-level observation. Minimum bar: one `scheduledOpenTheory_plug_par_left_sampler_equiv` so the coherence hypothesis discharges a process-level statement — or, cheaper and immediately valuable: `outputRel` instantiates the *existing* `openTheory_plug_*_sampler_equiv` theorems for a symmetric fair coin today (`hfair` is trivial there); proving that corollary would give the first real UC-level payoff on main without waiting for the scheduled theory.
4. `outputRel` re-derives `𝒮[·]`-extensionality by hand (`tsum_congr` over `probOutput_bind_eq_tsum`) — build it from `evalSPMF` equality and the existing congruence lemmas instead. And the generic counting lemmas (`card_filter_fin_lt`, `card_filter_fin_Ico`, `sum_ulift_bool`, `nested_ratio`) belong upstream (PolyFun or `ToMathlib`), not in a UC file.

Issue #550, opened 2026-08-27
`MultiTarget` SM-PRE/SM-TCR do not model the notions they name, and `slhdsa_euf_cma_security` omits assumptions the recovered SPHINCS+ proof requires


mmtobs:
Hi, first time opening an issue here, so apologies if I've missed a convention somewhere (please tell me if so).

I've been exploring the hash-based parts of this repository for a related project and ran into a few things that look like they might be mistakes. One disclaimer up front: I know the hash-based literature and its proofs reasonably well (I worked on the EasyCrypt formal verification of XMSS and SPHINCS+), but I'm new to Lean and to parts of its foundations, so please read the following with that in mind. I'd rather raise these and be corrected than leave them sitting. Glad to clarify anything.

All of this is from reading [`MultiTarget.lean`](https://github.com/Verified-zkEVM/VCVio/blob/3ecd5523d5beaa10b026f663174e8b9d9708c24b/VCVio/CryptoFoundations/HardnessAssumptions/MultiTarget.lean) and [`HashSig/SLHDSA/Security.lean`](https://github.com/Verified-zkEVM/VCVio/blob/3ecd5523d5beaa10b026f663174e8b9d9708c24b/HashSig/SLHDSA/Security.lean) at `3ecd5523`. Items 1–5 are about the *statements*, so they're independent of the `sorry`.

**1. `MultiTarget.TcrProblem` is not SM-TCR** ([`MultiTarget.lean:75-101`](https://github.com/Verified-zkEVM/VCVio/blob/3ecd5523d5beaa10b026f663174e8b9d9708c24b/VCVio/CryptoFoundations/HardnessAssumptions/MultiTarget.lean#L75-L101)). No public parameter is sampled: `f` is a field of the problem, so `choose` can evaluate it freely while picking targets. There's also no distinct-tweak restriction. This essentially collapses the problem to regular collision resistance. The actual problem should sample the public seed, withhold it from the adversary during target selection, and enforce distinct tweaks. (Cf. [HK22](https://eprint.iacr.org/2022/346) Def. 2 and [BDHMS24](https://eprint.iacr.org/2024/910) Fig. 5.)

[...]

PR #580, opened 2026-08-29
Foundations for tweakable-hash assumptions, replacing old multi-target notions

mmtobs:
# Foundations for tweakable-hash assumptions, replacing old multi-target notions

Closes items 1–3 of #550.

Open for feedback on the questions below, as well as whether the general approach makes sense (again, not too well-versed in Lean, VCVio, or the used foundations.) 

## The defect

`HardnessAssumptions/MultiTarget.lean` states SM-TCR and SM-PRE over a **fixed function**: `f` is a
field of the problem, with no sampled public parameter, no oracle phase, and no distinct-tweak
restriction.

## What it adds

- `TweakableHashCollection` (+ `.empty`, `.cons`). **No `seedGen` field** — one seed per game, from
  the attacked member. `cons th thColl` places the attacked member in its own collection at index
  `none`, so it is evaluable at tweaks that are not targets; membership is definitional
  (`cons_eval_none` is `rfl`), not a hypothesis. Indexing is by an explicit `ι` rather than by
  message length as in HK22 and BDHMS, which cannot name two members of equal length.
- One shared collection oracle. The collection is a *parameter*, not a parallel `-C` game family;
  the stand-alone notion is the instance at `ι := Empty`, and that this recovers rather than
  approximates it is proved (`isEmpty_domain_collectionSpec_empty`).
- `TweakReserved` / `TweakFresh`, the conditions the collection and challenge oracles respectively
  reject on. The challenge history is `List Q` read through `tweakOf : Q → Tweak`, so each game
  records what its winning condition needs and the collection oracle is indifferent to the choice.
- SM-DT-TCR and SM-DT-PRE: seed sampled and withheld, two-phase adversary, distinct tweaks enforced
  across both oracles. The two phases are two fields at two types, so "the oracle is gone once `P`
  is revealed" is a typing fact, not a convention.
- **11 branch pins** fixing every oracle branch and the append-at-the-end history order.
  Mutation-tested: reversing the order, adding a self-distinctness check, and dropping either half
  of the disjointness check each fail exactly the pins they should.

## What it changes

- Replaces the old games outright; nothing referenced them.
- `namespace MultiTarget` → `TweakableHash`, one file per game, following `AsymmEncAlg/`.
  **Separable** — isolated in one mechanical commit if you would rather it were its own PR.
- `emb : M' → M` must be injective, else the oracle's uniform draw on `M'` is not uniform on its
  image.
- Fixes two citations: ePrint 2024/910 had the wrong authors, `Pre.lean` cited the TCR figure.

## Out of scope

SM-UD, SM-rTCR, SM-DSPR, SM-DT-OpenPRE; SLH-DSA instantiation; reductions and the `p`-fold loss.
Item 6 of #550 is a follow-up.

## Open questions

1. These oracles reject a distinctness-violating query on arrival; HK22's `DIST` and BDHMS's `VQS_t` check it in the winning condition. The former seems cleaner in general (and was also partially adopted by more recent work in, e.g., DKKW25) .The docstring argues why this is fine, but this is not a proof. Okay, or prefer the definitions directly from the SPHINCS+ papers?
2. `standalone` and `not_tweakReserved_iff` have no consumer yet — drop (under "no speculative
   features") or fine to keep?
3. Attribution in the files okay?

## Verification

Full build, `VCVioTest`, smoke, `axiomsweep --check` (baseline unchanged), `test-axiomsweep`,
`mk_all`, `lint-style`, pmf/polyfun boundaries, extern/interop isolation, agent-docs, both SLH-DSA
KATs — green locally, zero warnings in the touched files.

## After this

If this shape is agreed, the remaining notions follow straightforwardly: same skeleton — `Problem` /
`Adversary` / `challengeOracle` / `oracles` / `Experiment` / `Advantage` plus branch pins — and the
collection oracle as-is, `tweakOf` already covering a tweaks-only history (SM-UD) and a
`(tweak, ρ, message)` record (SM-rTCR). SM-UD also reuses `M'` / `emb` / `emb_injective` verbatim,
DKKW25 Def. 5 carrying the same subspace parameter. Naming extends without renames: `SMDTUD.lean`,
`SMDTRTCR.lean`, `SMDTDSPR.lean`, `SMDTOPENPRE.lean`.


mmtobs:
@quangvdao Here is my attempt at a PR for the first couple of points in issue #550. Mostly contains the foundations for everything, not yet full integration. However, as also mentioned in the PR message, I tried to develop it with the remainder of the properties and proofs/applications in mind, such that adding the rest should be straightforward.

quangvdao:
Superseded by #573, which has now merged and preserves the complete foundational contribution and attribution from this PR.

PR #582, opened 2026-08-29
feat(Ring): add generic SampleableType instances for ring carriers

Skanislav:
## Summary

`$ᵗ T` needs a `SampleableType T` instance, and VCVio provides none for the scheme ring carriers: ML-DSA takes `[SampleableType (RqVec p.l)]` as a hypothesis on every statement, Falcon hand-rolls one via `inferInstanceAs`, and ML-KEM has nothing. A downstream MLWE game instantiated at concrete ML-KEM-768 therefore fails to elaborate:

```lean
noncomputable def keyHopProblem : … where
  sampleUniform := $ᵗ (TqVec params.k)   -- failed to synthesize SampleableType (Vector MLKEM.Tq _)
```

This PR adds the instance once, generically, so it covers ML-KEM, ML-DSA and Falcon and lets the scheme-level hypotheses go away.

## Why instance search can't find it on its own

`MLKEM.Rq` unfolds through three abbrevs to `PolyBackend.Poly (NegacyclicRing.backend (vectorNegacyclicRing Coeff 256))`, where `vectorNegacyclicRing` is a plain `def` (`@[implicit_reducible]`, which instance search does not unfold). Search stops there and never sees `Vector Coeff 256`, so the existing `SampleableType (Vector α n)` instance is invisible. `TransformPoly` is a structure, so nothing transports through it either. The same opacity is why the scheme files already need `change … ; infer_instance` for `DecidableEq`.

## Changes

**Generic layer**
- `Ring/Core.lean`: `PolyBackend.equivPi : backend.Poly ≃ (Fin degree → Coeff)`, packaged from the existing `coeff_build` / `build_coeff` laws, plus `Fintype backend.Poly` given `[Fintype Coeff]`.
- `Ring/Transform.lean`: `TransformPoly.equivPoly : TransformPoly ring ≃ ring.Poly` and `Fintype (TransformPoly ring)`.
- `Ring/Sampling.lean` (new): `SampleableType backend.Poly` and `SampleableType (TransformPoly ring)` given `[SampleableType Coeff]`, via `SampleableType.ofEquiv` over those equivalences. Stated on the semantic carrier `backend.Poly`, so the discrimination-tree key is `PolyBackend.Poly _ *` and it unifies with the stuck term above without unfolding the ring — the same mechanism the existing `Zero ring.Poly` / `AddCommGroup ring.Poly` instances in `Kernel.lean` rely on. Uniformity is inherited from the coefficient sampler through a bijection, so no new proof obligations. Kept out of `Kernel.lean` so the rest of `Ring/` stays free of VCVio's oracle-computation imports.

**Schemes**
- ML-KEM: add `NeZero modulus` (the only missing scheme-specific piece — `SampleableType (ZMod n)` needs it). `Rq`, `Tq`, `RqVec`, `TqVec`, `TqMatrix` are now all sampleable. Drop the now-redundant `local instance` in `Concrete/Encoding.lean`.
- Falcon: remove the hand-rolled `Fintype (Rq n)` / `SampleableType (Rq n)` instances so there is a single instance per type (two non-defeq samplers for the same type would break `rfl`-level reasoning downstream).
- ML-DSA / Falcon: drop `[SampleableType (RqVec _)]`, `[SampleableType (TqMatrix _ _)]`, `[SampleableType (Rq p.n)]` hypotheses and the matching `omit` lists across `Scheme`, `Security`, `SecurityNMA`, `SecurityHVZK`.
- `Ring/Smoke.lean`: `inferInstance` canaries for all three schemes, so a future change to the ring layer that breaks resolution fails at build time.
- `docs/agents/lattice.md`: entry for the new file.

## Verification

- `lake build` of all non-test roots: clean; CI warning budget (`check-warning-log.py`, zero non-`sorry` warnings): passes.
- `lake exe axiomsweep --check`: no new taint (40 sorry / 20 native, unchanged).
- `lake exe lint-style LatticeCrypto`, `mk_all --lib LatticeCrypto --check`, PMF-boundary, Extern/Interop isolation, agent-docs checks: all pass.
- Scratch check at concrete ML-KEM with no local instances: `do let a ← $ᵗ (MLKEM.TqVec k); let r ← $ᵗ MLKEM.Rq; …` elaborates; `Pr[= f | $ᵗ MLKEM.Rq] = (Fintype.card MLKEM.Rq)⁻¹` holds; `Fintype.card MLKEM.Rq = modulus ^ ringDegree` is provable via `equivPi`.

## Downstream note

Any tree carrying its own `instance : SampleableType MLKEM.Rq := inferInstanceAs (… Vector …)` should delete it on bump — it would coexist with this one as a second, non-definitionally-equal sampler.


quangvdao (approved):
Reviewed the generic ring-carrier equivalences, transported uniform samplers, instance-resolution canaries, scheme-level assumption cleanup, and exact-head validation. The update from main preserves the reviewed patch exactly. Looks good to merge.

PR #583, opened 2026-08-30
feat(MLKEM): expose concrete encoding facts — DecidableEq on encoded types, public encoder sizes

Skanislav:
## Summary

`$ᵗ`-style elaboration of the concrete ML-KEM scheme fails on a missing `DecidableEq`, and the byte length of every concrete encoder is a `private` theorem. Both facts are true by construction of `concreteEncoding` and invisible outside `LatticeCrypto/MLKEM/Concrete/Encoding.lean`:

```lean
example := MLKEM.asKEMScheme concreteNTTRingOps mlkem768Encoding mlkem768Primitives
-- failed to synthesize DecidableEq mlkem768Encoding.EncodedTHat
```

Every generic KEM entry point — `encapsulationKeyCheck`, `encaps`, `decaps`, `asKEMScheme` (`KEM.lean:26-94`) — requires `[DecidableEq encoding.Encoded{THat,U,V}]`, and `KPKE.lean:44` derives `DecidableEq (PublicKey …)` from the first. `concreteEncoding` sets all three carriers to `ByteArray`, but instance search can't see it. VCVio's own test suite carries the workaround: `LatticeCryptoTest/MLKEM/Helpers.lean:23-36` declares six `instance : DecidableEq (mlkemNNNEncoding).EncodedU/V := inferInstanceAs (DecidableEq ByteArray)`, one per parameter set and component.

Separately, `byteEncode_size : (byteEncode d f).size = 32 * d` (`Concrete/Encoding.lean:262`) is `private`. It is the fact behind `Params.ciphertextBytes = 32 * (du * k + dv)` and `publicKeyBytes = 384 * k + 32`, but nothing outside the file can connect an encoder's output to those constants, so a downstream wire-format argument has to take the `v`-component length from FIPS 203 instead of proving it. #520 exposed element laws for the encodings but not the size law.

This PR adds the instances once, generically, and publishes the size facts as named laws on `concreteEncoding`, so a downstream file can instantiate the concrete KEM and reason about its wire format with no local instances or assumptions.

## Why instance search can't find it on its own

`concreteEncoding` is an `@[expose] def`: the body is available to the kernel and to `simp`/`rfl`, but instance resolution unfolds only reducible constants, so `(concreteEncoding params).EncodedU` never becomes `ByteArray` during search. Making the whole 20-field structure literal an `abbrev` would unfold it at every reducible-transparency check; the cheaper fix is to state the instances on the projection itself, so the discrimination-tree key is `Encoding.EncodedU (concreteEncoding _)` and unification succeeds without unfolding anything — the same mechanism as the ring-carrier instances in #582.

There is a second opaque layer above it: `mlkem512Encoding` / `mlkem768Encoding` / `mlkem1024Encoding` (`Extern/MLKEM/Instance.lean:123-129`) are plain `def`s of `concreteEncoding _`, so instances keyed on `concreteEncoding _` would not reach `mlkem768Encoding.EncodedU`. That is why the test helper needed one instance per parameter set. The three aliases are one-line wrappers with no proof content; they become `abbrev`s and the generic instances reach them.

## Changes

**`LatticeCrypto/MLKEM/Concrete/Encoding.lean`**
- `DecidableEq (concreteEncoding params).EncodedTHat` / `.EncodedU` / `.EncodedV`, each `inferInstanceAs (DecidableEq ByteArray)`, directly after `concreteEncoding`.
- `byteEncode_size` is no longer `private`. Public size laws stated on the concrete encoding so consumers never touch the internal encoders: `byteEncodeVec_size` (`32 * d * k`), `concreteEncoding_byteEncode12Vec_size` (`384 * k`), `concreteEncoding_byteEncodeDUVec_size` (`32 * du * k`), `concreteEncoding_byteEncodeDV_size` (`32 * dv`), plus the two identities tying them to `Params.ciphertextBytes` and `Params.publicKeyBytes`. Proofs are `Array.size_ofFn` and `byteEncode_size`; no new reasoning.

**`Extern/MLKEM/Instance.lean`**
- `mlkem512Encoding`, `mlkem768Encoding`, `mlkem1024Encoding`: `def` → `abbrev`. Bodies unchanged.

**`LatticeCryptoTest/MLKEM/Helpers.lean`**
- The six hand-rolled `DecidableEq` instances are deleted; the generic ones resolve. A `#synth DecidableEq mlkem768Encoding.EncodedTHat` canary guards the resolution path so a future `abbrev` → `def` regression fails at build time.

**`docs/agents/lattice.md`**
- Concrete encoded types carry `DecidableEq`; encoder sizes are named laws.

## Verification

- `lake build LatticeCrypto Extern LatticeCryptoTest`: clean; CI warning budget (`check-warning-log.py`, zero non-`sorry` warnings) passes.
- `./scripts/check-extern-isolation.sh`, `./scripts/check-interop-isolation.sh`: pass (instances live in `LatticeCrypto/`, the `abbrev` change in `Extern/`; no new cross-layer import).
- `lake exe lint-style LatticeCrypto`, `mk_all --lib LatticeCrypto --check`, `lake exe axiomsweep --check`: pass, no new taint.
- Scratch check with no local instances:
  ```lean
  #synth DecidableEq mlkem768Encoding.EncodedTHat
  noncomputable example := MLKEM.asKEMScheme concreteNTTRingOps mlkem768Encoding mlkem768Primitives
  example (v : Rq) : (mlkem768Encoding.byteEncodeDV v : ByteArray).size = 128 := by
    simpa using concreteEncoding_byteEncodeDV_size mlkem768 v
  ```

## Downstream note                                                                                                                                                                                                                           
Trees carrying `instance : DecidableEq mlkemNNNEncoding.EncodedX := inferInstanceAs (DecidableEq ByteArray)` can delete them; the duplicates are definitionally equal to the new ones, so keeping them is harmless but redundant. `simp only [mlkem768Encoding, …]` still works — `abbrev` keeps the equation lemma.


quangvdao (approved):
Thanks for the work here. We took the PR over to make the last few changes we wanted before merging it to main: the full Lean-only LatticeCryptoTest library now runs in per-PR CI, and the FIPS 203 ByteEncode/ByteDecode references are corrected to Algorithms 5 and 6. The complete LatticeCryptoTest target passes, including the DecidableEq alias checks and the end-to-end asKEMScheme canary. Approved.

PR #622, opened 2026-09-02
test(CryptoFoundations): cover the THF games end to end

mmtobs:
Follow-up to #573, which landed the games, and #596, which extended the family. Test-only plus one
CI file; no `VCVio/` source is touched.

## Why

Paths below are given in full where a bare filename would be ambiguous: after commit 3 both
`HardnessAssumptions/TweakableHash/` and `VCVioTest/` contain an `SMDTTCR.lean` and an
`SMDTPRE.lean`.

**`SM_DT_PRE_Experiment` has no end-to-end test anywhere in the tree.** It is exercised only by the
four `SM_DT_PRE_challengeOracle_run_of_*` pins inside
`HardnessAssumptions/TweakableHash/SMDTPRE.lean`.

**Those pins probe a different mutation surface than a game-level canary.** A pin states what the
*oracle* returns; a canary states what the *game* then decides. `…_run_of_full` says the oracle
answers `none` past the cap; it does not say the game loses at the unrecorded index. Two mutations
confirm the gap is real — each was applied to the source and both test modules re-elaborated:

| Mutation | Site in `HardnessAssumptions/TweakableHash/` | Oracle pins broken | Canaries broken |
|---|---|---|---|
| make the out-of-range index lookup total | `SMDTTCR.lean:127` | **none — module still elaborates** | `oracle_separation_canary`, `lose_overCap`, `lose_reusedTweak`, `lose_outOfRange` |
| make the phase-1 coin deterministic | `SMDTTCR.lean:86` | **none — module still elaborates** | `run_coinFlip` |

Five further mutations (append order, cap, both halves of `TweakFresh`, SM-PRE target recording)
break pins and canaries together. The split is principled: oracle mutations surface at the
transcript equations, winning-condition mutations at the verdict theorems, and only the second class
is invisible to the pins.

**`VCVioTest/MultiTargetCollection.lean` builds via `VCVioTest.lean` but is style-linted nowhere.**
It is the only file in this lane at none of the three `linting.yml` registration points. #601
registered `VCVioTest/CryptoFoundations/SignatureAlg.lean` at all three in the PR that introduced
it; this restores the same treatment.

**It is also the only test module in the lane not named after its source module.** `SMDTDSPR`,
`SMDTOpenPRE`, `SMDTUDC`, `SMDTTCRFinalValidity` and `SMDTPREFinalValidity` all mirror theirs
exactly; `MultiTargetCollection` is named after `HardnessAssumptions/MultiTarget.lean`, which #573
[...]

PR #623, opened 2026-09-02
feat(CryptoFoundations): parameterize SM-DT-UD by a message subspace

mmtobs:
Follow-up to #596, which added the game.

## Why

`SM_DT_UD_C_SourceFinalValidity.Problem` draws the hidden real-world input from
`inputGen : ProbComp M` — an opaque distribution on the **full** message space. "Uniform on a
subspace of size `|M′|`" is not recoverable from that, so three quantities in the literature cannot
be attached to the game at all: **DKKW25 Corollary 1's term `Adv^{SM-UD}_{Th,H,1}`**, instantiated
under `H ⊆ M` at a *proper* subspace, and **Table 1's `q/|M′|`** and **Supplementary Material B
Theorem 4's `12q/√|M′|`**, the classical and quantum bounds on the notion.

This is consistency, not a new convention. `SM_DT_PRE_Problem` (`SMDTPRE.lean:69`) and
`SM_DT_PRE_SourceFinalValidity.Problem` (`SMDTPREFinalValidity.lean:42`) already carry these fields
and already sample from `M'`. PRE and UD are the two games in the directory whose challenge oracle
*draws* its own message; they should agree on how the drawing space is described.

## What

`Problem` gains the three fields the PRE problems already carry, in their order, with `inputGen`
retyped to the subspace:

```lean
structure Problem (ι PkSeed Tweak M M' Y : Type) where
  th : TweakableHash PkSeed Tweak M Y
  emb : M' → M
  emb_injective : Function.Injective emb
  inputGen : ProbComp M'
  outputGen : ProbComp Y
  thColl : TweakableHashCollection ι PkSeed Tweak Y
  numTargets : ℕ
```

The real challenge world becomes `x ← inputGen; return th.eval pk t (emb x)`. `emb` is applied
inside `response`, so `challengeOracle` and `challengeOracle_run` are untouched. No definition or
proof uses `emb_injective`; it is a carried side condition for downstream reductions, exactly as in
`SMDTPRE.lean`, and the field docstring gives the reason (a non-injective `emb` makes the law of
`emb` fail to identify `M′` with an image of cardinality `|M′|`; its pushforward need not be uniform on that image).

Also added, mirroring `SMDTOpenPRE.lean` but asking uniformity of the *subspace* rather than of `M`,
which supplies the input-distribution hypothesis needed for a bound in `|M′|` at a strict `M′ ⊊ M`:

```lean
def Problem.HasUniformInputs [SampleableType M'] (prob : Problem …) : Prop :=
  prob.inputGen = $ᵗ M'
```

**Conservative.** `M' := M`, `emb := id`, `emb_injective := Function.injective_id` recovers the
previous game *definitionally* (stated in the module docstring). The type-parameter change is
breaking, but the blast radius is closed: `git grep -l SM_DT_UD_C -- '*.lean'` at `b4449d1a` returns
exactly two files, and this PR migrates both. No reduction consumes UD-C.

### Edit: Commit 1 renames the module first

`SMDTUDC.lean` → `SMDTUDFinalValidity.lean`, `SM_DT_UD_C_SourceFinalValidity` →
`SM_DT_UD_SourceFinalValidity`, with the test module and namespace following.
Consistent with the naming of other properties.

Doing it in commit 1 means the subspace work lands on the module under its final name, rather than
renaming a file the reviewer has just read. It is a pure rename — 77 insertions, 77 deletions, no
definition or proof touched — so it reviews independently and can be dropped if you would rather
keep the name.

`SMDTDSPR.lean` and `SMDTOpenPRE.lean` have the same suffix problem and are handled separately in #625, since they are unrelated to subspaces.

## Source-alignment boundary

The abstract Lean game intentionally permits an arbitrary ideal `outputGen`. This PR now exposes
`Problem.HasUniformOutputs` alongside `Problem.HasUniformInputs`; the DKKW distributional
instance requires both. Concrete random-oracle bounds additionally require the corresponding oracle
model and query-budget hypotheses.

DKKW Definition 5 rejects an invalid challenge query on arrival, whereas this module uses a sticky
source-final-validity monitor. This PR repairs the message-subspace and uniform-distribution
interface only; it does not by itself prove that the two validity presentations are equivalent for
SM-DT-UD. That conversion, or a matching rejection-on-arrival SM-DT-UD game, remains an explicit
follow-up.

## Verification

Base: `main` at `b4449d1a`. All exit 0, no new warnings.

`lake build`; `lake build VCVioTest HashSigTest`; `lake exe axiomsweep --check` (baseline unchanged);
`lake exe mk_all --lib {VCVio,VCVioTest} --module --check` (no new files); `lake exe lint-style` over
the full `linting.yml` list (`VCVioTest.SMDTUDC` is already registered);
`check-{extern,interop,complexity-backend}-isolation`; `check-pmf-boundary` (`SMDTUDC.lean` is absent
from the baseline, so its ceiling is zero; it carries no `PMF`/`SPMF` identifiers either side).

All four existing canaries in `VCVioTest/SMDTUDC.lean` hold with no edit to their statements or
proofs, as do the ten `rfl`-proved lemmas behind them. But they run at `M' = M` with `emb := id`,
where `emb` is unobservable — a game that ignored `emb` entirely would pass every one of them. So
the second commit adds a one-element `Input ⊊ Bool` embedding to `true`, with pins showing the
transcript records the *embedded* element and `HasUniformInputs` discharged at the subspace. That is
the value-level check that the parameterization is live rather than decorative.


quangvdao (approved):
Approved at exact head 3212bce5 after maintainer audit, source-alignment corrections, full hosted CI on the repaired content, and a validated restack onto merged main/#622. The only restack conflict was the lint registry; it was resolved as the union of current SLH-DSA canaries, #622 TCR/PRE canaries, and this PR’s SMDTUDFinalValidity canary.

PR #624, opened 2026-09-02
feat(CryptoFoundations): relate the reject-on-arrival and final-validity games

mmtobs:
Follow-up to #573 (rejection-on-arrival) and #594/#596 (source final validity).

## Why

The tweakable-hash directory carries two presentations of the same tweak discipline:

- **rejection-on-arrival** (`Collection.lean`, `SMDTTCR.lean`, `SMDTPRE.lean`, from #573) — the
  challenge oracle answers `Option Y`, returning `none` on a cap, a duplicate target tweak, or a
  clash with the collection history, and leaves the state untouched;
- **source final validity** (`FinalValidity.lean`, from #594) — the oracles answer plain `Y`,
  always, and set a sticky `valid : Bool` that the winning condition conjoins.

Both exist by design. `docs/design/slh-dsa-fips205-generalization.md` — the SLH-DSA lane's plan,
which divides the work into named slices — names #573 as owner of "the consolidated
reject-on-arrival SM-DT-TCR and SM-DT-PRE collection games", and requires the slice that added the
monitor (there called CF1) to

> preserve the reject-on-arrival experiments merged through #573 as distinct games. Public
> names and docstrings MUST make the semantic choice explicit; CF1 MUST NOT silently change the
> winning condition of an existing name.

What that plan never assigns to anyone is a theorem relating the two, so **every reduction has
to pick a presentation, with no licence to move a bound from one to the other.** A reduction that
produces reject-on-arrival adversaries cannot consume a hardness assumption stated against the
monitor, or vice versa.

## What

`VCVio/CryptoFoundations/HardnessAssumptions/TweakableHash/ToFinalValidity.lean`, one conversion per
game, for SM-DT-TCR and SM-DT-PRE.

```lean
theorem SM_DT_TCR_experiment_toSourceFinalValidity (adv : SM_DT_TCR_Adversary prob) :
    SM_DT_TCR_SourceFinalValidity.Experiment adv.toSourceFinalValidity = SM_DT_TCR_Experiment adv

theorem SM_DT_TCR_advantage_toSourceFinalValidity (adv : SM_DT_TCR_Adversary prob) :
    SM_DT_TCR_Advantage adv = SM_DT_TCR_SourceFinalValidity.Advantage adv.toSourceFinalValidity
```

plus `advantage_le_toSourceFinalValidity` as an `le_of_eq` corollary, and `SM_DT_PRE_*` twins.

**What is proved is an equality** — of the two experiments as distributions, hence of the two
advantages. The `≤` form is kept because that is the shape a reduction applies. The equality means that this rejection-on-arrival to source-final-validity conversion has no
quantitative loss. No reverse adversary conversion is claimed here.

Four sibling docstrings asserted that no relation exists. The two game-agnostic ones
(`Collection.lean`, `FinalValidity.lean`) now point at the module and describe the conversion as a
per-game family, so they stay accurate as games are added; the two game-specific ones
(`SMDTTCRFinalValidity.lean`, `SMDTPREFinalValidity.lean`) cite their own game's theorem. The
semantic choice each documents is still explicit — now so is its consequence.

No change to any existing game's definition, oracle result types, or winning condition.

## Verification

Base: `main` at `b4449d1a`. All exit 0, no warnings.

`lake build`; `lake build VCVioTest HashSigTest`; `lake exe axiomsweep --check` (baseline unchanged;
no axioms, no `sorry`); `lake exe mk_all --lib {VCVio,VCVioTest} --module --check`;
`lake exe lint-style` over the full `linting.yml` list plus the new test module;
`check-{extern,interop,complexity-backend}-isolation`; `check-pmf-boundary`
(`ToFinalValidity.lean` is a new `VCVio/` file, so its ceiling is zero; it carries no
`PMF`/`SPMF` identifiers).

The public boundary consists of the two problem maps, the two adversary maps, and their
experiment/advantage theorems. Wrapper handlers, coupling invariants, and state-projection machinery
are private implementation details; the latter is an implementation-only import.

The earlier `VCVioTest/ToFinalValidity.lean` repeated the bridge theorem to obtain its advertised
canaries, so it supplied no independent failure signal and has been removed. End-to-end
rejection-on-arrival game behavior is covered directly by #622; the universally quantified
experiment equalities in this module are the conversion contract.


quangvdao (approved):
Approved at exact head 08084a93 after comprehensive formalization/API-boundary audit, proof-helper/test-bloat cleanup, full hosted CI, scratch downstream public-API validation, and validated pre-restack on #625. The public surface is the intended two problem maps, two adversary maps, and six experiment/advantage results; low-level handlers remain private.

quangvdao (approved):
Re-approved at exact head 5290780a after refreshing to repaired #625 head 983993a1. The dependency merge was tree-identical/conflict-free, and this exact head passed the complete hosted suite. The previously audited public/private API boundary is unchanged.

PR #625, opened 2026-09-02
refactor(CryptoFoundations): distinguish final-validity THF modules

mmtobs:
Follow-up to #596, which added the games. Renames and docstrings only.

## Why

Modules in `HardnessAssumptions/TweakableHash/` pair up along two axes, and three files sit on the
wrong side of one of them.

**The presentation axis.** `SMDTXXX.lean` is the rejection-on-arrival game; `SMDTXXXFinalValidity.lean`
is the sticky-monitor one. SM-DT-TCR and SM-DT-PRE follow this. But:

| Module | Namespaces it defines | Reads as |
|---|---|---|
| `SMDTDSPR.lean` | `SM_DT_DSPR_SourceFinalValidity` only | rejection-on-arrival |
| `SMDTOpenPRE.lean` | `SM_DT_OpenPRE_SourceFinalValidity` only | rejection-on-arrival |

Both are monitor-only modules wearing the unsuffixed name. A reader looking for the
rejection-on-arrival DSPR game opens `SMDTDSPR.lean` and finds the other one.

**The collection axis.** A module base name identifies the *property*, not its stand-alone versus collection-indexed instance, because the
collection-indexed game subsumes the stand-alone one at `TweakableHashCollection.empty` — which is
why every `Problem` here carries `thColl` and why the `standalone` constructors are instantiations
rather than separate games. That rule was never written down, so the EasyCrypt `…C` marker survived
on one module. (Handled in the `udc-subspace` branch, which already rewrites that file.)

## What

- `SMDTDSPR.lean` → `SMDTDSPRFinalValidity.lean`, `SMDTOpenPRE.lean` → `SMDTOpenPREFinalValidity.lean`,
  with their `VCVioTest/` twins.
- Test namespaces `SMDTDSPRSourceFinalValidityTest` / `SMDTOpenPRESourceFinalValidityTest` →
  `SMDTDSPRFinalValidityTest` / `SMDTOpenPREFinalValidityTest`, matching the SM-DT-TCR and
  SM-DT-PRE test modules.
- Module docstring titles → `Source-final-validity SM-DT-DSPR` / `…SM-DT-OpenPRE`, the form the
  SM-DT-TCR and SM-DT-PRE monitor modules already use.
- Registrations follow: `VCVio.lean`, `VCVioTest.lean`, `OpenPREFromTCRDSPR.lean`'s imports, and the
  three `linting.yml` points.
- **`Collection.lean` gains the naming rule** in its module docstring, next to the paragraph that
  already explains why the collection is a problem parameter rather than a second game. Without it
  the scheme is invisible and the next game added will guess.

`ITSR.lean` is untouched: it lives under `KeyedHash/`, has no collection and no monitor, so neither
axis applies.

## Verification

Base: `main` at `b4449d1a`. All exit 0, no new warnings.

`lake build`; `lake build VCVioTest HashSigTest`; `lake exe axiomsweep --check`;
`lake exe mk_all --lib {VCVio,VCVioTest} --module --check`; `lake exe lint-style` over the full
`linting.yml` list; `check-{extern,interop,complexity-backend}-isolation`; `check-pmf-boundary`.

`git show --stat -M` reports both source modules and both test modules as renames (`R`), and the
whole commit is +25 / −18 across 9 files — every changed line a name, an import, or a docstring
title. No canary statement or proof is touched, so every pin in the four affected test modules
holds unchanged.


quangvdao (approved):
Approved at exact head f152f967 after comprehensive Lean/formalization audit, naming-documentation repair, full hosted CI, and validated pre-restack on approved #623. Registry conflicts were resolved to the final module-name union while retaining all current SLH-DSA and #622 canaries.

quangvdao (approved):
Re-approved at exact head 983993a1 after merging the now-landed #623/main commit. The only conflict was the anticipated lint registry; the final-name/current-canary union builds and lints cleanly, and both parents independently passed full hosted/merge-queue CI.

PR #630, opened 2026-09-03
feat(SLHDSA): add reachable security-target ledgers and target counts

alexanderlhicks:
## Summary

First slice of the SLH-DSA security lane (D1A, target and address ledger) from
`docs/design/slh-dsa-fips205-generalization.md`, stated for arbitrary `d` with the `d = 1`
specialization as corollaries. It adds no games, adversaries, or probability statements.

- `HashSig/SLHDSA/Security/TargetCounts.lean`: the closed vocabulary of the eight distinct-target
  hash roles (`TargetRole`), the per-layer/total tree and WOTS+ instance counts, and the
  formula-derived `targetCount` for each role. Positivity under `Params.Valid`, the geometric-sum
  identity `xmssTreeCount p * (2^hp - 1) = 2^h - 1`, and the `d = 1` values.
- `HashSig/SLHDSA/Security/ReachableTargets.lean`: executable `List Adrs` ledgers over the
  canonical `LayerPosition` / `DigestParts` coordinates for every role, each proved to have length
  exactly `targetCount` (or bounded by it for the WOTS+ TCR and PRE roles; the UD ledger is exact) and
  to be structurally `Nodup`, with the fifteen pairwise disjointness facts between the six structural
  ledgers and the `Nodup` of their concatenation. `EncodedTargetLedgerConditions` names the per-role encoded-distinctness
  obligations a concrete primitive bundle must discharge; nothing here assumes global injectivity
  of compressed SHA-2 addresses.
- `HashSig/SLHDSA/Params.lean`: `Params.len_pos`.
- Both library modules use a plain `public section` with per-declaration `@[expose]` on the ten
  definitions that the test's `decide` and the next slice's `rfl` proofs reduce through, so the
  broad-expose ceiling of `scripts/check-expose-boundary.sh` (HashSig 46) is unchanged.
- `HashSigTest/SLHDSA/ReachableTargets.lean` (`lake exe slhdsa_target_ledger_tests`): canaries on
  two small validated profiles with hand-computed sizes, run-time distinctness, pairwise cross-role
  disjointness, and pinned membership/non-membership of the exact addresses Algorithms 18–20 hash
  for a fixed digest. Wired into `build.yml` and `linting.yml` like the sibling executables.

## Provenance

`ReachableTargets.lean` is ported from the archived donor
`archive/slhdsa-pr597-final-88314278` (author Quang Dao) with imports retargeted to the merged
`Position`/`Fors` APIs and the transitional `Security/Architecture` dependency replaced by the new
[...]

PR #631, opened 2026-09-03
feat(SLHDSA): discharge the encoded target-ledger conditions for every approved profile

alexanderlhicks:
## Summary

Second slice of the SLH-DSA security lane (D1A). It discharges, for both approved instantiations
and every FIPS 205 parameter set, the eight per-role encoded-distinctness obligations that
`EncodedTargetLedgerConditions` states and that the previous slice left as an honest structure.

This closes item 3 of the open list on the closed donor pull request, "approved-family proofs
discharging every field of `EncodedTargetLedgerConditions`".

- `HashSig/SLHDSA/Security/EncodedTargets.lean`: the SHA-2 compressed-address domain, field-level
  sufficient conditions for it, per-address and per-ledger membership in it, and the two
  instantiation theorems. `approvedEncodedTargetLedgerConditions` covers all twelve approved
  profiles at once.
- `HashSig/SLHDSA/Address.lean`: six lemmas (`Adrs.fits_iff`, three `isCanonical_of_fields_*`
  constructors, `type_le_six_of_isCanonical`, `fits_one_type_of_isCanonical`); the last two say a
  canonical address has a type code below seven, hence one that fits the single byte `ADRSc`
  reserves for it.
- `HashSig/SLHDSA/Concrete/FIPS.lean`: `sha2AdrsKey_injective_of_domain` drops its two
  `Fits 1 _.type` hypotheses, which canonicality now supplies; its only caller is this slice.
- `HashSigTest/SLHDSA/EncodedTargets.lean` (`lake exe slhdsa_encoded_ledger_tests`): canaries that
  push every ledger entry through the checked SHA-2 boundary, compare the encoded key lists for
  duplicates directly at both key widths, and pin the negative direction.

## Why the conditions are not free

A concrete encoder has narrower field domains than `ValidatedParams` imposes. SHA-2 compresses the
address to twenty-two bytes with a one-byte layer and an eight-byte tree, and its total projection
maps anything outside that domain to the all-zero key. That key is not a sentinel: it is the
genuine key of the all-zero WOTS-hash address, so an out-of-domain address aliases a legitimate
reachable one. The canaries pin exactly this aliasing, so the proof obligation is visibly about
avoiding a real collapse rather than about a formality.

`Params.ApprovedAddressBounds` names the arithmetic side conditions that rule the collapse out. Every
FIPS 205 parameter set and the limited SHA2-128-24 profile satisfy them by evaluation.

## Stack

Stacked on #630. Review this against `feat/slhdsa-d1a-reachable-targets-20260903`; the diff against
that branch is exactly this slice.

## What this does not claim

Nothing here is a security theorem. Distinct encoded tweaks are a precondition of the
distinct-target hash games, not a reduction. The trace-level statement that the construction's free
programs query only listed addresses, the canonical game instantiations, and the witness
translations are the following slices.

## Validation (local, exact head)

- `lake build HashSig HashSigTest`
- `lake exe slhdsa_encoded_ledger_tests` and `lake exe slhdsa_target_ledger_tests` — PASS
- `lake exe lint-style` on the three touched modules — clean
- `lake exe mk_all --lib HashSig --module --check` — no update necessary
- `scripts/check-agent-docs.py`, `scripts/check-pmf-boundary.sh`, `git diff --check` — clean
- `scripts/check-expose-boundary.sh` — HashSig 46 → 46 (the new module uses a plain
  `public section`; nothing in it needs `@[expose]`)
- full non-test library build, then `lake exe axiomsweep --check` — PASS, no new `sorryAx` and zero
  non-standard-axiom taint
- Independent adversarial reviews of every commit; each round's findings and dispositions are posted
  as a comment below


quangvdao (approved):
Reviewed at `c682b88ba1e5218c0d4803cc6b3d6fc18f4919ae`, after integrating the landed #630 parent and retargeting to main. Sol High reviewed the full slice, bounds/specification and negative encoder cases; parent inspected the complete new encoded-target theorem module, address/API changes, repairs, and final integration.

No remaining mathematical defect found. The SHA-2 compressed-domain and SHAKE full-width injectivity obligations are distinguished correctly; canonicality supplies the type bound. The repair corrects ADRSc subsection citations and updates the UD evidence after #630: a total count-cap completion is not the reduction's actual target list. Tests now cover a synthetic partial-UD selector through the generic partial-interface theorem for both encoders. Exact EasyCrypt selector/trace correspondence and security-game equivalence are explicitly not established here.

Targeted builds, both ledger executables, full HashSig/HashSigTest build, ordinary-import and standard-axiom probes, lint/manifests/docs/boundaries passed. After the squash-parent integration, all PR-specific source/config contents match the reviewed repair head and HashSig plus the encoded-test target rebuilt successfully (2758 jobs). Fresh full hosted CI and protected queue checks remain required. Native FFI implementations were not validated by these pure address-ledger tests.

PR #640, opened 2026-09-03
feat(Falcon): define the fuel-bounded signer and prove verification correctness

alik-eth:
The abstract sign/correctness slice of #471, restated the honest way. Based on current `main`; independent of the other open Falcon PRs. Closes two `sorry`s in the axiom baseline (`Falcon.sign`, `Falcon.verify_sign_correct`).

## What changes

**`Falcon.sign` is defined** (`LatticeCrypto/Falcon/Scheme.lean`): the Falcon+ signer as a fuel-bounded rejection loop. Each attempt draws a fresh 40-byte salt, hashes the target, runs `signAttempt`, and returns `some (salt, compress s₂)` when both the norm check and compression succeed; it returns `none` once the budget is exhausted. Compression is a separate gate from the norm check, since a preimage can pass the `ℓ₂` bound and still have a coefficient too large for `compress` at `p.sbytelen`. The centered lift `rqToIntPolyCentered` is what the signer compresses, and `toRq_rqToIntPolyCentered` shows reduction modulo `q` undoes it.

**`verify_sign_correct` is proved**, with the obligation #471 had defined away stated as a hypothesis (`LatticeCrypto/Falcon/Security.lean`):

```
theorem verify_sign_correct (pk sk) (msg) (maxAttempts) (sig)
    (hcompress : ∀ s slen bytes, prims.compress s slen = some bytes → prims.decompress bytes slen = some s)
    (hpreimage : ∀ c x, x ∈ support ((falconPSF p prims).trapdoorSample pk sk c) → (falconPSF p prims).eval pk x = c)
    (hsig : some sig ∈ support (sign p prims pk sk msg maxAttempts)) :
    verify p prims pk msg sig = true
```

`hpreimage` is the lattice-point obligation of the floating-point pipeline: `fromFFTPreimage` rounds an inverse FFT to obtain `v = z · B` and returns `(c − v₀, −v₁)`, and `s₁ + s₂ · h = c` holds exactly when the rounded `v` is the lattice point, i.e. when the inverse FFT's rounding error stays below `1/2` in every coordinate. #471 obtained it by redefining `fromFFTPreimage` to set `s₁ := c − s₂ · h`, which made the theorem about a different vector than the one the norm check runs on. Here `fromFFTPreimage` is untouched and the condition is explicit, so it is discharged where the arithmetic is exact (the degree-2 witness in #466 proves it outright for its instance, `NonVacuity.toy_eval_sOf`) and stays visible where it is not.

`hcompress` is the `compress_decompress` law of `Primitives.Laws`, taken on its own rather than through `Primitives.Laws` because that structure also carries `samplerZ_correct`, an exact discrete-Gaussian law that no finite-support `ProbComp` sampler satisfies.

**The codec round-trip is proved for the concrete Golomb-Rice codec** (`LatticeCrypto/Falcon/Concrete/Encoding.lean`): `decompress_compress : compress n s dlen = some d → decompress n d dlen = some s`. The two functions are restated over the bit stream. The previous definitions used `while` loops, which `do`-notation desugars through `Loop.forIn`, a `partial` definition, so nothing could be proved about them as written (the same reason #638 restated the public-key codec as bounded loops). The new `compress` emits, per coefficient, the sign bit, the seven low bits of `|x|`, `|x| / 128` zeros and a one, packs most significant bit first and zero-pads to `dlen`; `decompress` parses that stream and rejects a wrong length, an out-of-range magnitude, the negative zero, and nonzero padding, exactly the failure cases of the old decoder. Behaviour is pinned two ways: `lake exe falcon_test` passes 205/205 with the new codec, including the known-answer signatures through `concreteVerify` and the C reference verifying Lean-produced signatures, and a differential test of the new functions against the old imperative ones found no disagreement over 3680 polynomial encodings at boundary lengths, 9000 random byte strings, and 2500 bit-flipped encodings.

**At the concrete primitives only the coset condition remains** (`Extern/Falcon/SignCorrectness.lean`, new): `verify_sign_correct_verifyPrimitives` and `verify_sign_correct_concretePrimitives` discharge `hcompress` by the round-trip, so `verify` accepts every signature the concrete signer emits under `hpreimage` alone.

**The coset condition is reduced to the primitives** (`LatticeCrypto/Falcon/Coset.lean`, new). `LandsOnLattice prims sk z` says the two rounded inverse transforms in `fromFFTPreimage` are the integer polynomials `z₀ · g + z₁ · G` and `−(z₀ · f + z₁ · F)` for some integer `(z₀, z₁)`, i.e. the basis application lands on the lattice point `z · B`. Given that and the key relations `f · h = g`, `F · h = G`, the identity `s₁ + s₂ · h = c` is ring algebra, carried out in the semantic quotient `ℤ_q[X]/(X^n + 1)` through the existing soundness certificate `coeffSemantics` (`eval_fromFFTPreimage_of_landsOnLattice`); `hpreimage_of_landsOnLattice` lifts it over the sampler's support, and `capRelation_of_validKeyPair` derives `F · h = G` from the NTRU equation and `f · h = g` when `f` is invertible modulo `q`, which key generation guarantees. To make the NTRU equation usable, the integer lift's multiplication is now the `Finset`-sum specification `negacyclicMulPure` rather than the imperative schoolbook loop; the two were already the same function at runtime (`implemented_by`), and the loop was opaque to proofs. `verify_sign_correct_concretePrimitives_of_landsOnLattice` (and the `verifyPrimitives` version) then assume only key validity, invertibility of `f`, and `LandsOnLattice` for the sampler's outputs.

So the residual for the real signer is now exactly one deterministic statement about the fixed-point primitives: that `fftInt`/`ifftRound` round the basis application to the exact lattice point. The abstract sampler itself runs over exact reals (`Real.cos`/`Real.sin` twiddles), so no floating-point analysis of `ffSampling` enters; what is left is the rounding-error bound of `FXR.vect_FFT`/`vect_iFFT` on integer-valued inputs, deferred.

Neither the NTRU equation nor `validKeyPair` is needed: the signer emits only attempts that passed the norm check, and that check is `verify`'s own. The proof is an induction on the attempt budget: the accepting attempt is a `trapdoorSample` output that passed `isShort`; `verify` decompresses to the same `s₂` by `hcompress`, recomputes `s₁ = c − s₂ · h`, which equals the sampled `s₁` by `hpreimage`, and runs the same norm check.

## Not claimed

Nothing about the distribution of `sign`'s output, about `keyGenFromSeed` (still a placeholder), or about the concrete floating-point sampler satisfying `hpreimage`. The codec restatement is not a faithfulness claim against a c-fn-dsa revision beyond what the test executable checks. Composing #514's per-operation bounds through `ffSampling` and the inverse FFT to discharge `hpreimage` for the real arithmetic is the deferred step.

## Checks

`verify_sign_correct`, `toRq_rqToIntPolyCentered`, `sign`, `decompress_compress`, the coset theorems (`toRq_mul`, `eval_fromFFTPreimage_of_landsOnLattice`, `hpreimage_of_landsOnLattice`, `capRelation_of_validKeyPair`), and all four concrete corollaries depend only on `propext`, `Classical.choice`, `Quot.sound`. Both touched files pass `-Dweak.linter.mathlibStandardSet=true` with no new warnings (the remaining `sorry` warnings are the pre-existing `keyGenFromSeed` and one-shot `euf_cma_security`). Full build of all libraries and `LatticeCryptoTest` clean; PMF boundary, Extern isolation, Interop isolation, and `lake exe axiomsweep --check` pass with the baseline regenerated.

### DIFF
--- a/Extern.lean
+++ b/Extern.lean
@@ -9,6 +9,7 @@ public import Extern.Falcon.KeyGen
 public import Extern.Falcon.SamplerZ
 public import Extern.Falcon.Sampling
 public import Extern.Falcon.Sign
+public import Extern.Falcon.SignCorrectness
 public import Extern.Hashing
 public import Extern.MLDSA.FFI
 public import Extern.MLDSA.Instance
--- a/Extern/Falcon/SignCorrectness.lean
+++ b/Extern/Falcon/SignCorrectness.lean
@@ -9,6 +9,7 @@ import all Extern.Falcon.Instance
 import all Extern.Falcon.FPRBridge
 public import Extern.Falcon.FPRBridge
 public import LatticeCrypto.Falcon.Security
+public import LatticeCrypto.Falcon.Coset
 
 /-!
 # Verification correctness at the concrete Falcon primitives
@@ -17,7 +18,9 @@ public import LatticeCrypto.Falcon.Security
 trapdoor sampler lands on the coset. At `Falcon.Concrete.FPRBridge.verifyPrimitives` and
 `Falcon.Concrete.concretePrimitives` the codec is the concrete Golomb-Rice codec, whose round-trip
 `Falcon.Concrete.decompress_compress` is a theorem, so only the coset condition `hpreimage`
-remains: the lattice-point obligation of the floating-point inverse FFT.
+remains. `Falcon.hpreimage_of_landsOnLattice` reduces that condition further, to the
+lattice-point condition `Falcon.LandsOnLattice` on the fixed-point FFT pipeline together with
+the key relations; the `_of_landsOnLattice` variants below take only those.
 -/
 
 public section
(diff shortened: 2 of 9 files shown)


alik-eth:
Superseded by #663, which was stacked on this branch and carried all four of its commits into main (79c470bc). Every declaration added here, including `Falcon.sign`, `verify_sign_correct`, the codec round-trip and the coset reduction, is on main.

PR #763, opened 2026-09-18
fix(probability): format compact event notation

dtumad:
Compact native event notation such as `Pr{let x ← mx}[p x]` currently triggers Mathlib's whitespace linter in hypotheses because the inherited `doSeqItem` formatter requests a leading break. Format the first statement directly after `Pr{`, preserve explicit opening line breaks, and keep Lean's ordinary formatting for subsequent statements and braced sequences. Normalize existing spaced uses and document the compact spelling.

Add compiler guards for compact, previously spaced, multiline, and braced syntax. The notation's parser and measure semantics are unchanged.

Validation: `./scripts/validate.sh --lint --test --axioms` passed, including native import guards, all proof/test libraries, style and environment lint, boundaries, and axiom/initialization ratchets. No new axiom or sorry debt; lint baselines unchanged.

Stacked on #758. Related to #532.

PR #758, opened 2026-09-18
refactor(probability): strengthen native integration and event laws

dtumad:
Native measure proofs still imposed discrete spaces on unobserved intermediate results and globally measurable valuations. This strengthens integration and event APIs with chosen-space AE laws, replaces artificial discrete intermediate spaces with observation pullbacks, and adds structural oracle interchange and support-based comparison laws without measurable-space or countability requirements on hidden outputs. ML-DSA uses the stronger interchange rule.

A standalone native canary covers usual real spaces, lossy optional results, AE valuations and events, arbitrary hidden outputs, and a continuous Gaussian query; its import guard excludes the retired probability surface.

Validation: `./scripts/validate.sh --lint --test --axioms` passed, including environment lint, tests, and axiom accounting with no new debt. Checkpoint 1 of #532; further conversions will be published separately.

Issue #532, opened 2026-08-21
Design and roadmap: Mathlib-aligned denotational probability semantics


dtumad:
## Status

The design is settled: Mathlib `Measure` and `Kernel` are VCVio's probability semantics. The
initial implementation (#531) has been followed by native sampling, integration, coupling,
total variation, unary/relational WP, stateful security, instrumentation, and scheme-definition
conversions. The design record is `docs/reading/denotational-probability-semantics.md`; the live
checkpoint and proof-conversion guide is `docs/design/measure-conversion-roadmap.md`.

This issue remains the campaign tracker. Earlier comments describe historical implementation
states; this body records the current policy and remaining work. PMF/SPMF and their compatibility
classes are retiring, rather than coequal or preferred representations for finite proofs.

Recent landed checkpoints include core tracking and scheme definitions (#742), native expected
query costs (#749), and abort analysis (#754). Independent products are in #756. Fischlin exact
expected signing costs (#752) and Schnorr transform guarantees (#755) are separate in-flight
work that subsequent conversions must preserve.

## Canonical representations

| Need | Representation |
|---|---|
| Operational possible returns and invariants | Core/PolyFun support, attachment, indexed reachability, and WP |
| Closed successful outputs | Mathlib `Measure α`, with `IsSubprobabilityMeasure` |
| Losslessness | Mathlib `IsProbabilityMeasure` |
| Environment/state-indexed semantics | Mathlib `Kernel ρ α` / `Kernel σ (α × σ)` |
| Explicit optional, exceptional, writer, and state outcomes | Measures on the actual outcome types; observation selects which effects to discard |
| Event probabilities | Measure-backed `Pr{...}[...]`, or measure application to a measurable event |
| Expectations | Mathlib integration of the chosen observation against the denotation |
| Program logic | Core `Std.Internal.Do`, upstream ordered algebras and operation specifications |
| Returned values of resumptions | Subprobability measures; retain cutoff, failure, and divergence distinctions |
| Exact finite execution | Existing executable samplers, including `FinRatPMF.Raw`, with native measure certificates |
| Infinite trace laws | Future measurable coalgebra and Mathlib kernel/trajectory machinery |

`evalDist` / `𝒟[...]` are primary native semantics. `evalSPMF` / `𝒮[...]`, legacy `Pr[...]`,
scalar evaluation functions, and compatibility mixins are deprecated migration debt.

## API and conversion policy

[...]

PR #764, opened 2026-09-18
feat(probability): add native continuation comparison laws

dtumad:
Provide native common-prefix event comparisons for converting the PRF reader proofs. AE conditional bounds against finitely many reference events integrate their varying allowance. Reachable bounds use core attachment and the actual continuation-measure observer; hidden source and result payloads need no measurable space. Constant allowances retain the prefix's success mass, and all four disagreement/bad-world shapes share this argument.

The Mathlib-facing rules support arbitrary prefix measures, AE continuation families, and reference output spaces that vary with the index. Existing Mathlib kernels use the same rule. Neither probability/finiteness certificates nor measurability of the allowance are required. One finite-set integral lemma owns the shared comparison proof.

Export the computational API through `VCVio.Native`. Regressions reject PMF/SPMF and retired compatibility classes, and cover a Gaussian source with real/Boolean kernels, arbitrary AE continuations under a Dirac measure, and unmeasured source/result types.

Validation: `./scripts/validate.sh --lint --test --axioms` passed. All baselines unchanged; the axiom sweep checks 21,900 declarations across 752 modules with no new taint. No new `change` statements, heartbeat increases, or probability shims.

This is an API prerequisite for #532's full PRFTagReader conversion. The reader/slot/composition family and its existing scalar disagreement consumers remain open; no retiring declaration is moved or wrapped here. Stacked on #763 and #758.


quangvdao:
Superseded by #771, which landed this content on `main` (`210d73fd`). Checked before closing: merging this head into `main` leaves only conflict residue (pre-reformat `Pr{ let` spacing, an older roadmap section, a stale test-import list), and every declaration name in the top of this stack (#764) already exists on `main`. The branch is left in place because #761 is still based on it.

